PR #397, opened 2024-04-22
ci: add Anvil option to JS benchmark runner

fvictorio:
Manually re-applying changes from https://github.com/NomicFoundation/hardhat/pull/5150

I tried cherry-picking them, but it was causing me a lot of issues, I'm not sure why.

### DIFF
--- a/crates/tools/js/benchmark/README.md
+++ b/crates/tools/js/benchmark/README.md
@@ -12,3 +12,11 @@ pnpm run benchmark
 The measurements will be printed to stdout as machine-readable json and to stderr and saved to `./benchmark-output.json` disk as json.
 
 Please see `pnpm run help` for more.
+
+### Anvil mode
+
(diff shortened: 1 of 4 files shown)


Wodann (approved):
LGTM, but I'll cede to Franco for a final review

fvictorio:
I don't think the `@foundry-rs/hardhat-anvil` part is necessary if we are starting anvil manually? I removed it in 61eccfe3ab0cb775a5a79db6683cd9e0d1eda147 and tried it and worked fine without it. And I'd rather not use that dependency (it's a bit of ahack, and I'm not sure it's maintained).

But if I'm missing something here, feel free to revert that commit.

agostbiro:
@fvictorio yeah `@foundry-rs/hardhat-anvil` it's not needed anymore, thanks for removing!

PR #401, opened 2024-04-23
Remove usage of batch calls

fvictorio:


.github/workflows/edr-benchmark.yml
```diff
@@ -35,7 +35,8 @@ jobs:
     environment: github-action-benchmark
     runs-on: self-hosted
     # Only run for trusted collaborators since third-parties could run malicious code on the self-hosted benchmark runner.
-    if: github.ref == 'refs/heads/main' || (github.event.pull_request.author_association == 'OWNER' || github.event.pull_request.author_association == 'MEMBER' || github.event.pull_request.author_association == 'COLLABORATOR')
+    # if: github.ref == 'refs/heads/main' || (github.event.pull_request.author_association == 'OWNER' || github.event.pull_request.author_association == 'MEMBER' || github.event.pull_request.author_association == 'COLLABORATOR')
```
Wodann, line 38: Why was this commented?

crates/edr_eth/src/remote/client.rs
```diff
@@ -969,23 +814,19 @@ impl RpcClient {
...
     pub async fn get_transaction_receipts(
         &self,
         hashes: impl IntoIterator<Item = &B256> + Debug,
     ) -> Result<Option<Vec<BlockReceipt>>, RpcClientError> {
-        let requests: Vec<RequestMethod> = hashes
+        let requests = hashes
             .into_iter()
-            .map(|transaction_hash| RequestMethod::GetTransactionReceipt(*transaction_hash))
-            .collect();
+            .map(|transaction_hash| self.get_transaction_receipt(transaction_hash));
```
Wodann, line 826: Doesn't clippy complain about this?

Can't it be `map(RpcClient::get_transaction_receipt)`?

crates/edr_eth/src/remote/request_methods.rs
```diff
@@ -100,6 +100,7 @@ pub enum RequestMethod {
 }
 
 impl RequestMethod {
```
Wodann, line 102: Is this only used when tracing is enabled?

.github/workflows/edr-benchmark.yml
fvictorio: Sorry, I thought I had added a comment about this but clearly I didn't. I think this job is not working with the new repo; my guess is that some extra configuration is needed. I opened https://github.com/NomicFoundation/edr/issues/402 about it.

crates/edr_eth/src/remote/request_methods.rs
fvictorio: Yes. I added this because otherwise `cargo hack` would complain when running clippy without the tracing feature.

Is there a better way to handle this?

crates/edr_eth/src/remote/client.rs
fvictorio: My understanding is: `get_transaction_receipt` is a method, so it can't be called as `RpcClient::get_transaction_receipt`. And [apparently](https://users.rust-lang.org/t/using-method-directly-instead-of-writing-lambda/85442/7) Rust doesn't support methods in those scenarios, so you can't do `.map(self.get_transaction_receipt)`.

Wodann: The syntax for doing `.map(self.get_transaction_receipt` is `RpcClient::get_transaction_receipt`, as the functions first argument is `self`

fvictorio: `.map(RpcClient::get_transaction_receipt)` doesn't compile though

crates/edr_eth/src/remote/request_methods.rs
Wodann: Nope, just validating

crates/edr_eth/src/remote/client.rs
Wodann: Sorry, I updated my response. I overlooked the two arguments and noticed that when I tried it locally

PR #408, opened 2024-04-25
Include hardhat tests

fvictorio:
This PR copies a subset of the `hardhat-core` tests into our repo and runs them using the workspace version of EDR. Reviewing the PR is of course impossible, so I'll just detail what I did; please tell me if you think I missed something.

## Copying the code

This was the easy part. I copied `hardhat-core` as `hardhat-tests`, removed a lot of files (mainly `src/`, but also `sample-projects/`, `scripts/`, `CHANGELOG.md`, etc.). I also wrote a short [README](https://github.com/NomicFoundation/edr/blob/include-hardhat-tests/hardhat-tests/README.md) to explain what this directory is about.

## Running with Hardhat as a dependency

After removing the `src` directory the tests of course stopped working. To make them work again I installed `hardhat` as a dev dependency, and then transformed all the relative imports like `import "../../../src/foo/bar.ts` into `import "hardhat/foo/bar.ts"`.

After doing this, I could run the tests again. Most of them passed; the ones that didn't were assuming some things about the directory structure that was no longer true. But all of these were also tests that didn't use EDR, so I removed them.

## Using the workspace version of EDR

Luckily, doing this was super straightforward. I just added this to the [root's package.json](https://github.com/NomicFoundation/edr/blob/include-hardhat-tests/package.json):

```
  "pnpm": {
    "overrides": {
      "hardhat@2>@nomicfoundation/edr": "workspace:*"
    }
  }
```

Something I don't like about this is that this can't be done just for the `hardhat-tests` workspace. This seems to be a pnpm limitation. But I think it's highly unlikely that we 1) add another workspace that uses hardhat as a dependency and 2) that that workspace _needs_ the npm version of EDR instead of the local one. So I think this risk is fine.

To verify that this was indeed using the local version, I added an `assert!(false)` to the `Provider::with_config` method in `edr_napi`, which is the entry point used by Hardhat, and ran the tests. They panicked, as expected.

The only remaining problem was that this could stop working for some reason in the future and we wouldn't notice it because the npm version of EDR would be used and things would be fine anyway. To prevent this, I added a test in [`hardhat-tests/test/edr-override.ts`](https://github.com/NomicFoundation/edr/blob/include-hardhat-tests/hardhat-tests/test/edr-override.ts) that checks that the local version is used. Please check it to see what it does.

## Removing unnecessary tests

Finally, I wanted to remove all the tests that weren't related to the Hardhat network. Most of them were the tests outside `test/internal/hardhat-network`, as expected. But to check that this is the case, before removing each of them I ran them with a local version of EDR that panics and checked that they passed. Then I did the same thing with the tests I _didn't_ remove: I ran each of them separately and checked that they panicked.

## Workflow

I added a [`.github/workflows/hardhat-tests.yml` workflow](https://github.com/NomicFoundation/edr/blob/include-hardhat-tests/.github/workflows/hardhat-tests.yml). Nothing fancy here: it runs the lint and test scripts of `hardhat-tests` and caches the remote RPC calls and the stack traces artifacts.


Wodann:
Is there an easy way to synchronise between Hardhat and EDR; e.g. using git subtree?

Wodann (approved):
Thanks for the write-up. Very helpful!

I posted one question, but otherwise LGTM once CI is successful

fvictorio:
> Is there an easy way to synchronise between Hardhat and EDR; e.g. using git subtree?

I'm not sure, but we are not going to keep them synchronized anyway. The way to think about it is that these tests are our responsibility now, and they are (eventually) going to be removed from the Hardhat repo.

Issue #388, opened 2024-03-20
Hardhat fork error in v2.21.0 - "missing trie node" 


DeFiFoFum:
### Version of Hardhat

v2.21.0

## Issue
When I upgrade to v2.21.0 and then try to use the forking configuration below, I get the error: `ProviderError: The response reported error -32000: missing trie node. (optional data: None).`

## Fork Configuration
I use a helper function called `setupFork` which follows the forking documentation in the docs: https://hardhat.org/hardhat-network/docs/guides/forking-other-networks#pinning-a-block

Here is an example: https://github.com/DeFiFoFum/hardhat-template/blob/main/lib/evm/forkHelper.ts#L39

```ts
  await network.provider.request({
    method: 'hardhat_reset',
    params: [
      {
        forking: {
          jsonRpcUrl: forkingNetworkConfig.url,
          blockNumber: blockNumber,
        },
      },
    ],
  })
```

## Workaround: use `hardhat@2.14.0`
To fix this issue, I downgraded to `hardhat@2.14.0` and pinned that version. Now the fork configuration works as expected. 

<img width="601" alt="hardhat-bug-report" src="https://github.com/NomicFoundation/hardhat/assets/78645267/d4dcc067-36f7-4cc6-a018-f6643a51a07a">

## Reproduce
Install `hardhat@2.21.0` and then run the `setupFork` function I provided above. 

### Search terms

trie node 2.21.0

Wodann:
Related to #387

Wodann:
@DeFiFoFum and @RealJohnnyTime which node provider are you using when this problem occurs? Infura? Alchemy? Something else?

DeFiFoFum:
> Hi @DeFiFoFum, which service/node are you connecting to?

> @DeFiFoFum and @RealJohnnyTime which node provider are you using when this problem occurs? Infura? Alchemy? Something else?

I generally go down the list on ChainList.org until I find an RPC node which can support archive requests. I also work on multiple networks such as Ethereum, Binance Smart Chain, Polygon, Arbitrum and Linea primarily. Off the top of my head I don't remember exactly which ones were causing the issue, I just know that my workflow started failing after I upgraded above `2.14.0` on CICD and I couldn't get it working again without reverting back to `2.14.0` using the same RPC endpoint. 🤔

Wodann:
[...]

PR #410, opened 2024-04-25
refactor: retry once when receiving some sporadic JSON-RPC errors

### LINKED ISSUE #387: Sporadic failure in tests on CI due to `missing trie node` JSON-RPC response
The `data::tests::full_block_*` tests fail sporadically in CI, with the following errors:

[Occurrence 1 (macOS)](https://github.com/NomicFoundation/hardhat/actions/runs/8701451560/job/23863671535?pr=5126#step:6:1205):

```bash
---- data::tests::full_block_mainnet_19318016 stdout ----
Error: The response reported error `-32000`: `missing trie node 41c10bface9b142aaf6e4f20e97600744cbcbf2db34cf0ec03ede5e84eb29e98 (path ) state 0x41c10bface9b142aaf6e4f20e97600744cbcbf2db34cf0ec03ede5e84eb29e98 is not available, not found`. (optional data: None). Request: {"id":1452,"jsonrpc":"2.0","method":"eth_getStorageAt","params":["0xc49b737fa56f9142974a54f6c66055468ec631d0","0x360894a13ba1a3210667c828492db98dca3e2076cc3735a920a3ca505d382bbc","0x126c4ff"]
```

[Occurrence 2 (Ubuntu)](https://github.com/NomicFoundation/hardhat/actions/runs/8702358798/job/23866467039#step:6:1213): 

```bash
---- data::tests::full_block_mainnet_cancun2 stdout ----
Error: The response reported error `-32000`: `missing trie node f90999ba63edf443f4c27152ca78f729afca615bc4637aaa44662dfdcb5c62c9 (path ) state 0xf90999ba63edf443f4c27152ca78f729afca615bc4637aaa44662dfdcb5c62c9 is not available, not found`. (optional data: None). Request: {"id":546,"jsonrpc":"2.0","method":"eth_getStorageAt","params":["0x2cfc3a07fd9ee6d3227b24fc4ee09b34ae1500b7","0x150595343228aa057f2d5dac3ed2d5618e3d52c474caabf780038e663d7cbf5f","0x12a7e3e"]}
```

The error is being bubbled up from our remote JSON-RPC requests when using Alchemy as a provider.

**Definition of Done**
Understand why the sporadic failure occurred and potentially create a fix

Wodann:
Resolves #387 & #388

### DIFF
--- a/crates/edr_eth/src/remote/client.rs
+++ b/crates/edr_eth/src/remote/client.rs
@@ -236,7 +236,7 @@ impl RpcClient {
         request: SerializedRequest,
     ) -> Result<T, RpcClientError> {
         let is_missing_trie_node_error =
-            error.code == -32000 && error.message.contains("missing trie node");
+            error.code == -32000 && error.message.to_lowercase().contains("missing trie node");
 
         let result = if is_missing_trie_node_error {
             self.send_request_body(&request)
--- a/crates/edr_eth/src/remote/client.rs
+++ b/crates/edr_eth/src/remote/client.rs
@@ -235,11 +235,10 @@ impl RpcClient {
         error: jsonrpc::Error,
         request: SerializedRequest,
     ) -> Result<T, RpcClientError> {
-        fn is_missing_trie_node_error(error: &jsonrpc::Error) -> bool {
-            error.code == -32000 && error.message.contains("missing trie node")
-        }
+        let is_missing_trie_node_error =
+            error.code == -32000 && error.message.contains("missing trie node");
 
-        let result = if is_missing_trie_node_error(&error) {
+        let result = if is_missing_trie_node_error {
             self.send_request_body(&request)
                 .await
                 .and_then(Self::parse_response_str)?
@@ -489,6 +488,8 @@ impl RpcClient {
                 .data
                 .into_result(),
         )
+        // We retry at the application level because Alchemy has sporadic failures that are returned
+        // in the JSON-RPC layer
         .or_else(|error| async { self.retry_on_sporadic_failure(error, request).await })
         .await
     }
--- a/crates/edr_eth/src/remote/client.rs
+++ b/crates/edr_eth/src/remote/client.rs
@@ -8,7 +8,7 @@ use std::{
     time::{Duration, Instant},
 };
 
-use futures::stream::StreamExt;
+use futures::{future, stream::StreamExt, TryFutureExt};
 use hyper::header::HeaderValue;
 pub use hyper::{header, http::Error as HttpError, HeaderMap};
 use reqwest::Client as HttpClient;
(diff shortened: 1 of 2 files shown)


crates/edr_eth/src/remote/client.rs
```diff
@@ -220,27 +220,39 @@ impl RpcClient {
...
+    async fn retry_on_sporadic_failure<T: DeserializeOwned>(
+        &self,
+        error: jsonrpc::Error,
         request: SerializedRequest,
-        response: String,
     ) -> Result<T, RpcClientError> {
-        let response: jsonrpc::Response<T> = Self::parse_response_str(&response)?;
+        fn is_missing_trie_node_error(error: &jsonrpc::Error) -> bool {
+            error.code == -32000 && error.message.contains("missing trie node")
+        }
```
fvictorio, line 240: Our of curiosity, why is this a function and not a variable? That is, why not do this instead:

```rust
        let is_missing_trie_node_error =
            error.code == -32000 && error.message.contains("missing trie node");

        let result = if is_missing_trie_node_error {
```

Is using a function more idiomatic in Rust for some reason?

crates/edr_eth/src/remote/client.rs
```diff
@@ -220,27 +220,39 @@ impl RpcClient {
...
 
-        response
-            .data
-            .into_result()
-            .map_err(|error| RpcClientError::JsonRpcError {
-                error,
-                request: request.to_json_string(),
-            })
+        let result = if is_missing_trie_node_error(&error) {
+            self.send_request_body(&request)
```
fvictorio, line 243: Should we wait some milliseconds here before re-sending? I'm fine with not doing it, just wondering out loud.

fvictorio (approved):

crates/edr_eth/src/remote/client.rs
Wodann: I don't think it matters in this case. At the middleware layer we're already retrying with a delay to avoid too many requests, so it would be duplicating part of that logic.

Wodann: https://github.com/NomicFoundation/edr/pull/410/commits/0519bc68a5dbed5ac60ba299d30712f5cb63e233

crates/edr_eth/src/remote/client.rs
Wodann: There is no particular reason. I actually don't know why I didn't do it your way; I like it better 😁 

Changed in https://github.com/NomicFoundation/edr/pull/410/commits/0519bc68a5dbed5ac60ba299d30712f5cb63e233

agostbiro (approved):
Thanks LGTM

crates/edr_eth/src/remote/client.rs
```diff
@@ -220,27 +220,38 @@ impl RpcClient {
...
-    fn extract_result<T: DeserializeOwned>(
+    async fn retry_on_sporadic_failure<T: DeserializeOwned>(
+        &self,
+        error: jsonrpc::Error,
         request: SerializedRequest,
-        response: String,
     ) -> Result<T, RpcClientError> {
-        let response: jsonrpc::Response<T> = Self::parse_response_str(&response)?;
+        let is_missing_trie_node_error =
+            error.code == -32000 && error.message.contains("missing trie node");
```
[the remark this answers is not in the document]
Wodann: Addressed in https://github.com/NomicFoundation/edr/pull/410/commits/0449e1b1f8b2e453d48c8be7a1db3069c734514f

PR #416, opened 2024-04-29
ci: enable benchmark runner

agostbiro:
This PR brings the following changes:

- Enable benchmark CI in this repo after migrating from the HH repo
- Updated execution time and result variance measurements. 
- Execution time variance went up since last measurements which is to be expected due to faster execution times after performance optimizations. To address this, the number of iterations have been adjusted.
- A welcome surprise is that the result variance in `openzeppelin` is fixed since the last measurement so the workaround for this has been removed.
- Result variance on `neptune-mutual` also went down drastically, so instead of skipping validation there, I enabled the same small tolerance approach that was previously used with `openzeppelin`.

I've tested that the runner can write to the [benchmark results repo.](https://github.com/nomic-foundation-automation/edr-benchmark-results)

About a month ago, we wanted to migrate to a new benchmark runner instance, because the `synthetix` results didn’t fit into memory at the time which caused a large variance in running times due to unpredictable GC. Lack of server inventory prevented migrating to a new server back then and constant time state checkpointing drastically reduced memory usage, so while this PR originally intended to migrate to a benchmark runner with more memory, it turned out not to be necessary anymore.


Wodann (approved):
LGTM.

There is one thing that I find impossible to validate in PRs to the benchmarking, which I think is good to mention:

Changes to permitted failures in the snapshot.json have no meaning to me. I'm not sure whether they're actually supposed to fail; or whether we just have trouble generating consistent results.

agostbiro:
> Changes to permitted failures in the snapshot.json have no meaning to me. I'm not sure whether they're actually supposed to fail; or whether we just have trouble generating consistent results.

Thanks, good point! In this PR only permitted failures for `openzeppelin` and `neptune-mutual` changed. The reason for these changes is that we can now produce consistent results for `openzeppelin` and result variance greatly decreased for `neptune-mutal` due to a change since the last snapshot (we don't know exactly which unfortunately). The only case where now failures are allowed to vary is `neptune-mutual`. So if permitted failures change for any scenario other than `neptune-mutual` in the future, the reason should be clarified in review.

I'll go ahead and merge this without waiting for @fvictorio's approval to unblock other PRs as I discussed the functional changes (as opposed to code changes) with him already verbally on Monday.

PR #423, opened 2024-04-29
docs: add instructions for profiling with perf/DTrace

agostbiro:
Add instructions for profiling with perf/DTrace.

### DIFF
--- a/book/src/01_getting_started/05_profiling.md
+++ b/book/src/01_getting_started/05_profiling.md
@@ -1,5 +1,35 @@
 # Profiling
 
+## Perf/DTrace
+
+The [cargo-flamegraph](https://github.com/flamegraph-rs/flamegraph) tool can be used to collect performance profiling data using [perf](https://en.wikipedia.org/wiki/Perf_(Linux)) on Linux and [DTrace](https://en.wikipedia.org/wiki/DTrace) on MacOS/Windows and then visualize it as a flamegraph.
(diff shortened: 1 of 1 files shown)


book/src/01_getting_started/05_profiling.md
```diff
@@ -1,5 +1,35 @@
 # Profiling
 
+## Perf/DTrace
+
+The [cargo-flamegraph](https://github.com/flamegraph-rs/flamegraph) tool can be used to collect performance profiling data using [perf](https://en.wikipedia.org/wiki/Perf_(Linux)) on Linux and [DTrace](https://en.wikipedia.org/wiki/DTrace) on MacOS/Windows and then visualize it as a flamegraph.
```
Wodann, line 5: Why didn't this work when running from JS? If the shared library (created by N-API) is built with debug symbols, running `perf` should also samples from EDR, if I'm not mistaken.

That was my hope with your approach too.

Wodann (approved):
LGTM. Just one question

book/src/01_getting_started/05_profiling.md
agostbiro: I got `failed to sample program` when trying to pass `node` as the binary to profiel. I spent some time trying to figure it out, but gave up after a while. Possibly a custom build of `node` would help, but felt that was too much effort.

PR #425, opened 2024-05-01
Fix EIP-712 problem with int data

fvictorio:


fvictorio:
Closing in favor of https://github.com/NomicFoundation/edr/pull/446

PR #428, opened 2024-05-01
build: bump revm to commit aceb093

Wodann:
I'm okay waiting to merge this until after benchmarks are re-enabled. I'm curious to see whether performance is at all affected by changes in `revm`

### DIFF
--- a/crates/edr_evm/src/transaction/executable.rs
+++ b/crates/edr_evm/src/transaction/executable.rs
@@ -145,7 +145,7 @@ impl From<ExecutableTransaction> for TxEnv {
                 access_list: Vec::new(),
                 blob_hashes: Vec::new(),
                 max_fee_per_blob_gas: None,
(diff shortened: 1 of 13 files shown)


crates/edr_evm/src/transaction/executable.rs
```diff
@@ -145,6 +145,9 @@ impl From<ExecutableTransaction> for TxEnv {
                 access_list: Vec::new(),
                 blob_hashes: Vec::new(),
                 max_fee_per_blob_gas: None,
+                // https://github.com/NomicFoundation/edr/issues/427
```
Wodann, line 148: This is a prague feature, which we don't support yet.

agostbiro (approved):

fvictorio:
> I'm okay waiting to merge this until after benchmarks are re-enabled. I'm curious to see whether performance is at all affected by changes in revm

Yes, let's wait :+1:

PR #436, opened 2024-05-07
build: bump Rust to v1.78

Wodann:
Bumps Rust to 1.78. Also renames `.cargo/config` to `.cargo/config.toml` to resolve compile-time warnings.

No changeset required as this is not a user-facing change.


agostbiro (approved):
Thanks, LGTM

PR #437, opened 2024-05-07
feat: verbose tracing for hardhat-tracer

agostbiro:


crates/edr_napi/src/trace.rs
```diff
@@ -89,24 +89,60 @@ pub struct TracingStep {
...
 
 impl TracingStep {
     pub fn new(step: &edr_evm::trace::Step) -> Self {
+        let stack_top = step.stack.top().map(u256_to_bigint);
+        let stack = step
+            .stack
+            .full()
+            .map(|stack| stack.iter().map(u256_to_bigint).collect());
+        // Memory in steps is Vec<u8>
+        // TODO can we assume 32-byte alignment?
```
agostbiro, line 108: @fvictorio: return `Buffer` instead

PR #444, opened 2024-05-10
feat: support input as alias in eth_call and eth_sendTransaction

### LINKED ISSUE #360: Support `input` in tx objects in addition to `data`
### Version of Hardhat

2.19.1

### What happened?

When sending eth_call to a contract for accessing read-only function with golang of Ethereum geth 1.13, the response from Hardhat is "unknown selector" error. 

The reason for this error is that Hardhat expect contract input provided in the field `data` , but ethereum geth 1.13 is sending it in field name `input`. 

This is RPC message sent with golang of geth 1.13.5

```
{"jsonrpc":"2.0","id":1,"method":"eth_call","params":[
    {   
        "from":"0x0000000000000000000000000000000000000000",
        "input":"0x8b1329e0",
        "to":"0x5fbdb2315678afecb367f032d93f642f64180aa3"
    },  
    "latest"
]}
```

and this is RPC message sent with hardhat `run` command:

```
{"jsonrpc":"2.0","method":"eth_call","params":[
    {   
        "from":"0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266",
        "to":"0x5fbdb2315678afecb367f032d93f642f64180aa3",
        "data":"0x8b1329e0"
    },  
    "latest"
],"id":4}
```

As you can see Hardhat encodes the RPC data with `data` field, while geth uses `input` field. For this reason I am getting `unknown selector` error because Hardhat simply doesn't read contract input when decoding JSON text sent over HTTP

How to reproduce: 
1. Create a contract with read-only function
2. Download geth 1.13 and generate contract wrappers using `abigen` tool
3. Create a program in golang to send eth_call to the read-only function

The result is "unknow selector" error and execution of a fallback function

### Search terms

_No response_

Wodann:
Resolves #360

### DIFF
--- a/.changeset/serious-adults-smash.md
+++ b/.changeset/serious-adults-smash.md
@@ -1,0 +1,5 @@
+---
+"@nomicfoundation/edr": patch
+---
+
+Fixed eth_call and eth_sendTransaction by allowing input instead of data to match gRPC
--- a/.changeset/serious-adults-smash.md
+++ b/.changeset/serious-adults-smash.md
@@ -2,4 +2,4 @@
 "@nomicfoundation/edr": patch
 ---
 
-Fixed eth_call and eth_sendTransaction by allowing input instead of data to match gRPC
+Fixed eth_call and eth_sendTransaction to allow passing input data with both `input` and `data` fields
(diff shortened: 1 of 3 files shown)


agostbiro (approved):
Thanks, LGTM. Do you wanna review @fvictorio ? Otherwise we can merge

fvictorio (approved):

PR #449, opened 2024-05-14
Include all builds as normal dependencies

fvictorio:
Hardhat users are frequently running into https://github.com/npm/cli/issues/4828. As a blunt workaround, we are going to turn all the builds into dependencies of the entry-point package `@nomicfoundation/edr`, instead of being optional dependencies. For this to work, two things have to happen:

1. The `package.json` of EDR has to use them as plain `dependencies` instead of `optionalDependencies`
2. The `package.json`s of the builds don't have to include things like `os` and `arch`, because the package manager might check them and complain if they don't match.

Since the `optionalDependencies` field of the entry-point package.json are added/modified by the NAPI-RS CLI, the only way to deal with 1 is to modify the `package.json` after running `napi prepublish`. This is done with `jq` and `sponge` (sponge is needed because you can't do `jq '...' package.json > package.json` and `jq` doesn't have a "modify in place" flag like sed's `-i`). `sponge` doesn't come pre-installed, so we need to install `moreutils` in the publish job.

For 2, it's enough to just modify the `package.json`s we already have under `crates/edr_napi/npm`. The `arch`, `os` and `libc` fields are not re-added by `napi`.

### DIFF
--- a/crates/edr_napi/package.json
+++ b/crates/edr_napi/package.json
@@ -17,11 +17,15 @@
   "napi": {
     "name": "edr",
     "triples": {
+      "defaults": false,
       "additional": [
         "aarch64-apple-darwin",
+        "x86_64-apple-darwin",
         "aarch64-unknown-linux-gnu",
         "aarch64-unknown-linux-musl",
-        "x86_64-unknown-linux-musl"
+        "x86_64-unknown-linux-gnu",
+        "x86_64-unknown-linux-musl",
+        "x86_64-pc-windows-msvc"
       ]
     }
   },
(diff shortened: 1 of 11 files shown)


crates/edr_napi/package.json
```diff
@@ -17,11 +17,15 @@
   "napi": {
     "name": "edr",
     "triples": {
+      "defaults": false,
```
fvictorio, line 20: This change isn't strictly necessary for this PR, but while working on this I noticed that slang does it and I think it makes sense. This way we don't rely on napi-rs's defaults that might change without us noticing it.

fvictorio:
Also: I tested this in the `prerelease` branch ([job](https://github.com/NomicFoundation/edr/actions/runs/9078124803/job/24945136358)) and it seems to work. You can install `@nomicfoundation/edr@0.4.0-alpha.3` to try it.

agostbiro (approved):
Thanks, LGTM! I tested out `@nomicfoundation/edr@0.4.0-alpha.3` on MacOS ARM with a fresh install and it worked well 🎉

Issue #455, opened 2024-05-15
eth_sign arguments are reversed


area:
### Version of Hardhat

2.22.4

### What happened?

Calling `eth_sign` via RPC against a node run with `npx hardhat node` does not work and returns an error.

### Minimal reproduction steps

```
npx hardhat node
curl -H 'Content-Type: application/json' http://127.0.0.1:8545 -X POST --data-raw '{"method":"eth_sign","params":["0x8626f6940E2eb28930eFb4CeF49B2d1F2C9C1199","0x06fdde03"],"id":574,"jsonrpc":"2.0"}'
```
returns
```
{"jsonrpc":"2.0","id":574,"error":{"code":-32602,"message":"invalid value \"0x06fdde03\" supplied to : ADDRESS at line 1 column 89","data":{"message":"invalid value \"0x06fdde03\" supplied to : ADDRESS at line 1 column 89","data":{"method":"eth_sign","params":["0x8626f6940E2eb28930eFb4CeF49B2d1F2C9C1199","0x06fdde03"]}}}}
```

Reversing the arguments:

```
curl -H 'Content-Type: application/json' http://127.0.0.1:8545 -X POST --data-raw '{"method":"eth_sign","params":["0x06fdde03", "0x8626f6940E2eb28930eFb4CeF49B2d1F2C9C1199"],"id":574,"jsonrpc":"2.0"}'
```

and a signature is returned as expected

```
{"jsonrpc":"2.0","id":574,"result":"0x59002ebdb95ef3b258613f6bba6e91fd392bc49d86ad8a1e1afd32b929ee572968529c9189dde90d4a0d315132fcce215b28bfa479fe44cb05021a51e37ab6b81c"}
```

However, the [specification](https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_sign) says that the address doing the signing should be the first argument, with the message to sign being second.

Like https://github.com/NomicFoundation/edr/issues/399, this was run across this upgrading from a pre-EDR hardhat to a post-EDR hardhat, so it's possible the issue is there, but haven't (successfully) investigated. The text of the error being returned is certainly from EDR, but where the mistake is in the stack is unclear.

### Search terms

eth_sign, RPC, EDR

fvictorio:
[...]

PR #457, opened 2024-05-16
Fix eth_sign parameters order

fvictorio:
eth_sign was implemented as an alias of personal_sign. This was not correct, as the order of the parameters is different for both methods.

Closes https://github.com/NomicFoundation/edr/issues/455


agostbiro (approved):
Thanks, LGTM!

PR #460, opened 2024-05-21
ci: remove synthetic benchmarks

agostbiro:
Remove synthetic benchmarks as they've been superseded by scenario benchmarks.


.cargo-husky/hooks/pre-commit
```diff
@@ -5,7 +5,7 @@
 
 if [ -z "${ALCHEMY_URL}" ] || [ -z "${INFURA_URL}" ]; then
   echo WARNING: skipping remote tests because the ALCHEMY_URL and INFURA_URL environment variables are not defined. Local test results may differ from those in CI.
-  ALL_FEATURES="--features tracing,bench-once,serde,std"
+  ALL_FEATURES="--features tracing,serde,std"
 else
   ALL_FEATURES=--all-features
```
Wodann, line 10: You can also remove the `--profile bench` from line 17. This avoids changing profiles between the pre-commit hook and normal `cargo check`

Wodann (changes requested):
One request for change. Otherwise LGTM.

I merged the [other PR](https://github.com/NomicFoundation/edr/pull/461) without realising it was targetting this branch, so I messed up the merge ordering. You might need to split them or merge them as one.

.cargo-husky/hooks/pre-commit
agostbiro: Thanks good catch!

Wodann (approved):

PR #466, opened 2024-05-23
chore: fork Foundry Forge

agostbiro:
Adds a fork of Foundry's Forge as a library to the EDR workspace to serve as a basis of the Solidity tests features. As discussed, this is a clean fork, applying patches from upstream is not a prio. The forked commit is: https://github.com/foundry-rs/foundry/commits/0a5b22f07

Forge depends on a lot of crates within Foundry. These are included in this fork. The test suites for all these crates are included as well. Foundry Solidity build tools are necessary to run these tests, so these are also included.  

At the linked commit, the Foundry repo has 123275 lines of Rust code, this fork has 56243 lines of Rust code. I've only changed the included Foundry code to pass our lints and disable flaky tests.  I suspect that we can further remove a significant part of these 56k lines in the future. 

The Foundry repo uses workspace dependencies so I added these to our root Cargo.toml. Unifying these dependencies with EDR crates is a [follow up task.](https://github.com/NomicFoundation/edr/issues/467) The Foundry repo had a lot of customized compilation flags in their root Cargo.toml, I ported these to our root manifest as well to speed up compilation.


.cargo-husky/hooks/pre-commit
```diff
@@ -13,9 +13,7 @@ fi
 # Set -x  here to avoid printing secrets
 set -xe
 
-# Setting RUSTFLAGS env for clippy makes it not include custom rules
-RUSTFLAGS=-Dwarnings cargo check --workspace --all-targets ${ALL_FEATURES}
-cargo clippy --all --all-targets ${ALL_FEATURES} -- -D warnings
```
agostbiro, line 18: I've removed the separate `cargo check` as Clippy is running it anyway and it was triggering recompilation for some reason.

.cargo/config.toml
```diff
@@ -28,7 +28,6 @@ rustflags = [
     "-Wclippy::float_cmp_const",
     "-Wclippy::fn_params_excessive_bools",
     "-Wclippy::from_iter_instead_of_collect",
-    "-Wclippy::if-not-else",
```
agostbiro, line 31: I removed this lint as Foundry code has a pattern of checking for error conditions first with `if !invariant { error } else { ... }`. I think this is a reasonable pattern that we shouldn't prevent.

.gitignore
```diff
@@ -29,9 +29,6 @@ pids
 # Directory for instrumented libs generated by jscoverage/JSCover
 lib-cov
 
-# Coverage directory used by tools like istanbul
-coverage
```
agostbiro, line 33: Removed this as it excluded the `coverage` directory in the Foundry code and we don't use JS code coverage

.github/workflows/edr-ci.yml
```diff
@@ -46,12 +46,12 @@ jobs:
...
-        uses: actions-rs/cargo@v1
-        with:
-          command: hack
-          args: check --feature-powerset --no-dev-deps
+  
+  #      - name: Cargo hack
+  #        uses: actions-rs/cargo@v1
+  #        with:
+  #          command: hack
+  #          args: check --feature-powerset --no-dev-deps
```
agostbiro, line 54: This is temporary

.cargo/config.toml
Wodann: Why not just allow the pattern in the foundry crates?

Cargo.toml
```diff
@@ -1,14 +1,186 @@
...
+# Uncomment this section if you're using a debugger.
+debug = 1
 
+# Optimized release profile.
 [profile.release]
 rpath = true
+opt-level = 3
+debug = "line-tables-only"
+lto = "fat"
+strip = "debuginfo"
```
Wodann, line 44: Does this imply stripping of debug info? This would prohibit sampling profiling

crates/edr_napi/Cargo.toml
```diff
@@ -17,7 +17,7 @@ log = { version = "0.4.20", default-features = false }
 napi = { version = "2.16.0", default-features = false, features = ["async", "error_anyhow", "napi8", "serde-json"] }
 napi-derive = "2.16.0"
 edr_defaults = { version = "0.3.5", path = "../edr_defaults" }
```
Wodann, line 19: Is there a way to link these versions to the workspace too?

Cargo.toml
```diff
@@ -1,14 +1,186 @@
...
+bitvec.opt-level = 3
+
+# fuzzing
+proptest.opt-level = 3
+foundry-evm-fuzz.opt-level = 3
+
+# keystores
+scrypt.opt-level = 3
+
+[workspace.dependencies]
```
Wodann, line 84: Is there a follow-up to make the EDR crates use the workspace dependencies?

Wodann:
It's hard for me to evaluate the side-effects of all of the changes to the workspace's `Cargo.toml`. Do you have a full understanding of that? Are there any things to be aware of?

.cargo/config.toml
agostbiro: We're planning to heavily modify the foundry crates, and it's not great if there are different coding standards for different parts of the code base + I don't think the pattern is actually harmful.

Cargo.toml
agostbiro: Yep it's here: https://github.com/NomicFoundation/edr/issues/467

crates/edr_napi/Cargo.toml
agostbiro: Yep there is a follow up task: https://github.com/NomicFoundation/edr/issues/467

Cargo.toml
agostbiro: Good catch on this setting. I looked up the [default release params](https://doc.rust-lang.org/cargo/reference/profiles.html#release) and there it's `debug = false` actually, so the `debug` and `strip` flags are superfluous here and I've removed them.

Regarding sampling profiling, that needs full debug info, so we have to run it with an override currently as well: `CARGO_PROFILE_RELEASE_DEBUG=true cargo flamegraph ...`

agostbiro:
> It's hard for me to evaluate the side-effects of all of the changes to the workspace's `Cargo.toml`. Do you have a full understanding of that? Are there any things to be aware of?

Thanks for your review! The main thing to look out for is not to break `edr_napi` I think. This I've checked locally and the CI checks for the various supported platforms passed as well, so for me it's ok.

.gitignore
```diff
@@ -14,7 +14,6 @@ workspace.code-workspace
 # ignoring the node_modules part, as it'd ignore every node_modules, and we have some for testing
 
 # Logs
-logs
```
agostbiro, line 17: Removed this as it excluded the `logs` directory in Foundry test data. We can re-add it in subdirectories if it becomes necessary

.github/workflows/edr-ci.yml
```diff
@@ -47,11 +47,11 @@ jobs:
...
-      - name: Cargo hack
-        uses: actions-rs/cargo@v1
-        with:
-          command: hack
-          args: check --feature-powerset --no-dev-deps
+      - name: Cargo hack for EDR crates
+        run: |
+          chmod +x ./github/scripts/cargo-hack-edr.sh
+          ./github/scripts/cargo-hack-edr.sh
+        shell: bash
```
agostbiro, line 54: `cargo-hack`, the tool that makes sure that all feature combinations compile is failing for the Foundry crates. Since we aren't planning to publish them, this is ok, so I changed the CI to only run `cargo-hack` for EDR crates.

agostbiro:
> I think this has effects beyond publishing. In my normal workflow I always test syntax with `--all --no-default-features` and `--all --all-features` when doing `cargo check`. If either of those would now start failing due to foundry having invalid syntax, that would disrupt my workflow as well.

Right, `--all --all-features` is covered by clippy, but `--all --no-default-features` could've failed undetected. 

I addressed this in [75b1443](https://github.com/NomicFoundation/edr/pull/466/commits/75b1443475507a1b03d44d13da0958379aadb633) by splitting up the `cargo hack` check and now we run `cargo check --all --no-default-features` for all crates and `cargo hack check --feature-powerset --exclude-no-default-features --no-dev-deps` for EDR crates only.

.github/workflows/edr-ci.yml
agostbiro: Update: https://github.com/NomicFoundation/edr/pull/466#issuecomment-2132907024

Wodann:
Thanks!

Wodann (approved):

Issue #299, opened 2024-02-25
Make `hardhat-tracer` work with HH+EDR


fvictorio:
**Definition of done**

`hardhat-tracer` should work with HH+EDR as well as it works with HH+EJS. Performance is not significantly degraded when used with HH+EDR.

**To-do**

- [x] #331
- [x] #332
- [x] #333
- [x] #334
- [x] #297

fvictorio:
Done: https://github.com/zemse/hardhat-tracer/commit/ffe997494aa506ef11896a013a8ae4fa5c82a10c

PR #469, opened 2024-05-24
Add traces to `debug_traceCall` and `debug_traceTransaction`

fvictorio:


agostbiro (approved):
Thanks LGTM as a quick fix with the caveat, that I'd expect this change to cause a significant slow down in the `debug_trace*` methods due to doubling tracing work when `verbose_tracing = true` with `hardhat-tracer`.

Since neither `hardhat-tracer`, nor `debug_trace*` methods are covered by automated benchmarks, I don't know if we fulfill the 

> Performance is not significantly degraded when used with HH+EDR.

criterion from https://github.com/NomicFoundation/edr/issues/299

fvictorio:
Agree, but I think a potential slowdown in the `debug_` methods is acceptable, at least for now.

PR #474, opened 2024-05-27
test: execute solidity tests through napi

agostbiro:
Adds basic setup to execute Solidity tests with Forge from JS through Napi. The test contracts were compiled with Hardhat.

Traces from Forge are not exposed to JS through Napi. We probably want to unify these with the existing JS trace format first.

The implementation is currently tightly coupled with Foundry config format and build system. We'll probably want to remove this coupling next.


crates/edr_napi/src/solidity_tests/runner.rs
```diff
@@ -0,0 +1,147 @@
...
+        coverage: builder.coverage,
+        debug: builder.debug,
+        test_options: builder.test_options.unwrap_or_default(),
+        isolation: builder.isolation,
+        output: compiled,
+    })
+}
+
+fn project_root() -> PathBuf {
+    PathBuf::from(concat!(env!("CARGO_MANIFEST_DIR"), "/../../testdata"))
```
Wodann, line 59: I'm trying to wrap my head around this. My understanding is:

`CARGO_MANIFEST_DIR` = `./crates/edr_napi/`
Concatenated, you get: `./crates/edr_napi/../../testdata`
which equals: `./testdata`

However, `testdata` is found at `crates/foundry/testdata`.

What am I missing here? 😅

crates/edr_napi/src/solidity_tests/runner.rs
```diff
@@ -0,0 +1,147 @@
...
+        ffi: true,
+        verbosity: 3,
+        memory_limit: 1 << 26,
+        ..Default::default()
+    }
+}
+
+/// The RPC endpoints used during tests.
+fn rpc_endpoints() -> RpcEndpoints {
+    RpcEndpoints::new([("alchemy", RpcEndpoint::Url("${ALCHEMY_URL}".to_string()))])
```
Wodann, line 108: How does this handle the case where the `ALCHEMY_URL` environment variable is not set?

crates/edr_napi/src/solidity_tests.rs
```diff
@@ -0,0 +1,112 @@
...
+            tokio::sync::mpsc::unbounded_channel::<(String, forge::result::SuiteResult)>();
+        let (tx_end_result, mut rx_end_result) = tokio::sync::mpsc::unbounded_channel();
+
+        let callback_fn = self.results_callback_fn.clone();
+        let runtime = runtime::Handle::current();
+        runtime.spawn(async move {
+            let mut results = Vec::<(String, forge::result::SuiteResult)>::new();
+
+            while let Some(name_and_suite_result) = rx_results.recv().await {
+                results.push(name_and_suite_result.clone());
```
Wodann, line 70: If you'd like to avoid the overhead of cloning, you can explore a similar approach to the one I took in other parts of the N-API:

You can wrap the result in an `Arc` and return an `#[napi] struct SomeStruct { inner: Arc<..> }` which implements getter properties to retrieve the contents of the `inner` value.

crates/edr_napi/src/solidity_tests/test_suite.rs
```diff
@@ -0,0 +1,100 @@
...
+        })
+    }
+}
+
+/// A test contract to execute.
+#[derive(Clone)]
+#[napi(object)]
+pub struct TestContract {
+    /// The contract ABI as a JSON string.
+    pub abi: String,
```
Wodann, line 52: I'm writing this comment here, as this reminded me of it.

Hardhat uses a non-standard ABI for `console.log`. This might have unexpected consequences if we're compiling through Hardhat while using parts of `foundry` that make have expectations of the ABI.

Raising this for awareness, as it might have some impact on tests failing unexpectedly

cc: @fvictorio

crates/edr_napi/test/solidity-tests.ts
```diff
@@ -0,0 +1,58 @@
...
+    const resultsFromCallback: Array<SuiteResult> = [];
+    const runner = new SolidityTestRunner((...args) => {
+      resultsFromCallback.push(args[1] as SuiteResult);
+    });
+
+    const result = await runner.runTests(testSuites);
+    assert.equal(resultsFromCallback.length, testSuites.length);
+    assert.equal(resultsFromCallback.length, result.length);
+
+    for (let res of result) {
```
Wodann, line 19: Is there a simpler way to validate results other than doing a name comparison string?

Are the results guaranteed to be in the same order as the inputs?

Wodann (changes requested):
Overall LGTM. I left some suggestions, comments, and questions

crates/edr_napi/test/solidity-tests.ts
agostbiro: Nope, they'll be in the order of completion. Forge returns them as a BTreeMap, but can't expose that through NAPI afaik, so I opted for a list of `(name, value)` pairs to preserve the semantics. If it's not important though to preserve the order of completion, we could just return a map of `name => value` which would make it easy to query for specific test suites. cc @fvictorio

crates/edr_napi/src/solidity_tests.rs
agostbiro: Thanks that's a good idea, but I think it's ok to clone here, because it's not blocking test execution, so I'd keep it as is to keep it simple. A concern could be that the clone is blocking the Tokio worker thread, but I don't think it's big enough to warrant moving it to the blocking thread pool.

crates/edr_napi/src/solidity_tests.rs
```diff
@@ -0,0 +1,112 @@
...
+pub struct SolidityTestRunner {
+    /// The callback to call with the results as they become available.
+    results_callback_fn: ThreadsafeFunction<SuiteResult>,
+}
+
+// The callback has to be passed in the constructor because it's not `Send`.
+#[napi]
+impl SolidityTestRunner {
+    /// Creates a new instance of the SolidityTestRunner.
+    #[doc = "Creates a new instance of the SolidityTestRunner. The callback function will be called with suite results as they finish."]
```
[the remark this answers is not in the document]
agostbiro: Thanks, fixed in [95c365f](https://github.com/NomicFoundation/edr/pull/474/commits/95c365fd011235ac2eb6516899ba595c9b180c84)

crates/edr_napi/src/solidity_tests/runner.rs
agostbiro: It's ok currently because forking mode isn't used. Going forward it should be passed through a config object from JS

crates/edr_napi/src/solidity_tests/runner.rs
agostbiro: Thanks good catch the path was wrong! Fixed in [bf5cb47](https://github.com/NomicFoundation/edr/pull/474/commits/bf5cb47d15d1962f214e9e728557603032cd2595)

It didn't cause failures, because the contracts are passed in through JS, not loaded from disk

crates/edr_napi/test/solidity-tests.ts
Wodann: As a caller of the API, I think it would be more ergonomic if the order of the array of results was in the same order as the input array. That way you can easily iterate them side-by-side and there is no need for unwieldy code and expensive name comparisons.

I had a look at the code and it seems possible to achieve this by using an integer ID to assign a result to an index of a pre-allocated array matching the length of the input array. Alternatively, a safer alternative might be to use oneshot channels per test suite. That way you could `.collect` all results of those oneshot channels.

agostbiro: Thanks good points, but thinking about this more I think it's good to return from Rust as `(name, value)` pairs preserving the order of completion for maximum flexibility, as it's easy to turn that into any other desired format in JS (since JS knows the input order). Otoh JS can also know the order of completion from the callbacks, so maybe it's not important. Would like to hear @fvictorio 's feedback who is experimenting from the JS side.

crates/edr_napi/src/solidity_tests.rs
agostbiro: Ah got it, if there were traces included in the test results, I'd definitely do it, since those are huge. Without traces I think it's not so clear that it's worth it without measuring on a benchmark, so I wouldn't do it in this PR.

Wodann: Sounds good

agostbiro:
The test suite bytecode is built with Hardhat, but Foundry project compile was still executed to as the Forge test runner expected some output from that as input. This was causing problems in CI as solc wasn’t available on linux-musl. To resolve this I did a hacky fix [Disable Foundry project compilation](https://github.com/NomicFoundation/edr/pull/474/commits/dfb1e827acc5d328ff4a6f653507e14eb545069c) and added a follow up https://github.com/NomicFoundation/edr/issues/487.

crates/foundry/forge/src/lib.rs
```diff
@@ -238,4 +291,16 @@ impl TestOptionsBuilder {
         let base_invariant = self.invariant.unwrap_or_default();
         TestOptions::new(output, root, profiles, base_fuzz, base_invariant)
     }
+
+    /// Creates an instance of [`TestOptions`]. This takes care of creating
+    /// "fuzz" and "invariant" fallbacks, and extracting all inline test
+    /// configs, if available.
+    pub fn build_hardhat(self) -> Result<TestOptions, InlineConfigError> {
```
Wodann, line 298: How are we currently compiling inline tests?

Wodann (changes requested):
I had a look at the introduced workaround for Foundry compilation, but I lack understanding of the code to be able to validate its inner workings.

In its current state, it at least seems like the CI is still failing for invariant tests.

crates/foundry/forge/src/lib.rs
agostbiro: We don't currently, doc is just leftover. I'd keep it like this until the issue is more comprehensively resolved in the follow up

agostbiro:
> I had a look at the introduced workaround for Foundry compilation, but I lack understanding of the code to be able to validate its inner workings. 

It's just a super simple workaround for now to avoid triggering code paths that would require Foundry compilation of the project. More comprehensive solution to be done in https://github.com/NomicFoundation/edr/issues/487.

> In its current state, it at least seems like the CI is still failing for invariant tests.

Right, I went a bit trigger happy there, should be fixed now.

fvictorio (approved):

crates/edr_napi/src/solidity_tests.rs
```diff
@@ -0,0 +1,111 @@
...
+    pub async fn run_tests(&self, test_suites: Vec<TestSuite>) -> napi::Result<Vec<SuiteResult>> {
+        let test_suites = test_suites
+            .into_iter()
+            .map(|item| Ok((item.id.try_into()?, item.contract.try_into()?)))
+            .collect::<Result<Vec<_>, napi::Error>>()?;
+        let mut runner = build_runner(test_suites)?;
+
+        let (tx_results, mut rx_results) =
+            tokio::sync::mpsc::unbounded_channel::<(String, forge::result::SuiteResult)>();
+        let (tx_end_result, mut rx_end_result) = tokio::sync::mpsc::unbounded_channel();
```
Wodann, line 61: Question: I raised this in chat as well, but should this be an unbounded channel or a oneshot?

Wodann (approved):
One question. Otherwise LGTM.

crates/edr_napi/src/solidity_tests.rs
agostbiro: Yeah thanks, I've changed that locally and I'll push one more commit to fix the progress reporting only happening at the end. Will reping for review

agostbiro:
@Wodann @fvictorio I pushed a new commit, could you review it please: [Make sure callbacks are triggered as soon as a result is ready](https://github.com/NomicFoundation/edr/pull/474/commits/1cc21e8ecbfbf95cbb16eed444ec8029bb5c3e95)

I tested it manually in the benchmark branch.

Wodann (approved):

Wodann:
Thanks for the fix! Good find!

There are still some CI errors. Once those have been resolved, feel free to merge

PR #480, opened 2024-05-29
ci: add prettier md lint and fix style

agostbiro:
Adds CI lint for Markdown files with prettier + applies fixes.


crates/tools/python/README.md
```diff
@@ -13,4 +13,3 @@ conda env create -f env.yml
 conda activate hardhat
 jupyter notebook --NotebookApp.iopub_data_rate_limit=1.0e10
 ```
-
```
Wodann, line 16: Is this removing a line ending at end-of-file? Those should be retained imo

See https://stackoverflow.com/a/5813359 for reasoning

Wodann (changes requested):

crates/tools/python/README.md
agostbiro: There were two lines, prettier just removed the extra one

Wodann (approved):

PR #483, opened 2024-05-31
refactor: use generic trait bound for RpcClient

Wodann:
This PR splits off the `RpcClient` into two separate crates:
1. `edr_rpc_client`: contains a genericised version of `RpcClient` that accepts any type which implements the `trait RpcMethod`. As `RpcClient`'s caching mechanism requires knowledge about the block number and chain ID I added the `chain_id_request` and `block_number_request` methods to `trait RpcMethod`, alongside an associated type that specifies the `CacheableMethod` type.
The `RpcClient` type can be reused by chain types with varying JSON-RPC methods, as listed [here](https://www.notion.so/Multichain-overview-9c388d40a9f144f3b6ae400436c8926f?d=eb31dab817c94773a3539f4aecb47f3d&pvs=4#7b188e1e985a419aa3a500f89374675b).
2. `edr_rpc_eth`: introduces an `EthRpcClient` for all chains that support the [Ethereum L1 JSON-RPC specification](https://ethereum.github.io/execution-apis/). It allows variance in the `Block` and `Transaction` types, but this could be extended to `Receipt` or other types as necessary. It's main goal is to avoid duplication of `pub enum RequestMethod` between Optimism, Arbitrum, Ethereum L1.

Alternatively to the `block_number_request` and `chain_id_request` methods in **1**, they could be reworked to `block_number_without_caching` and `chain_id_without_caching`.


crates/edr_rpc_client/src/cache.rs
```diff
@@ -0,0 +1,133 @@
...
+pub trait CacheableMethod: Sized {
+    /// The type representing the cached method.
+    type Cached<'method>: CachedMethod + TryFrom<&'method Self>
+    where
+        Self: 'method;
+
+    /// Creates a method for requesting the block number.
+    fn block_number_request() -> Self;
+
+    /// Creates a method for requesting the chain ID.
```
agostbiro, line 33: This is a bit confusing. `eth_block_number` and `eth_chainId` are not cacheable

crates/edr_rpc_client/src/cache.rs
```diff
@@ -0,0 +1,133 @@
...
+};
+
+use serde::de::DeserializeOwned;
+
+pub use self::hasher::KeyHasher;
+use self::key::{ReadCacheKey, WriteCacheKey};
+use crate::RpcClientError;
+
+/// Trait for RPC method types that can be cached.
+pub trait CacheableMethod: Sized {
```
agostbiro, line 24: The trait `CacheableMethod` having `block_number_request` and `chain_id_request` methods is a bit confusing as `eth_blockNumber` and `eth_chainId` are not cacheable.

Wodann: Do you have an idea for a more applicable trait name?

The flow is as follows: 
- the method type enum that's passed into the `RpcClient` should be cacheable
- our caching solution needs to know the block number and chain ID, hence the existence of those two functions.
- a subset of the method type enum can actually be cached. This is called a `CachedMethod`

Wodann: One alternative name I can imagine is to
- rename `CacheableMethod` to `RpcMethod` and `CachedMethod` to `CacheableMethod`

Would that clarify things?

Wodann: I tried to clarify things in https://github.com/NomicFoundation/edr/pull/483/commits/b34068e14bf6e8e631106e5f412530a0a6b26944

agostbiro: Thanks the change makes sense!

agostbiro (changes requested):
Thanks LGTM overall, just one change request: 

Please use the same approach (unit vs integration test) for testing the client in `edr_rpc_client` and `edr_rpc_eth`, unless there is a good argument not to. I don't have strong views which is preferable here, but I like to default to unit tests due to easier code sharing.

Wodann:
> unless I'm missing a reason not to.

In the `edr_rpc_client` crate, the test functions use a private member to validate logic: `assert!(client.cached_block_number.read().await.is_some());`. `edr_rpc_eth` only uses the public API.

agostbiro:
> In the `edr_rpc_client` crate, the test functions use a private member to validate logic:
> 
> `assert!(client.cached_block_number.read().await.is_some());`
> 
> `edr_rpc_eth` only uses the public API in combination with 1) a mocking server; and 2) remote access to Alchemy. If you don't think that's a valid reason to consider them integration tests, please let me know.

Ok so there is a hard reason why client tests in `edr_rpc_client` need to be "unit tests". 

I can see why you implemented client tests as "integration tests" in `edr_rpc_eth` from a logical perspective, but from a practical perspective I think this differentiation will be annoying, because often changes to one crate will need changes in the other and the different testing pattern can trip us up. So I'd prefer to have client tests in `edr_rpc_eth` as "unit tests" unless them being "integration tests" brings a practical benefit (since the crates are internal I can't think of any, but happy to be convinced otherwise).

Wodann:
The main benefit I see is separation of the tests into dedicated test files, instead of either being interspersed with source files or resulting in incredibly long source files with a `test` submodule.

I'm happy to move the tests into the client.rs source file if it hampers your dev experience, though.

agostbiro:
Awesome thanks! We can move the tests to a submodule `client/tests.rs` if it gets too large.

agostbiro (approved):

PR #485, opened 2024-05-31
ci: add workflows for PR reviews slack notifications

fvictorio:
These workflows will notify us on Slack when a PR review is requested or submitted. They use the jobs defined here: https://github.com/NomicFoundation/github-actions-workflows/tree/main/.github/workflows

### DIFF
--- a/.github/workflows/review-requested-slack-notification.yml
+++ b/.github/workflows/review-requested-slack-notification.yml
@@ -1,0 +1,10 @@
+name: PR review requested Slack notification
+
+on:
+  pull_request_target:
+    types: [review_requested]
+
+jobs:
+  requested:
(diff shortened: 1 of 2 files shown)


Wodann (commented):
Is this necessary, given that we already get GitHub notifications?

fvictorio:
I would personally like it, but maybe we can make it opt-in?

cc @alcuadrado

alcuadrado:
It's opt-out within slack. You just need to mute the conversation with "Github review notifications".

alcuadrado:
> It's opt-out within slack. You just need to mute your conversation with "Github review notifications".

This doesn't work very well. You still get a red circle in the conversation. If you want, I can remove you from the workflow at the org level.

Personally, I haven't felt them as duplication, as these are much more specific than GitHub notifications, more like a call to action, while GH notifications keep me informed about everything.

Wodann (approved):

PR #493, opened 2024-06-05
refactor: add generic trait bound to ExecutableTransaction

Wodann:
This PR is intended as an incremental change towards chain-specific blockchains. In particular, it adds a generic trait bound to `ExecutableTransaction`. For now, we always use the `L1ChainSpec` at compile-time.

### DIFF
--- a/crates/edr_evm/src/chain_spec.rs
+++ b/crates/edr_evm/src/chain_spec.rs
@@ -23,7 +23,8 @@ impl ChainSpec for L1ChainSpec {
     type SignedTransaction = edr_eth::transaction::Signed;
 }
 
-// TODO: This is a temporary solution until the revm fork has been merged. In
+// TODO: https://github.com/NomicFoundation/edr/issues/495
+// This is a temporary solution until the revm fork has been merged. In
 // the fork each chain type has its own transaction type with an accompanying
 // trait.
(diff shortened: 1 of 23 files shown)


crates/edr_evm/src/chain_spec.rs
```diff
@@ -0,0 +1,46 @@
...
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct L1ChainSpec;
+
+impl ChainSpec for L1ChainSpec {
+    type SignedTransaction = edr_eth::transaction::Signed;
+}
+
+// TODO: This is a temporary solution until the revm fork has been merged. In
+// the fork each chain type has its own transaction type with an accompanying
+// trait.
```
agostbiro, line 28: Nice to have: if there is an issue/PR for this it'd be good to link here.

agostbiro (approved):
Thanks LGTM, with one nice to have.

crates/edr_evm/src/chain_spec.rs
Wodann: Addressed in https://github.com/NomicFoundation/edr/pull/493/commits/ca51139abae53f862413947ab1b789b95550b2b5

PR #494, opened 2024-06-06
refactor: Provide helpful error msg when deserializing big timestamp

### LINKED ISSUE #435: Improve error message for large block timestamps
See https://github.com/NomicFoundation/edr/issues/399

Instead of

```
data did not match any variant of untagged enum U64OrUsize at line 1 column 110
```

We should show a better error, something like `Timestamp used for {method} is too large. Received {value}; the maximum value is {max_value}.` (actual message tbd).

Xanewok:
Fixes #435 

In place of the existing `U64OrUsize` we provide a newtype `Timestamp` around `u64` with a custom Deserialize impl that leniently first deserializes to a number/string and then attempts to parse the value to see if it's within the valid range.

We keep providing `From` and `Into` for the new type, so the existing code keeps using the impls whenever possible.

### DIFF
--- a/crates/edr_provider/src/requests/serde.rs
+++ b/crates/edr_provider/src/requests/serde.rs
@@ -351,7 +351,7 @@ where
 ///
 /// While technically the Yellow Paper allows specifying the timestamp as
 /// any scalar value, using `u64` in practice is more than enough.
-#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Serialize)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Serialize)]
 pub struct Timestamp(pub u64);
 
 impl From<u64> for Timestamp {
@@ -367,7 +367,7 @@ impl From<Timestamp> for u64 {
 }
 
 impl Timestamp {
-    pub fn as_u64(&self) -> u64 {
+    pub fn as_u64(self) -> u64 {
         self.0
     }
 }
@@ -387,17 +387,22 @@ where
 {
     let error_msg = || {
         serde::de::Error::custom(
-            "Timestamp must be a positive integer or a string not exceeding 2^64.",
+            "Timestamp must be a positive integer or a (`0x`-prefixed hex) string not exceeding 2^64.",
         )
     };
 
     match serde_json::Value::deserialize(deserializer)? {
+        // Unlike other JSON-RPC types, timestamp can also be a number
         serde_json::Value::Number(num) => num.as_u64().ok_or_else(error_msg),
-        serde_json::Value::String(value) => {
-            if let Some(hex_str) = value.strip_prefix("0x") {
-                u64::from_str_radix(hex_str, 16).map_err(|_err| error_msg())
+        serde_json::Value::String(ref value) => {
+            if let Some(value) = value.strip_prefix("0x") {
+                u64::from_str_radix(value, 16).map_err(|_err| error_msg())
             } else {
-                value.parse::<u64>().map_err(|_err| error_msg())
+                // symmetrical and more explicit about the base 10
+                #[allow(clippy::from_str_radix_10)]
+                // Since it exceptionally can also be a number, be more relaxed
+                // and accept stringified numbers as well (base 10 by default).
+                u64::from_str_radix(value, 10).map_err(|_err| error_msg())
             }
         }
         _ => Err(error_msg()),
@@ -568,19 +573,19 @@ mod tests {
             serde_json::from_str::<Timestamp>("1000000000000000000000000000000")
                 .unwrap_err()
                 .to_string(),
-            "Timestamp must be a positive integer or a string not exceeding 2^64."
+            "Timestamp must be a positive integer or a (`0x`-prefixed hex) string not exceeding 2^64."
         );
         assert_eq!(
             serde_json::from_str::<Timestamp>("\"1000000000000000000000000000000\"")
                 .unwrap_err()
                 .to_string(),
-            "Timestamp must be a positive integer or a string not exceeding 2^64."
+            "Timestamp must be a positive integer or a (`0x`-prefixed hex) string not exceeding 2^64."
         );
         assert_eq!(
             serde_json::from_str::<Timestamp>("\"0x11223344556677889900\"")
                 .unwrap_err()
                 .to_string(),
-            "Timestamp must be a positive integer or a string not exceeding 2^64."
+            "Timestamp must be a positive integer or a (`0x`-prefixed hex) string not exceeding 2^64."
         );
     }
 }
--- a/crates/edr_provider/src/requests/serde.rs
+++ b/crates/edr_provider/src/requests/serde.rs
@@ -377,35 +377,14 @@ impl<'de> Deserialize<'de> for Timestamp {
     where
         D: Deserializer<'de>,
     {
-        deserialize_timestamp_u64(deserializer).map(Timestamp)
-    }
-}
-
-fn deserialize_timestamp_u64<'de, D>(deserializer: D) -> Result<u64, D::Error>
-where
-    D: Deserializer<'de>,
-{
-    let error_msg = || {
-        serde::de::Error::custom(
-            "Timestamp must be a positive integer or a (`0x`-prefixed hex) string not exceeding 2^64.",
-        )
-    };
-
-    match serde_json::Value::deserialize(deserializer)? {
-        // Unlike other JSON-RPC types, timestamp can also be a number
-        serde_json::Value::Number(num) => num.as_u64().ok_or_else(error_msg),
-        serde_json::Value::String(ref value) => {
-            if let Some(value) = value.strip_prefix("0x") {
-                u64::from_str_radix(value, 16).map_err(|_err| error_msg())
-            } else {
-                // symmetrical and more explicit about the base 10
-                #[allow(clippy::from_str_radix_10)]
-                // Since it exceptionally can also be a number, be more relaxed
-                // and accept stringified numbers as well (base 10 by default).
-                u64::from_str_radix(value, 10).map_err(|_err| error_msg())
-            }
-        }
-        _ => Err(error_msg()),
+        // TODO: Accept only `0x`-prefixed hex strings (and possibly base 10 strings).
+        U64::deserialize(deserializer)
+            .map(|value| Timestamp(value.as_limbs()[0]))
+            .map_err(|_| {
+                serde::de::Error::custom(
+                    "Timestamp must be a positive integer or a string not exceeding 2^64 - 1",
+                )
+            })
     }
 }
 
@@ -567,25 +546,26 @@ mod tests {
     fn deserialize_timestamp() {
         serde_json::from_str::<Timestamp>("12345").unwrap();
         serde_json::from_str::<Timestamp>("\"12345\"").unwrap();
+        serde_json::from_str::<Timestamp>("\"0x0\"").unwrap();
         serde_json::from_str::<Timestamp>("\"0x1122334455667788\"").unwrap();
 
         assert_eq!(
             serde_json::from_str::<Timestamp>("1000000000000000000000000000000")
                 .unwrap_err()
                 .to_string(),
-            "Timestamp must be a positive integer or a (`0x`-prefixed hex) string not exceeding 2^64."
+            "Timestamp must be a positive integer or a string not exceeding 2^64 - 1"
         );
         assert_eq!(
             serde_json::from_str::<Timestamp>("\"1000000000000000000000000000000\"")
                 .unwrap_err()
                 .to_string(),
-            "Timestamp must be a positive integer or a (`0x`-prefixed hex) string not exceeding 2^64."
+            "Timestamp must be a positive integer or a string not exceeding 2^64 - 1"
         );
         assert_eq!(
             serde_json::from_str::<Timestamp>("\"0x11223344556677889900\"")
                 .unwrap_err()
                 .to_string(),
-            "Timestamp must be a positive integer or a (`0x`-prefixed hex) string not exceeding 2^64."
+            "Timestamp must be a positive integer or a string not exceeding 2^64 - 1"
         );
     }
 }
--- a/crates/edr_provider/src/requests/serde.rs
+++ b/crates/edr_provider/src/requests/serde.rs
@@ -347,6 +347,63 @@ where
     Ok(result)
 }
 
+/// Value representing timestamp in seconds.
+///
+/// While technically the Yellow Paper allows specifying the timestamp as
+/// any scalar value, using `u64` in practice is more than enough.
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Serialize)]
+pub struct Timestamp(pub u64);
+
+impl From<u64> for Timestamp {
+    fn from(value: u64) -> Self {
+        Self(value)
+    }
+}
+
+impl From<Timestamp> for u64 {
+    fn from(value: Timestamp) -> Self {
+        value.0
+    }
+}
+
+impl Timestamp {
+    pub fn as_u64(&self) -> u64 {
+        self.0
+    }
+}
+
+impl<'de> Deserialize<'de> for Timestamp {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        deserialize_timestamp_u64(deserializer).map(Timestamp)
+    }
+}
+
+fn deserialize_timestamp_u64<'de, D>(deserializer: D) -> Result<u64, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let error_msg = || {
+        serde::de::Error::custom(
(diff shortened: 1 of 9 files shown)


crates/edr_provider/src/requests/serde.rs
```diff
@@ -347,6 +347,63 @@ where
...
+}
+
+impl From<Timestamp> for u64 {
+    fn from(value: Timestamp) -> Self {
+        value.0
+    }
+}
+
+impl Timestamp {
+    pub fn as_u64(&self) -> u64 {
```
Xanewok, line 370: It's not used but provided it for clarity; can remove it if you'd like per YAGNI

crates/edr_provider/src/requests/serde.rs
```diff
@@ -347,6 +347,63 @@ where
     Ok(result)
 }
 
+/// Value representing timestamp in seconds.
+///
+/// While technically the Yellow Paper allows specifying the timestamp as
+/// any scalar value, using `u64` in practice is more than enough.
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Serialize)]
+pub struct Timestamp(pub u64);
```
Xanewok, line 355: Let me know if you'd like me to define this in the upper module; maybe it'd make more sense to have it there but we already re-export a type from here so I wasn't sure.

crates/edr_provider/src/requests/serde.rs
```diff
@@ -347,6 +347,63 @@ where
...
+    };
+
+    match serde_json::Value::deserialize(deserializer)? {
+        serde_json::Value::Number(num) => num.as_u64().ok_or_else(error_msg),
+        serde_json::Value::String(value) => {
+            if let Some(hex_str) = value.strip_prefix("0x") {
+                u64::from_str_radix(hex_str, 16).map_err(|_err| error_msg())
+            } else {
+                value.parse::<u64>().map_err(|_err| error_msg())
+            }
```
Wodann, line 401: This deviates from what used to happen. We don't want to allow any type of string here, but one that conforms to the Ethereum format: `^0x([1-9a-f]+[0-9a-f]*|0)$` (also see [this](https://github.com/NomicFoundation/hardhat/blob/ethereumjs/issue4970/packages/hardhat-core/src/internal/core/jsonrpc/types/base-types.ts#L208)).

I recommend using the `U64` type to deserialize the string here, as that already implements this logic. All you'd have to do is to first check the number of bytes in the hex string after `0x` (this part is required). If it exceeds 8 bytes, then return the custom error.

crates/edr_provider/src/requests/serde.rs
```diff
@@ -500,4 +557,18 @@ mod tests {
 
         assert_eq!(parsed.n, n);
     }
+
+    #[test]
+    fn deserialize_timestamp() {
+        serde_json::from_str::<Timestamp>("12345").unwrap();
+        serde_json::from_str::<Timestamp>("1000000000000000000000000000000").unwrap_err();
+        serde_json::from_str::<Timestamp>("\"1000000000000000000000000000000\"").unwrap_err();
```
Wodann, line 565: It would be good to validate which error is returned here.

crates/edr_provider/tests/eth_request_serialization.rs
```diff
@@ -3,10 +3,10 @@ mod common;
 use edr_eth::{
     filter::{LogFilterOptions, LogOutput, OneOrMore},
     transaction::EthTransactionRequest,
-    Address, BlockSpec, BlockTag, Bytes, PreEip1898BlockSpec, B256, U256, U64,
+    Address, BlockSpec, BlockTag, Bytes, PreEip1898BlockSpec, B256, U256,
 };
 use edr_evm::alloy_primitives::U160;
-use edr_provider::{IntervalConfigRequest, MethodInvocation, U64OrUsize};
```
Wodann, line 9: Is `U64OrUsize` still used anywhere else? If not, feel free to remove it.

Wodann (changes requested):

crates/edr_provider/tests/eth_request_serialization.rs
Xanewok: It's already removed in https://github.com/NomicFoundation/edr/pull/494/files#diff-f92267848751eff8250755669ad8dbb49e569a42f27bcdca2589fad3e50098a0L513

crates/edr_provider/src/requests/serde.rs
Xanewok: I'm also seeing some more inconsistencies, for example we only have `Option<u64>` using default Deserialize in `AccountOverrideOptions` for `eth_call` https://github.com/NomicFoundation/edr/blob/eb00b1533b75134cc453661f22e70d8f31add309/crates/edr_rpc_eth/src/override.rs#L18

but the [official API schema](https://ethereum.github.io/execution-apis/api-documentation/) says it accepts only a hex string of the quantity format you mentioned:

image

Maybe it'd make sense to separate a work item of "go over all of our (de)serialize logic and make sure we're compliant"?

On the other hand per the common networking adage of "be conservative in what you send, be liberal in what you accept" and the fact that we're supposed to be a development platform, I'm not sure if us not being as strict as the spec is necessary the bad thing here.

I'll add it for the agenda for tomorrow's sync; maybe @fvictorio you have some more context or opinions on this.

Xanewok:
Reshuffled the tests to be first (to double-check what we do on `main` already) and force-pushed with expanded `deserialize_timestamp` test; hope you don't mind :pray:

crates/edr_provider/tests/eth_request_serialization.rs
Wodann: My bad. Thanks!

crates/edr_provider/src/requests/serde.rs
Wodann: The example you wrote is valid. `0x01` is a single byte equal to 1. There are no leading zeroes.

Wodann: > I'm also seeing some more inconsistencies, for example we only have `Option<u64>` using default Deserialize in `AccountOverrideOptions` for `eth_call`
> 
> https://github.com/NomicFoundation/edr/blob/eb00b1533b75134cc453661f22e70d8f31add309/crates/edr_rpc_eth/src/override.rs#L18
> 
> but the [official API schema](https://ethereum.github.io/execution-apis/api-documentation/) says it accepts only a hex string of the quantity format you mentioned:
> 
> Maybe it'd make sense to separate a work item of "go over all of our (de)serialize logic and make sure we're compliant"?
> 
> On the other hand per the common networking adage of "be conservative in what you send, be liberal in what you accept" and the fact that we're supposed to be a development platform, I'm not sure if us not being as strict as the spec is necessary the bad thing here.
> 
> I'll add it for the agenda for tomorrow's sync; maybe @fvictorio you have some more context or opinions on this.

There is a special deserialize function used in the linked file:

```rust
    #[serde(
        default,
        skip_serializing_if = "Option::is_none",
        with = "edr_eth::serde::optional_u64"
    )]
    /// Account nonce override.
    pub nonce: Option<u64>,
```
That uses `U64`'s deserialization under the hood.

Xanewok: > The example you wrote is valid. 0x01 is a single byte equal to 1. There are no leading zeroes.

But that's not true, right? It:
- neither matches the `^0x([1-9a-f]+[0-9a-f]*|0)$` regexp you pasted and what's in the spec (`0x` followed by either a single `0` or a non-zero amount of `1-f` followed by a possibly zero amount of `0-f`), 
- nor the value seems to be concerned about the bytes per se, specifically `01` in base 10 has leading zeroes in the same sense `01` has as well in base 16. I didn't find a rule that requires us to zero-pad us to whole bytes (and the regex from the spec straight up invalidates it)

> They have additional logic on top of the FromStr implementation.

If I'm not missing something, they seem to handle null strings but other than that, in the `visit_str` they just call `FromStr::from_str` implicitly via `str::parse::<F: FromStr>`, so back to the linked function above and again: `0x01` seems to be converted correctly with the leading zero.

> That uses U64's deserialization under the hood.

Ah, I missed the attributes above :facepalm: I'm used to the docs being at the top and so I missed that when visually scanning, sorry about that.

On a side-note, why doesn't we just use `Option<U64>` with the default serde impls? For convenience, to only be concerned with primitive `u64` after deserializing?

Wodann: I think you're right that the regex pattern is wrong.

A single byte is always a pair of two hexadecimal numbers. Hence `01` or `0f` is a single byte. The first of the two characters in a non-zero byte is not considered to be an individual zero. Only `00` is.

Xanewok: We discussed synchronously and we initially agreed that:
1. we will attempt to follow spec (disallow leading zeroes) and only deviate if it brings enough benefit and is supported by Geth (cc @fvictorio)
2. we should do this outside of this PR
3. Geth seems to be parsing hex strings with full bytes, so that's another source of confusion here

Xanewok: @Wodann are you sure? This would also accept `0X` prefix as well as octal `0o`/`0O` and binary `0b`/`0B` which I'm pretty sure we do not want to accept, given that we accept to conform only to the spec `0x`: https://github.com/recmo/uint/blob/7b8b0d313a4960091c07eb3356c6a2512a46b585/src/string.rs#L200-L210

Xanewok: I don't want to prolong this but since we're already splitting hairs here, it's just worth noting that it'd not be consistent with what we do elsewhere, which is `String::deserialize`, then check for `0x` and then call `U64::from_str` (on a value that's guaranteed to start with `0x`, so okay).

Maybe just being consistent with the rest of the code would be okay, where we do the flow outlined above like everywhere else, and we could focus on compliance/correctness in a separate PR? WDYT?

Xanewok: Right, I forgot that we support direct numbers, which is where your suggestion for `U64::deserialize` came from, but we should not accept the different modes, so I tried to make it more explicit what we support and why in the code and in the error message.

Since we also accept JS numbers as the extension to the regular JSON-RPC API, it'd be weird not to support them as a regular string (so also a base 10 number). I've added a comment and made it more explicit about the base, let me know what you think.

crates/edr_provider/src/requests/serde.rs
```diff
@@ -351,7 +351,7 @@ where
 ///
 /// While technically the Yellow Paper allows specifying the timestamp as
 /// any scalar value, using `u64` in practice is more than enough.
-#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Serialize)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Serialize)]
```
Xanewok, line 354: drive-by - since it's 8 bytes on 64-bit machine (our only targets?) it's cheaper to copy it around.

crates/edr_provider/src/requests/serde.rs
```diff
@@ -377,35 +377,14 @@ impl<'de> Deserialize<'de> for Timestamp {
...
-                // symmetrical and more explicit about the base 10
-                #[allow(clippy::from_str_radix_10)]
-                // Since it exceptionally can also be a number, be more relaxed
-                // and accept stringified numbers as well (base 10 by default).
-                u64::from_str_radix(value, 10).map_err(|_err| error_msg())
-            }
-        }
-        _ => Err(error_msg()),
+        // TODO: Accept only `0x`-prefixed hex strings (and possibly base 10 strings).
+        U64::deserialize(deserializer)
```
Xanewok, line 381: @Wodann like we discussed, this uses `U64::deserialize` to keep using the old logic. We can revisit fixing this and other deserialization logic in a follow-up PR.

crates/edr_provider/src/requests/serde.rs
```diff
@@ -347,6 +347,47 @@ where
...
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        // TODO: Accept only `0x`-prefixed hex strings (and possibly base 10 strings).
+        U64::deserialize(deserializer)
+            .map(|value| Timestamp(value.as_limbs()[0]))
+            .map_err(|_| {
+                serde::de::Error::custom(
+                    "Timestamp must be a positive integer or a string not exceeding 2^64 - 1",
```
fvictorio, line 385: I'd say:

```suggestion
                    "Timestamp must be a non-negative number not exceeding 2^64 - 1",
```

- Non-negative, because `0` is accepted by `evm_increaseTime` (and it's probably fine to accept it).
- "Number" instead of "positive integer or a string" for conciseness. The fact that we accept both numbers and strings doesn't seem super relevant here.

hardhat-tests/test/internal/hardhat-network/provider/modules/evm.ts
```diff
@@ -146,6 +146,21 @@ describe("Evm module", function () {
...
+            "evm_increaseTime",
+            ["10000000000000000000000000000000"],
+            "Timestamp must be a positive integer or a string not exceeding 2^64 - 1"
+          );
+          await assertInvalidArgumentsError(
+            this.provider,
+            "evm_increaseTime",
+            ["0x1111111111111111111111111111111111111"],
+            "Timestamp must be a positive integer or a string not exceeding 2^64 - 1"
+          );
```
fvictorio, line 162: Can we add 4 test cases more here?

1. Using exactly `2^64` produces the error too (using both hex-prefixed strings and decimal numbers)
2. Using exactly `2^64-1` doesn't produce the error (both hex-prefixed and decimal)

crates/edr_provider/src/requests/serde.rs
fvictorio: I only skimmed this discussion (let me know if there's something in particular I should pay attention to) but this caught my eye:

> This would also accept 0X prefix as well as octal 0o/0O and binary 0b/0B which I'm pretty sure we do not want to accept

I checked and the released version of EDR accepts all of these (which is wrong by the spec, and geth doesn't accept):

```
send("eth_getBlockByNumber", [0, false])
send("eth_getBlockByNumber", ["0", false])
send("eth_getBlockByNumber", ["0b0", false])
send("eth_getBlockByNumber", ["0o0", false])
```

I will open an issue about this, but we don't need to do anything about that in this PR.

fvictorio: https://github.com/NomicFoundation/edr/issues/509

hardhat-tests/test/internal/hardhat-network/provider/modules/evm.ts
Xanewok: I did that for the `evm_setNextBlockTimestamp`.

However, I now realized that `evm_increaseTime` actually internally converts the value to an `i64`... Does that mean we should support negative as well? This also means supporting up to 2**63-1. Is this something we want to pursue as part of this PR?

Wodann: IIRC, `evm_increaseTime` was allowed to receive negative values in the past, but this was a bug. We fixed that in EDR.

The reason that we use an `i64` in `ProviderData` is that the `fn block_time_offset_seconds` can return a negative value, if the first block is mined in the past.

Potentially we want to change that in the future and return an error. This might have been designed this way in Hardhat on purpose though.

I'd refer to @fvictorio for historic context.

crates/edr_provider/src/requests/serde.rs
Xanewok: Done in https://github.com/NomicFoundation/edr/pull/494/commits/913edb7a511e996113c84e84560b13297029346a

hardhat-tests/test/internal/hardhat-network/provider/modules/evm.ts
fvictorio: > Potentially we want to change that in the future and return an error.

No, this is not an error condition, it happens all the time. I don't remember exactly why; I think it's because in Hardhat the config is eagerly loaded (which then uses `now()` for the initial timestamp) but the node is created lazily after the first request, so normally the time offset will be `-N`, where `N` is the number of seconds that happened between the start of the task and the first request.

I'm not sure about the `i64` part, but I'm happy with just having those tests for `evm_setNextBlockTimestamp`, so no need to block on that.

fvictorio (approved):

hardhat-tests/test/internal/hardhat-network/provider/modules/evm.ts
Xanewok: It looks the situation may not be obvious on the `i64` front, so I'll wait for @Wodann here. When everyone agrees, I think we could finally merge this.

Wodann (approved):

hardhat-tests/test/internal/hardhat-network/provider/modules/evm.ts
Wodann: Approved.

Xanewok: Thanks!

PR #498, opened 2024-06-06
refactor: use enum to distinguish transaction signature types

Wodann:
This PR introduces a way to statically distinguish between a fake signature and an ECDSA signature with recoverable address.

The introduced `signature::Fakeable<SignatureT>` also handles caching of the address to ensure that it only will be recovered once.

These changes allow the future removal of the `ExecutableTransaction`, as the caller address will now always be part of `transaction::Signed`


crates/edr_eth/src/signature/fakeable.rs
```diff
@@ -0,0 +1,217 @@
...
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        use serde::ser::SerializeMap;
+
+        let mut map = serializer.serialize_map(Some(3))?;
+        map.serialize_entry("r", &self.r())?;
+        map.serialize_entry("s", &self.s())?;
+        // We serialize the `v` field for backwards compatibility
```
agostbiro, line 159: Could you expand on what needs to be backwards compatible?

crates/edr_eth/src/signature/fakeable.rs
```diff
@@ -0,0 +1,217 @@
...
+                    }
+                    .encode(out);
+                } else {
+                    let ecdsa = Ecdsa {
+                        r: self.r(),
+                        s: self.s(),
+                        v: *v,
+                    };
+
+                    println!("ecdsa: {ecdsa:?}");
```
agostbiro, line 115: Nit: this might be a debugging leftover

crates/edr_eth/src/signature/fakeable.rs
```diff
@@ -0,0 +1,217 @@
...
+    fn length(&self) -> usize {
+        match &self.data {
+            FakeableData::Fake { v } => {
+                let v_length = if let Some(y_parity) = self.y_parity() {
+                    y_parity.length()
+                } else {
+                    v.length()
+                };
+
+                self.r().length() + self.s().length() + v_length
```
agostbiro, line 133: This might be more bullet proof if it proxied to length of`Ecdsa/EcdsaWithYParity`

crates/edr_eth/src/signature/fakeable.rs
```diff
@@ -0,0 +1,217 @@
...
+            }
+            FakeableData::Recoverable { signature } => signature.length(),
+        }
+    }
+}
+
+impl<SignatureT: Recoverable + Signature> PartialEq for Fakeable<SignatureT> {
+    fn eq(&self, other: &Self) -> bool {
+        self.r() == other.r() && self.s() == other.s() && self.v() == other.v()
```
agostbiro, line 142: What's the reason for the custom implementaton?

crates/edr_eth/src/signature.rs
```diff
@@ -98,23 +47,10 @@ pub enum SignatureError {
...
-    Data(Vec<u8>),
-    /// Message hash
-    Hash(B256),
-}
-
 #[derive(Clone, Copy, Debug, Eq, PartialEq, Hash)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 /// An ECDSA signature
-pub struct Signature {
+pub struct Ecdsa {
```
agostbiro, line 53: The name ECDSA is a bit weird for me for a signature struct, as ECDSA is the algorithm and the signature is its output, but I see that it's used as `signature::Ecdsa` so that makes it unambiguous, so not a change request, just feedback.

crates/edr_eth/src/transaction/signed/eip155.rs
```diff
@@ -23,17 +23,13 @@ pub struct Eip155 {
     pub kind: TxKind,
     pub value: U256,
     pub input: Bytes,
-    pub signature: Signature,
+    #[cfg_attr(feature = "serde", serde(flatten))]
+    pub signature: signature::Fakeable<signature::Ecdsa>,
```
agostbiro, line 27: I'm wondering why `signature::Fakeable<signature::Ecdsa>` and not just `signature::Ecdsa`?

agostbiro (commented):
Thanks for the PR! This is an important improvement. I went over it and left some comments on the code.

In general there are two conceptual things that I'm not 100% convinced of, but I also don't think they're merge blockers:

### `EcdsaWithYParity` struct

The `EcdsaWithYParity` struct is a bit weird from a conceptual perspective, because the y parity can be always recovered from the `v` value which the `Ecdsa` struct has. As I understand the distinction comes from the fact that EIP-2930 and later transactions only include the `r` and `s` values and an `odd_y_parity` flag, as opposed to `r`, `s` and `v` values. For the internal library of an Ethereum specific tool I think it's ok if this leaks into the signature types, but ideally we'd only have a `signature::Ecdsa` struct where we set the `v` value based on the `odd_y_parity` and `chain_id` from the transaction.

### `Signature` trait 

My understanding is that the purpose of the `Signature` trait is to provide a sum type over certain types in the `edr_eth` crate. I tend to prefer designs where sum types are  represented with enums as this brings more conceptual clarity and prevents type erasure and dynamic dispatch.

crates/edr_eth/src/transaction/signed/eip155.rs
agostbiro: On a second thought I think I understand now, but it's a bit weird to have a fakeable field in what is an RPC type (and transactions from RPC should definitely not be fake). Feels like we need two signed transaction types: one unverified (which comes from user input or we fetch it from RPC) and one verified where we can bypass the verification with fake signatures.

Wodann:
> ### `EcdsaWithYParity` struct
> 
> The `EcdsaWithYParity` struct is a bit weird from a conceptual perspective, because the y parity can be always recovered from the `v` value which the `Ecdsa` struct has. As I understand the distinction comes from the fact that EIP-2930 and later transactions only include the `r` and `s` values and an `odd_y_parity` flag, as opposed to `r`, `s` and `v` values. For the internal library of an Ethereum specific tool I think it's ok if this leaks into the signature types, but ideally we'd only have a `signature::Ecdsa` struct where we set the `v` value based on the `odd_y_parity` and `chain_id` from the transaction.

I agree that the `odd_y_parity` can be calculated from the `v` value. However, the RLP-encoding and -decoding of EIP-2930 and later transactions is different as well, given that the `odd_y_parity` is only 1 byte, as opposed to the 8-bytes that legacy and EIP-155 transactions RLP-encode/-decode.

As this distinction is known at compile-time, I expressed that with two different types: `Ecdsa` and `EcdsaWithYParity`. Previously this was achieved by having the `r`, `s,`, and `odd_y_parity` fields listed separately in the `Eip2930` and later transactions, whereas the `Legacy` and `Eip155` transactions included the `Signature`.

> ### `Signature` trait
> 
> My understanding is that the purpose of the `Signature` trait is to provide a sum type over certain types in the `edr_eth` crate. I tend to prefer designs where sum types are represented with enums as this brings more conceptual clarity and prevents type erasure and dynamic dispatch.

I'm not trying to achieve a sum type (unless you consider polymorphism a form of sum type). Given that there are two compile-time types `Ecdsa` and `EcdsaWithYParity` (see above argumentation) that contain the `r`, `s`, and `v` or `odd_y_parity` values, to achieve both compile-time and runtime polymorphism, I introduced the trait that returns those values.

It's [hard to say](https://stackoverflow.com/questions/28621980/what-are-the-actual-runtime-performance-costs-of-dynamic-dispatch) whether dynamic dispatch would be faster or slower than having a sum type as the compiler will potentially apply optimisations in both cases.

Given that there is only one location at which I'm using the trait, I can easily remove it by exposing the `r`, `s`, `v`, and `odd_y_parity` values directly on the `transaction::Signed`. I'm happy to do that to simplify the code.

crates/edr_eth/src/signature/fakeable.rs
Wodann: As the address is cached in a `OnceCell` that should not be incorporated in the comparison. Validating the actual signature suffices. I achieved that by comparing the `r`, `s,` and `v` values.

You pointing this out, made me realise that calculating those values has some overhead, so I separated the comparison of `Fake` and `Recoverable` in [this commit](https://github.com/NomicFoundation/edr/pull/498/commits/33d5df811d21864d06cb974daff96bc2d91c7e35).

crates/edr_eth/src/signature/fakeable.rs
Wodann: Addressed by https://github.com/NomicFoundation/edr/pull/498/commits/33d5df811d21864d06cb974daff96bc2d91c7e35

crates/edr_eth/src/signature/fakeable.rs
Wodann: Addressed by https://github.com/NomicFoundation/edr/pull/498/commits/33d5df811d21864d06cb974daff96bc2d91c7e35

crates/edr_eth/src/signature/fakeable.rs
Wodann: Addressed by https://github.com/NomicFoundation/edr/pull/498/commits/33d5df811d21864d06cb974daff96bc2d91c7e35

crates/edr_eth/src/signature.rs
```diff
@@ -123,345 +59,119 @@ pub struct Signature {
...
-    }
-}
+    /// Returns the signature's V-value.
+    fn v(&self) -> u64;
 
-impl From<Signature> for Vec<u8> {
-    fn from(src: Signature) -> Vec<u8> {
-        <[u8; 65]>::from(&src).to_vec()
-    }
+    /// Returns the signature's Y-parity value, if it exists.
```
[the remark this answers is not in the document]
Wodann: Addressed by https://github.com/NomicFoundation/edr/pull/498/commits/33d5df811d21864d06cb974daff96bc2d91c7e35

crates/edr_eth/src/signature.rs
```diff
@@ -123,345 +59,119 @@ pub struct Signature {
...
+    /// Contains the caller address.
+    ///
+    /// The only requirements on a fake signature are that when it is encoded as
+    /// part of a transaction, it produces the same hash for the same
+    /// transaction from a sender, and it produces different hashes for
+    /// different senders. We achieve this by setting the `r` and `s` values
+    /// to the sender's address. This is the simplest implementation and it
+    /// helps us recognize fake signatures in debug logs.
+    Fake {
+        // /// Whether the fake transaction uses Y-parity (0 or 1).
```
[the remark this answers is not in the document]
Wodann: Addressed by https://github.com/NomicFoundation/edr/pull/498/commits/33d5df811d21864d06cb974daff96bc2d91c7e35

crates/edr_eth/src/transaction/signed/eip155.rs
agostbiro: Right, thanks, I got confused by the serde implementation. Thought the internal type didn't have it.

agostbiro:
> As this distinction needs to happen at compile-time, I expressed that with two different types: `Ecdsa` and `EcdsaWithYParity`. Previously this was achieved by having the `r`, `s,`, and `odd_y_parity` fields listed separately in the `Eip2930` and later transactions, whereas the `Legacy` and `Eip155` transactions included the `Signature`.

Ok makes sense!
 
> I'm not trying to achieve a sum type, as the signature type that a transaction needs to receive is mutually exclusive (unless you consider polymorphism a form of sum type). Given that there are two compile-time types `Ecdsa` and `EcdsaWithYParity` (see above argumentation) that contain the `r`, `s`, and `v` or `odd_y_parity` values, to achieve both compile-time and runtime polymorphism, I introduced the trait that returns those values.

Ok so we want to abstract over different (data) types here. This is often done with polymorphism in object-oriented languages, but in languages that support sum types I prefer to do that as I find that it's easier to work with. In any case, I think it's ok to merge as is, so not a change request.

agostbiro (approved):
Thanks for the changes! I added some comments, the [comment](https://github.com/NomicFoundation/edr/pull/498#discussion_r1633321835) about signature type names is a nice to have, the rest is just info.

PR #499, opened 2024-06-07
build: Do not rely on RUSTFLAGS for Clippy lint configuration

Xanewok:
Since Rust 1.74 it's possible to configure lints directly in the Cargo.toml rather than relying on setting custom RUSTFLAGS: https://blog.rust-lang.org/2023/11/16/Rust-1.74.0.html#lint-configuration-through-cargo.

Setting `RUSTFLAGS` is fragile as there are some corner cases to look out for, like:
1. it [being exclusive with other sources of flags](https://doc.rust-lang.org/cargo/reference/config.html#buildrustflags) e.g. overwriting `build.rustflags` or `target.<cfg>.rustflags` or
2. forcing a separate rebuild whenever the flags change, preventing the reuse of build cache in common cases (e.g. for a regular build vs one with warnings denied).

See https://github.com/NomicFoundation/slang/pull/880 for what we did for Slang.

This also changes the pre-commit hook to not rely/set the flag and only runs `clippy`, rather than check + clippy, which seems wasteful in this case, as `cargo clippy` already runs the underlying rustc machinery for compilation checks.


Xanewok:
https://github.com/NomicFoundation/edr/actions/runs/9415777943/job/25937468081?pr=499 seems spurious, retrying:
```
 Alchemy Forked provider
         "before each" hook: Initialize provider for "Supports single and batched requests":
     Error: Timeout of 50000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/Users/runner/work/edr/edr/hardhat-tests/test/internal/hardhat-network/provider/modules/eth/websocket.ts)
```

Cargo.toml
```diff
@@ -12,3 +12,86 @@ rpath = true
 
 [profile.test]
 opt-level = 2
+
+[workspace.lints.rust]
+rust_2018_idioms = "warn"
+nonstandard_style = "warn"
+future_incompatible = "warn"
```
Wodann, line 19: Nice to have: alphabetic ordering

Wodann (approved):
One nice-to-have. Otherwise looks good to me.

For next time, please make sure to assign reviewers to your PR. That way people get a more obvious notification and will likely review quicker :)

PR #501, opened 2024-06-07
ci: Speed up the CI by building tests without opt

### LINKED ISSUE #439: Make CI run faster
The [EDR](https://github.com/NomicFoundation/edr/actions/workflows/edr-ci.yml) workflow takes more than 25 minutes to run. We should check if there's room for improvement here.

One easy gain is to remove the dependency between the tests and the "Check EDR" job. That would save some minutes, and the only downside is that we "waste" some computing time by running the tests when the build/lint fails.

We should also consider taking the slowest tests and only running them in main or during pre-release.

Xanewok:
Closes #439 

This implements the idea outlined in https://github.com/NomicFoundation/edr/issues/439#issuecomment-2100679905, which sets `opt-level` accordingly in the `test` Cargo profile.

I also had an idea to use the sparse crates.io checkout but didn't realize that in addition to it being stabilized, it was also immediately set as a default in Rust 1.68, so we should be using that already now.

To sum up some of the timings for building/running the tests (unit + integration + doctests):

| OS \ opt-level         | [2 (warm; baseline)](https://github.com/NomicFoundation/edr/actions/runs/9416944157/job/25941201261) | [1](https://github.com/NomicFoundation/edr/actions/runs/9417516559/attempts/1)       | [1 (second run)](https://github.com/NomicFoundation/edr/actions/runs/9417516559/attempts/2) | [0](https://github.com/NomicFoundation/edr/actions/runs/9418304660/attempts/1)      | [0 (second run)](https://github.com/NomicFoundation/edr/actions/runs/9418304660/attempts/2) |
|------------------------|--------------------|---------|------------|--------|------------|
| ubuntu-latest (build)  | 10m 8s             | 10m 50s | 10m 43s    | 4m 14s | 4m 8s  **(-59%)**     |
| ubuntu-latest (run)    | 47s                | 43s     | 44s        | 44s    | 43s        |
| macos-latest (build)   | 18m 32s            | 8m 36s  | 8m 38s     | 2m 47s | 2m 25s **(-87%)**    |
| macos-latest (run)     | 1m 56s             | 44s     | 44s        | 44s    | 36s        |
| windows-latest (build) | 21m 4s             | 19m 29s | 18m 40s    | 7m 53s | 6m 16s   **(-70%)**   |
| windows-latest (run)   | 1m 25s             | 1m 22s  | 1m 21s     | 49s    | 48s        |
| _total workflow time_   | _26m 44s_             | _25m 26s_  | _23m 39s_     | _13m 28s_    | _10m 56s_        |

1. What's surprising is the macos being overly slow at first on baseline and faster than ubuntu later - if I had to guess that may be some noise from other tenants?
2. What's also surprising is that time spent running also decreases when we decrease the opt-level. One guess would be that we keep using the default of `codegen-units = 256` but not run LTO and we end up bloating the binary, which is less cache friendly? That shouldn't have that much impact, though...

Despite no. 2 which should probably be investigated a little bit more at least locally, I think the build time savings are clear and much more outweigh the time spent running the tests themselves. Unless we're *very* off with test runtime, the rate of improvement should also translate to time savings when ran locally as well.

cc @Wodann since you originally proposed dropping the `opt-level` for `profile.test`


Xanewok:
@fvictorio  I've marked this as closing #439 since we're around the 10 min mark, which seems fine now. If we still want to pursue further improvements, feel free to edit the OP to drop the `Closes` remark.

Cargo.toml
```diff
@@ -10,5 +10,3 @@ rpath = true
 [profile.release]
 rpath = true
 
-[profile.test]
-opt-level = 2
```
Wodann, line 14: I'd propose a two-step process. We can create and use a separate profile for CI in which we can set the opt-level to 0.

Re point 2: I think we should validate locally whether the compilation + run time is also lower when using `opt-level = 0`

Moreover, we should validate how long the [`test-recent-mainnet-block`](https://github.com/NomicFoundation/edr/blob/main/.github/workflows/test-recent-mainnet-block.yml) workflow takes with opt-levels 0, 1, and 2 such that we ensure that that's running as fast as possible in CI too:

Wodann (changes requested):

Cargo.toml
Xanewok: Ah, good catch with `test-recent-mainnet-block`! I didn't realize we have tests where this might have an impact; will test this next. Are there any other tests or workflows that may be sensitive to that?

Also, do we expect to often locally run expensive tests in general? Is it useful? Sorry if that's a dumb question, I'm still trying to figure out what's what in EDR 😅

Wodann: The ones in the pre-commit hook are the tests that are part of the normal workflow. Whichever optimisation level takes least time there, would be preferable. That's the only thing you'd need to consider for that decision, I think :)

Xanewok: Posted local results in https://github.com/NomicFoundation/edr/pull/501#issuecomment-2158016805

Wodann (approved):
Thank you for the additional testing! Glad to see time decreases both locally and remotely! 😃

Wodann:
1. Yes, that's where the cacheable RPC requests are cached locally.
2. That sounds like a good idea. Feel free to combine all of the issue test files into one :)

Xanewok:
Combined some of the issues tests and got some good savings - there's still room for improvement but I'd consider this to be more or less done until we want to push it to extreme. For now, we got 9 minute run with Windows test taking the most (> 6 minutes), so the CI improvements seems sufficient for now.

Cargo.toml
fvictorio: FWIW I'm pretty sure that the `test-recent-mainnet-block` (execution) times are _completely_ dominated by networking (and by the very structure of the tests, these are never cahced), so I would be surprised if the optimization level makes a difference.

PR #505, opened 2024-06-11
refactor: remove ExecutableTransaction

Wodann:
This PR removes the `ExecutableTransaction`. Thanks to prior refactors, it's now possible to use the `signature::Fakeable` to retrieve the address of the caller/signer.

This will make it easier to have distinctions between signed types in L1 Ethereum VS Optimism, as we only need to support signed transactions and no longer `ExecutableTransaction` types too.


crates/edr_eth/src/signature.rs
```diff
@@ -52,14 +51,14 @@ pub enum SignatureError {
 /// A fakeable signature which can either be a fake signature or a real ECDSA
 /// signature.
 #[derive(Clone, Debug)]
-pub struct Fakeable<SignatureT: Recoverable + Signature> {
+pub struct Fakeable<SignatureT: Signature> {
```
Wodann, line 54: Previously I lazily recovered the address of a signature. Now I always ensure that the address is either provided or recovered. This ensures that we get the signature's recovery error upon construction, when it's clearest to the user  that they made a mistake.

The performance impact should be minimal, as we almost always know the `Address` anyway or were validating that the signature was valid already.

crates/edr_provider/src/requests/eth/transactions.rs
```diff
@@ -500,11 +496,6 @@ fn validate_send_raw_transaction_request<LoggerErrorT: Debug, TimerT: Clone + Ti
     data: &ProviderData<LoggerErrorT, TimerT>,
     transaction: &transaction::Signed,
 ) -> Result<(), ProviderError<LoggerErrorT>> {
-    // Validate signature
```
Wodann, line 503: This has been moved to the `alloy_rlp::Decodable::decode` function

agostbiro (approved):
Thanks LGTM with one question and there are some CI failures that need to be fixed before merging.

crates/edr_eth/src/signature/fakeable.rs
```diff
@@ -153,14 +132,7 @@ impl<SignatureT: Recoverable + Signature + PartialEq> PartialEq for Fakeable<Sig
...
                 },
-            ) => {
-                // SAFETY: The address is always initialized for fake signatures.
-                let address1 = unsafe { self.address.get_unchecked() };
-                // SAFETY: The address is always initialized for fake signatures.
-                let address2 = unsafe { other.address.get_unchecked() };
-
-                recovery_id1 == recovery_id2 && address1 == address2
-            }
+            ) => recovery_id1 == recovery_id2 && self.address == other.address,
```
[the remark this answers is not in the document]
Wodann: Addressed in https://github.com/NomicFoundation/edr/pull/505/commits/dae9dad64c1f67091c47410fcca73499c22da383

FYI, this is slightly less performant than it could be, as the recoverable signature should guarantee that the address is also equal, but I think that's fine.

PR #507, opened 2024-06-12
ci: use local EDR in benchmarks

agostbiro:
This PR adds a fix to use the local EDR in benchmarks. This got broken in https://github.com/NomicFoundation/edr/pull/416 by using a pre-release version of Hardhat for the benchmarks that didn't match the `hardhat@2>@nomicfoundation/edr`  specifier in the root `package.json`.


package.json
```diff
@@ -15,7 +16,7 @@
   },
   "pnpm": {
     "overrides": {
-      "hardhat@2>@nomicfoundation/edr": "workspace:*"
+      "hardhat>@nomicfoundation/edr": "workspace:*"
```
agostbiro, line 19: Remove `@2` to make sure it matches pre-release versions as well. I tested it locally that removing the `@2` matches pre-release versions as well.

crates/tools/js/benchmark/package.json
```diff
@@ -14,9 +14,9 @@
   "keywords": [],
   "author": "",
   "license": "ISC",
-  "dependencies": {
+  "devDependencies": {
     "argparse": "^2.0.1",
-    "hardhat": "^2.23.0-dev.0",
+    "hardhat": "*",
```
agostbiro, line 19: Specify Hardhat version at the workspace level

package.json
```diff
@@ -6,6 +6,7 @@
   "private": true,
   "devDependencies": {
     "@changesets/cli": "^2.27.1",
+    "hardhat": "^2.22.5",
```
agostbiro, line 9: Specify Hardhat version at the workspace level

Wodann (approved):

package.json
fvictorio: Ah cool, I don't know why I didn't do it this way in the first place.

crates/tools/js/benchmark/package.json
fvictorio: Are you sure that the semantic of `*` is "use the version at the worksapce level"? I always interpreted it as "use the latest dependency". Right now that matches the version in the root because that is the latest version, but I'm not sure that will always be the case.

If you want to guarantee that we are using the same version everywhere, we could adapt the [`check-dependencies`](https://github.com/NomicFoundation/hardhat/blob/main/scripts/check-dependencies.js) script from Hardhat and run it in the CI.

agostbiro: Uhh good catch this is what I get for asking ChatGPT how to use workspace level dependency definition 😬 

Fixed in: [Use specific version of Hardhat](https://github.com/NomicFoundation/edr/pull/507/commits/6f7a27d76a1070204c62c6394bf9da4c026f3f33)

fvictorio (approved):

PR #508, opened 2024-06-13
test: rename edr_evm issues tests

fvictorio:
- I added `hh` to the issue that comes from Hardhat. Right now it's obvious that a 4-digit issue is from Hardhat and a 3-digit one is from the EDR repo, but that can be confusing in the future.
- One of the Hardhat issues was moved to the EDR repo, so its number changed.
- I added descriptions to the names besides just having the issue numbers. We should probably do that in `edr_provider/tests/issues` too, but I didn't want to put time into that right now.


Wodann (approved):

agostbiro (approved):
Thanks LGTM

PR #513, opened 2024-06-14
refactor: make blockchain types generic

Wodann:
This PR adds a `trait ChainSpec` to `edr_evm` that allows specification of chain-specific internal/engine types. For now, we only deal with with L1 Ethereum.

Future work will implement a block builder and miner for Optimism and abstract that using a trait.

I used three traits `IntoRemoteBlock`, `EthRpcBlock`, and `EthRpcTransaction` to access the values that are required for state and blockchain implementations. I don't think this is particularly elegant, especially the `IntoRemoteBlock` but I wasn't sure what better way to abstract them. 

I'm open to suggestions for how I might be better able to abstract the information that needs to be retrieved from `ChainSpec::RpcBlock`. For now I stuck with this design, as it's hard to predict what this will look like with more chain types. As it stands, it is relatively easy to change the API in the future; e.g. if another chain type requires access to more properties of the RPC block or transaction.

### DIFF
--- a/crates/edr_evm/src/chain_spec.rs
+++ b/crates/edr_evm/src/chain_spec.rs
@@ -1,13 +1,24 @@
 use std::fmt::Debug;
 
 use alloy_rlp::RlpEncodable;
-use edr_eth::transaction::SignedTransaction;
+use edr_eth::{transaction::SignedTransaction, B256};
 use edr_rpc_eth::spec::{EthRpcSpec, RpcSpec};
 use revm::primitives::TxEnv;
 use serde::{de::DeserializeOwned, Serialize};
 
+use crate::{transaction::remote::EthRpcTransaction, EthRpcBlock, IntoRemoteBlock};
+
 /// A trait for defining a chain's associated types.
-pub trait ChainSpec: Debug + alloy_rlp::Encodable + RpcSpec {
+// Bug: https://github.com/rust-lang/rust-clippy/issues/12927
+#[allow(clippy::trait_duplication_in_bounds)]
+pub trait ChainSpec:
+    Debug
+    + alloy_rlp::Encodable
+    + RpcSpec<
+        RpcBlock<<Self as RpcSpec>::RpcTransaction>: EthRpcBlock + IntoRemoteBlock<Self>,
+        RpcTransaction: EthRpcTransaction,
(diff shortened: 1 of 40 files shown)


crates/edr_evm/src/chain_spec.rs
```diff
@@ -1,10 +1,24 @@
...
+pub trait ChainSpec:
+    Debug
+    + alloy_rlp::Encodable
+    + RpcSpec<
+        RpcBlock<<Self as RpcSpec>::RpcTransaction>: EthRpcBlock + IntoRemoteBlock<Self>,
+        RpcTransaction: EthRpcTransaction,
+    > + RpcSpec<RpcBlock<B256>: EthRpcBlock>
+{
     /// The type of signed transactions used by this chain.
     type SignedTransaction: alloy_rlp::Encodable
```
agostbiro, line 23: Just for my understanding: why were the bounds removed?

agostbiro (approved):
Thanks LGTM and I think the design is good for this stage. 

The only thing I got a little bit worried about looking at this PR is if we're going to corner ourselves somehow by having a mix of `Sized` and `?Sized` traits especially in combination with `'static` for futures. But this is out of scope for this PR, so I added a discussion topic to next week's technical sync.

crates/edr_evm/src/chain_spec.rs
Wodann: I don't understand which part of the code you're referring to. Could you please elaborate?

The section where this comment refers to only adds trait bounds

crates/edr_evm/src/chain_spec.rs
```diff
@@ -15,10 +29,23 @@ pub trait ChainSpec {
         + SignedTransaction;
```
agostbiro, line 29: > Just for my understanding: why were the bounds removed?

Meant the lines 11-14 that were removed in the original

crates/edr_evm/src/chain_spec.rs
agostbiro: Sorry, put the comment on the wrong line: https://github.com/NomicFoundation/edr/pull/513/files/a60e79318138bf2cfeeb84801c823c5c0a625f45#r1642537614

PR #514, opened 2024-06-14
improvement: add chain id to error message

fvictorio:
This error is a bit better if it includes the remote chain id, because then adding it to the default hardfork histories (as done in https://github.com/NomicFoundation/hardhat/pull/5394) will be easier.


agostbiro (approved):
Thanks, LGTM. There are some Hardhat integration tests that check the error message -- those might need to be udpated. And a changeset is needed.

fvictorio:
Based on both grepping and the fact that they pass, I don't think we need to update any tests.

PR #515, opened 2024-06-14
chore: use pnpm v9

fvictorio:
Moving to pnpm 9 to align with Hardhat and Ignition.

If you use [volta](https://volta.sh/), then doing `volta install pnpm@9` should be enough to migrate. I'm not sure about other installation methods.


Wodann (approved):

agostbiro (approved):
Thanks, LGTM

PR #516, opened 2024-06-14
fix: accept missing mix hashes in remote blocks

fvictorio:
See https://github.com/NomicFoundation/hardhat/discussions/5266#discussioncomment-9776377

I didn't add a test for this because neither Infura nor Alchemy support that creditcoin chain, and I don't think we should use public RPC URLs in our tests, since they don't tend to be very robust.

### DIFF
--- a/crates/edr_evm/src/block/remote.rs
+++ b/crates/edr_evm/src/block/remote.rs
@@ -79,6 +79,8 @@ impl RemoteBlock {
             gas_used: block.gas_used,
             timestamp: block.timestamp,
             extra_data: block.extra_data,
+            // TODO don't accept remote blocks with missing mix hash,
+            // see https://github.com/NomicFoundation/edr/issues/518
             mix_hash: block.mix_hash.unwrap_or_default(),
             nonce: block.nonce.ok_or(CreationError::MissingNonce)?,
             base_fee_per_gas: block.base_fee_per_gas,
--- a/crates/edr_evm/src/block/remote.rs
+++ b/crates/edr_evm/src/block/remote.rs
@@ -79,7 +79,7 @@ impl RemoteBlock {
             gas_used: block.gas_used,
             timestamp: block.timestamp,
             extra_data: block.extra_data,
-            mix_hash: block.mix_hash.ok_or(CreationError::MissingMixHash)?,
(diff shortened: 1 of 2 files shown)


crates/edr_evm/src/block/remote.rs
```diff
@@ -79,7 +79,7 @@ impl RemoteBlock {
             gas_used: block.gas_used,
             timestamp: block.timestamp,
             extra_data: block.extra_data,
-            mix_hash: block.mix_hash.ok_or(CreationError::MissingMixHash)?,
+            mix_hash: block.mix_hash.unwrap_or_default(),
```
Wodann, line 82: I think we should add a GitHub issue and `TODO` here to indicate that this should be reverted once we have multi-chain support.

Then we (or someone else) can add an implementation for this particular chai.

Wodann (changes requested):

Wodann (approved):
Otherwise LGTM

agostbiro (approved):
Thanks LGTM

jeremyjams:
Thank you for the fix! Do you have any idea when you plan to release it?
When EDR is released, would [going through this "process"](https://github.com/NomicFoundation/hardhat/pull/5087#issuecomment-2042188903) be enough to benefit from this fix from a fresh hardhat dependency?

crates/edr_evm/src/block/remote.rs
fvictorio: Done, thanks.

Can you re-review this? I had to solve a merge conflict and it was pretty straightforward, but it wouldn't hurt to double-check.

fvictorio:
@jeremyjams I guess we'll merge this today, and might release it later this week.

> When EDR is released, would https://github.com/NomicFoundation/hardhat/pull/5087#issuecomment-2042188903 be enough to benefit from this fix from a fresh hardhat dependency?

Correct.

Wodann (approved):

jeremyjams:
Hi, do you have any update on that?

jeremyjams:
Hello @fvictorio no problem, thank you for this release and your previous message :pray: (I'm sadly now facing another issue https://github.com/NomicFoundation/edr/issues/536)

Issue #503, opened 2024-06-11
Calling `hardhat_setStorageAt` after `eth_getStorageAt` doesn't work when forking


Rubilmax:
# Minimal reproducible example

Using hardhat 2.22.5 ([EDR ^0.4.0](https://github.com/NomicFoundation/hardhat/blob/bdd04d37f2ee41f68943f60bf0fd3ac5546a05ad/packages/hardhat-core/package.json#L105)), the following code rejects with "Contract with code hash `0xaee43571f51ad4922c41ab126fb98a76934d579c76f54b9e393ade09aad21646` does not exist.":

```typescript
    await getStorageAt(
      "0xbe9895146f7af43049ca1c1ae358b0541ea49704",
      "0x4f039c94bc7b6c8e7867b9fbd2890a637837fea1c829f434a649c572b15b2969",
    );

    await setStorageAt(
      "0xbe9895146f7af43049ca1c1ae358b0541ea49704",
      "0x4f039c94bc7b6c8e7867b9fbd2890a637837fea1c829f434a649c572b15b2969",
      "0x0000000000000000000000000000000000000000000000000000000000000001",
    );
```

While the following one does not:

```typescript
    await setStorageAt(
      "0xbe9895146f7af43049ca1c1ae358b0541ea49704",
      "0x4f039c94bc7b6c8e7867b9fbd2890a637837fea1c829f434a649c572b15b2969",
      "0x0000000000000000000000000000000000000000000000000000000000000001",
    );

    await getStorageAt(
      "0xbe9895146f7af43049ca1c1ae358b0541ea49704",
      "0x4f039c94bc7b6c8e7867b9fbd2890a637837fea1c829f434a649c572b15b2969",
    );
```

In other words, inverting the order of requests `eth_getStorageAt` and `hardhat_setStorageAt` changes the behavior of the EDR, while it should not (both should not reject)

fvictorio:
I can't reproduce this. @Rubilmax could you please create and share a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example)?

Rubilmax:
Sorry, update the description because I mixed up addresses.

Here is a reproducible example: https://github.com/Rubilmax/edr-setStorageAt-issue

Configure a rpc URL in the environment variable and run `yarn test`

Rubilmax:
[...]

PR #517, opened 2024-06-14
fix: split code from remote account in

Wodann:
Fixes https://github.com/NomicFoundation/edr/issues/503

### DIFF
--- a/crates/edr_evm/src/state/remote/cached.rs
+++ b/crates/edr_evm/src/state/remote/cached.rs
@@ -41,22 +41,9 @@ impl State for CachedRemoteState {
             return Ok(Some(account.info.clone()));
         }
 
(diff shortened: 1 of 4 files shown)


crates/edr_evm/src/state/remote/cached.rs
```diff
@@ -118,6 +104,30 @@ impl State for CachedRemoteState {
     }
 }
 
+/// Fetches an account from the remote state. If it exists, code is split off
+/// and stored separately in the provided cache.
+fn fetch_remote_account(
```
fvictorio, line 109: Out of curiosity, why make this a separate function instead of making it a method of `CachedRemoteState`?

fvictorio (approved):

agostbiro (approved):
Thanks, LGTM

Issue #421, opened 2024-04-29
Make release workflow more robust


fvictorio:
When publishing 0.3.6, we had an issue where one of the builds wasn't published. The underlying reason was that the meaning of `macos-latest` as an OS value changed since the last release. This was fixed in https://github.com/NomicFoundation/edr/pull/415, but we should make some deeper changes to prevent this from happening again.

At the very least, we should add a step that checks that all the `crates/edr_napi/npm` directories have a `.node` file, or fail otherwise. But there might be other things we can do instead or in addition to this.

Xanewok:
While we some some involved infrastructure in Slang, we also happen to do the check for that (and instrument the napi build a bit more ourselves) in https://github.com/NomicFoundation/slang/blob/3845dc78da4a29eccaa3193417ffd9506fab697d/crates/infra/cli/src/toolchains/napi/cli.rs#L88-L92.

Not saying we should do the same but maybe there is some helpful logic that we could carry over here.

PR #519, opened 2024-06-17
ci: make release workflow more robust

fvictorio:
The main change here is a couple of new steps, checking that the number of build jobs and the number of unique generated artifacts are the ones we expect.

There surely are better ways to do this, but this seems like a simple way to get it done. These are the relevant changes:
- Added a workflow-scoped `NUMBER_OF_TARGETS` variable. Its current value is 7.
- Added a check that the number of build jobs matches that number.
- Added a check that the number of _unique_ generated artifacts is the expected one.

The last item is the one that catches the problem we had before. I double checked this by temporarily setting the OSes to `macos-latest` again. This generated 6 unique artifacts, and was caught by that step.

I included two unrelated changes here, fixing some things I noticed while working on this:
- I added an entry to the matrix in the `test-macOS-windows-binding` job, which tests the `aarch64-apple-darwin` build in macos 14. We didn't have this before because there were no Apple ARM runners up until recently (in fact, the introduction of these runners is what caused the issue that motivates this PR).
- I noticed that the `test-linux-aarch64-musl-binding` job was testing the bindings in `node@lts`, while the rest of the tests where using node 18 and 20. I'm not sure if there was a reason for this, but I modified it to be consistent with the other tests.

---

Closes https://github.com/NomicFoundation/edr/issues/421

### DIFF
--- a/.github/workflows/edr-npm-release.yml
+++ b/.github/workflows/edr-npm-release.yml
@@ -87,13 +87,13 @@ jobs:
       run:
         working-directory: ./crates/edr_napi
     steps:
+      - uses: actions/checkout@v4
       - name: Check number of targets
         shell: bash
         run: |
           echo "Number of build jobs: ${{ strategy.job-total }}"
           echo "Expected number of build jobs: $NUMBER_OF_TARGETS"
           test ${{ strategy.job-total }} -eq "$NUMBER_OF_TARGETS"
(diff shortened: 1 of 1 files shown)


agostbiro (approved):
Thanks, LGTM

Wodann (approved):

PR #525, opened 2024-06-19
Stack traces test runner improvements

fvictorio:
I was re-visiting the stack traces test runner and noticed some minor things that could be improved. Nothing here is necessary, but it doesn't hurt.

I recommend reviewing this commit by commit.

### DIFF
--- a/hardhat-tests/test/internal/hardhat-network/stack-traces/test.ts
+++ b/hardhat-tests/test/internal/hardhat-network/stack-traces/test.ts
@@ -671,7 +671,11 @@ async function runDeploymentTransactionTest(
     gas: tx.gas !== undefined ? BigInt(tx.gas) : undefined,
   });
 
-  return trace as CreateMessageTrace;
+  if (!("deployedContract" in trace)) {
+    assert.fail("Expected trace to be a deployment trace");
+  }
+
+  return trace;
 }
 
 async function runCallTransactionTest(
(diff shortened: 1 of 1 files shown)


hardhat-tests/test/internal/hardhat-network/stack-traces/test.ts
```diff
@@ -191,22 +191,6 @@ function defineDirTests(dirPath: string, compilerOptions: SolidityCompiler) {
       describe(description, function () {
         defineTest(dirPath, testDefinition, sources, compilerOptions);
       });
-
-      if (process.env.HARDHAT_NETWORK_TESTS_WITH_OPTIMIZATIONS !== undefined) {
```
Xanewok, line 197: Is there still value in testing with different optimizer `runs` config? If there is, we can setup our workflow to test that separately or simply always define these tests with different numbers.

Xanewok (approved):

hardhat-tests/test/internal/hardhat-network/stack-traces/test.ts
fvictorio: I'm not sure how much value there is, but we weren't doing it so I guess it's not _that_ important.

If needed, we can add entries to `compilers-list.ts` with different runs. I think that should work with the current setup.

PR #527, opened 2024-06-19
ci: deploy book

fvictorio:


agostbiro (approved):
Thanks LGTM. One thought if we want deployments for branches as preview? I'm ok without it, just wanted to bring it up.

fvictorio:
Ah, deploy previews would be nice, but I guess we don't update the book so often for that to be that necessary right now. It would be great to add that in the future though (I don't know how to do that with GitHub pages).

Wodann:
Would this be the URL: https://nomicfoundation.github.io/edr/ ?

fvictorio:
https://nomicfoundation.github.io/edr/book/

PR #529, opened 2024-06-21
feat: adopt multi-chain revm

Wodann:


Wodann:
Some error messages changed, causing Hardhat tests to fail. I'll fix those.

crates/edr_eth/src/state.rs
```diff
@@ -1,3 +1,5 @@
+pub use revm_primitives::state::*;
```
agostbiro, line 1: Is this intentional? I'd prefer to use explicit imports if possible

crates/edr_evm/src/lib.rs
```diff
@@ -65,3 +60,8 @@ pub mod interpreter {
 pub mod precompile {
     pub use revm::precompile::{u64_to_address, PrecompileSpecId, Precompiles};
 }
+
+// /// Primitive types of Ethereum
+// pub mod primitives {
+//     pub use revm::primitives::*;
+// }
```
agostbiro, line 67: Is this intentional?

crates/edr_evm/src/runtime.rs
```diff
@@ -134,52 +140,57 @@ where
...
 }
 
-fn validate_configuration<BlockchainErrorT, StateErrorT>(
-    cfg: &CfgEnvWithHandlerCfg,
-    block: &BlockEnv,
-    transaction: &TxEnv,
-) -> Result<(), TransactionError<BlockchainErrorT, StateErrorT>> {
-    if cfg.handler_cfg.spec_id > SpecId::MERGE && block.prevrandao.is_none() {
-        return Err(TransactionError::MissingPrevrandao);
-    }
```
agostbiro, line 180: Just for my understanding: why don't we need to validate the randao here?

agostbiro (approved):
Thanks LGTM with some minor comments. Feel free to address them as you see fit before merging.

crates/edr_eth/src/state.rs
Wodann: I'm trying to find a good way to re-export the `revm_primitives` types to avoid other dependent crates or users from also having to import the exact same version. The simplest way I found is to re-export the public modules, but this public modules happens to have the same names as one of our own modules, `state`. Hence, why I re-exported all of its content in that particular sub-module.

Do you have any alternative proposals?

crates/edr_evm/src/lib.rs
Wodann: No it was not; thanks!

crates/edr_eth/src/state.rs
agostbiro: Ok thanks it makes sense to have a wildcard import for this use case.

crates/edr_evm/src/lib.rs
Wodann: Addressed in https://github.com/NomicFoundation/edr/pull/529/commits/255137ab45eafcb0023ba8fc16515cef14edfdbb

Issue #487, opened 2024-06-03
Decouple Foundry build and project format from JS Solidity test runner


agostbiro:
**Problem**

The JS Solidity test runner executes bytecode compiled with Hardhat, but it still depends on the Foundry project format (file layout, config) and build system to provide the following functionality (possibly incomplete):

- [In-line test configuration](https://book.getfoundry.sh/reference/config/inline-test-config?highlight=inline%20test%20configs#in-line-test-configuration)
- Regulate [file system access](https://book.getfoundry.sh/cheatcodes/fs?highlight=file%20system#file-cheat-codes)
- [Managing dependencies](https://book.getfoundry.sh/projects/dependencies?highlight=depe#dependencies)
- [Remapping dependencies](https://book.getfoundry.sh/projects/dependencies?highlight=remapping#remapping-dependencies)
- Link contracts 
- [vm. getDeployedCode](https://book.getfoundry.sh/cheatcodes/get-deployed-code)

**Definition of Done**

The Foundry build and project format is decoupled from JS Solidity test runner. TBD what functionality to preserve

PR #530, opened 2024-06-24
refactor: remove `foundry_config::Config` dependency from `MultiContractRunner`

agostbiro:
The goal of this PR is to remove the `foundry_config::Config` type from `MultiContractRunner`. It's a first step towards https://github.com/NomicFoundation/edr/issues/487

### DIFF
--- a/crates/edr_napi/src/solidity_tests/config.rs
+++ b/crates/edr_napi/src/solidity_tests/config.rs
@@ -1,0 +1,99 @@
+use std::{collections::HashMap, path::PathBuf};
+
+use alloy_primitives::{address, Address};
+use edr_eth::U256;
+use forge::{
+    inspectors::cheatcodes::CheatsConfigOptions,
+    opts::{Env as EvmEnv, EvmOpts},
+};
+use foundry_compilers::ProjectPathsConfig;
+use foundry_config::{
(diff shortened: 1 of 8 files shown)


crates/edr_napi/src/solidity_tests/config.rs
```diff
@@ -0,0 +1,99 @@
...
+        // Matches Foundry config defaults
+        let gas_limit: GasLimit = i64::MAX.into();
+        let evm_opts = EvmOpts {
+            env: EvmEnv {
+                gas_limit: gas_limit.into(),
+                chain_id: None,
+                tx_origin: DEFAULT_SENDER,
+                block_number: 1,
+                block_timestamp: 1,
+                ..Default::default()
```
Wodann, line 47: Are lints enabled? IIRC, a lint is enabled that warms when not doing this:
```suggestion
                ..EvmEnv::default()
```

Wodann (commented):
LGTM. Just one question/suggestion

Wodann (approved):

crates/edr_napi/src/solidity_tests/config.rs
agostbiro: Yeah I double checked. This case isn't covered by the [`default_trait_access`](https://rust-lang.github.io/rust-clippy/v0.0.212/index.html#default_trait_access) lint, because the type is clear from the struct literal.

Issue #246, opened 2023-12-13
Port runtime observation and solidity-data decoding features


alcuadrado:
This is a tracking issue created as a result of #242. We analyzed this scope with @fvictorio, split it into smaller submodule, and uncover their dependencies and came up with an execution plan.

## Features withing scope

* Generating solidity stack traces
* Inferring common error cases based on heuristics
* Decoding the `returndata` of failed evm calls when possible

## Submodules

The submodules are these:
![image](https://github.com/NomicFoundation/edr/assets/176499/4d24dc43-6dfe-4cb2-9d39-716779a15e6e)

- Codebase model:An internal model of the codebase and compilation output that can be used to back all the runtime analysis that we need to do. This is found [here](https://github.com/NomicFoundation/hardhat/blob/main/packages/hardhat-core/src/internal/hardhat-network/stack-traces/model.ts) in Hardhat. A 1:1 port would be Hard to express in Rust, and at the same time, there are multiple optimization opportunities. 
- build-info to codebase model: A module that given a build-info (i.e. compilation input & output pair), constructs the model described above.
- Contract identifier: A module that tries to identify which contract is being run in a certain evm call. Due to linked libraries, immutable variables inlining, and constructor parameters being appended to the bytecode, this is not a straightforward search. The most recent TS implementation of this is [here](https://github.com/NomicFoundation/hardhat/commit/618f44f4028f6f373c3944d951bf24b1302bbba4)
- EVM Trace Generation: A module that creates a static low-level trace of the execution of an EVM call. This has no knowledge of Solidity. It has already been implemented by @Wodann 
- Raw stack traces generation: A module that naively applies source maps to the EVM trace to generate a Solidity stack trace.
- Stack traces improver: A module that takes the naive solidity stack trace and improves them to be either more correct or meaningful. This is necessary due to the source maps generated by solidity being unreliable.
- Stack trace format: A format (i.e. data structure) that can be used to represent raw and improved stack traces. We can consider having two of them though. This can be ported 1:1 from Hardhat Network, as the format we used there has a mixture of responsibilities that we agreed was a mistake. It includes information about the `returndata` decoding result, and about any inferred error.
- Error inferred: A module that tries to detect common error cases in failed evm calls.
- Return data decoding: A module that tries to decode the `returndata` of a failed evm call. It should support revert reasons, custom errros and panics.
[...]

PR #545, opened 2024-07-05
feat: Port existing stack trace refinement 1:1 from Hardhat 

Xanewok:
Based on https://github.com/NomicFoundation/edr/pull/544.

This feature branch will track porting the existing stack trace logic 1:1 from Hardhat. The main goal is to gradually port it while retaining the existing logic. While porting, we aim to retain the JS interface ideally as-is; once the port is complete, we can then attempt to simplify the internal code (as it won't necessarily have to be exported/accessed via JS as much) or doing the decoupling work outlined in #246.

To continuously test the changes, we will use the vendored `hardhat-tests` suite. However, to not have to vendor the entire Hardhat repository, we will collect patches (created via `pnpm patch`) to facilitate the review and to allow better grouping of the related changes. Each commit with a patch should ideally point to a commit in the NomicFoundation/hardhat repository.

I tried to use the patch mechanism from the github repository directly via package.json but I don't think it's possible to do given the Hardhat repository setup (monorepo, the `root` package workaround) as I encountered errors. I'd be happy to switch to that, instead, if we found a way to make it work.

The initial work included here tries to tackle #247. We consume the compiler input/output JSON from the JS and create the relevant model in the Rust side, while exposing the same class instances and relationship to JS via `napi-rs`.

To test this locally, you can also:
1. clone the two repositories and in the Hardhat repository:
2. use the companion branch https://github.com/NomicFoundation/hardhat/tree/refactor/port-solidity-stack-traces
3. Link to the EDR (`edr_napi` crate) locally via `pnpm link`
4. run the test suite there for the `hardhat-core` package (and to optionally speed things up, only run the relevant suite via `describe.only("Stack traces", ...)`.


Xanewok:
I'm applying the "no changeset needed" for now not to hide the overall CI status at a glance while we're making the changes. I'll drop the label once we're ready to merge this to main.

Xanewok:
@NomicFoundation/edr Could I get an initial review of this? It'd be great to land changes incrementally and this already covers quite a bit of logic as it recreates the existing codebase model construction logic from Hardhat.

package.json
```diff
@@ -16,6 +16,9 @@
   "pnpm": {
     "overrides": {
       "hardhat>@nomicfoundation/edr": "workspace:*"
+    },
+    "patchedDependencies": {
+      "hardhat@2.22.6": "patches/hardhat@2.22.6.patch"
```
fvictorio, line 21: Ah, I hadn't understood that these patches are applied automatically on installation, cool!

fvictorio (commented):
LGTM. I only skimmed the Rust code, but I guess in general is as much a 1-1 translation as possible, right?

Something I wondered while looking at this is why some logic lives completely on the `edr_napi` crate (e.g., source maps), instead of living in `edr_solidity` and `edr_napi` just importing and exposing them.

Xanewok:
> why some logic lives completely on the edr_napi crate (e.g., source maps) ...

The main difficulty here is to retain the exact same object/class relationship, so I tried to be as faithful as possible. For instance, while some of the source map logic is abstract, ultimately it has to create the JS-facing class instances using facilities from `edr_napi`, so for the time being I'm bundling everything that I need together and once the need to publicly expose that disappears, I will then untangle the code to decrease the N-API exposed surface.

Xanewok:
Copied benchmark summaries from both of the latest runs:

## [Benchmark](https://github.com/NomicFoundation/edr/actions/runs/10559832240/attempts/1#summary-29252065569) - sanity check, just adding new logic (vs latest `main`)



| Benchmark suite | Current: 272b4c74b205bb3c708c4571e4066ba37da1eb27 | Previous: fdf1583763e93b944e5ebe2df2800ce54073e001 | Ratio |
|-|-|-|-|
| `All Scenarios` | `342935.803493` ms | `342710.45338099997` ms | `1.00` |
| `neptune-mutual-blue-protocol_8db6480` | `33781.23079` ms | `33761.015113` ms | `1.00` |
| `openzeppelin-contracts_0a5fba7a` | `21761.936611` ms | `20441.195663` ms | `1.06` |
| `rocketpool_6a9dbfd8` | `21576.112301` ms | `22240.754139999997` ms | `0.97` |
| `safe-contracts_914d0f8` | `1684.9832250000002` ms | `1703.387087` ms | `0.99` |
| `seaport_4f4e7c20` | `7113.415929999999` ms | `6752.069426999999` ms | `1.05` |
| `synthetix_9a3a109f` | `250565.653007` ms | `251436.68594` ms | `1.00` |
| `uniswap-v3-core_d8b1c63` | `6452.471629` ms | `6375.346011` ms | `1.01` |





## [Benchmark](https://github.com/NomicFoundation/edr/actions/runs/10560832959/attempts/1#summary-29255181307) - patching Hardhat to use the new, ported logic (vs latest `main`)



| Benchmark suite | Current: ac2e0125a4644296e27cff3b733183c553cdd95f | Previous: fdf1583763e93b944e5ebe2df2800ce54073e001 | Ratio |
|-|-|-|-|
| `All Scenarios` | `305142.086981` ms | `342710.45338099997` ms | `0.89` |
| `neptune-mutual-blue-protocol_8db6480` | `29736.097993` ms | `33761.015113` ms | `0.88` |
| `openzeppelin-contracts_0a5fba7a` | `17890.910111` ms | `20441.195663` ms | `0.88` |
| `rocketpool_6a9dbfd8` | `19060.544268` ms | `22240.754139999997` ms | `0.86` |
| `safe-contracts_914d0f8` | `1316.443957` ms | `1703.387087` ms | `0.77` |
| `seaport_4f4e7c20` | `5582.380408` ms | `6752.069426999999` ms | `0.83` |
| `synthetix_9a3a109f` | `225767.973043` ms | `251436.68594` ms | `0.90` |
| `uniswap-v3-core_d8b1c63` | `5787.737201` ms | `6375.346011` ms | `0.91` |


I'll remove the local patch for now as discussed with @fvictorio and this should be good to review, I hope? :crossed_fingers: :tada:

fvictorio (approved):

agostbiro (approved):
Thanks LGTM overall and congrats @Xanewok, this is a big one!

I have one doubt, but not a merge blocker: there are some parts where info is printed using `println!`. I'm assuming this is because the corresponding JS code would `console.log` there. There were some discussions about how we want to handle logging from EDR and this won't compose well with our planned usage of a log facade.  I added a discussion point on this to today's technical sync.

Wodann (approved):
Awesome work on this, @Xanewok! LGTM 🚢

Xanewok:
Thanks everyone for the reviews! I'm happy that this can finally land now.

> there are some parts where info is printed using println!.

we discussed this and the code is only being used in the test suite. I can refactor it in the follow-up to use a Debug trait and add some comments that the logging should only for debugging/in the test suite or let it go through the `log` facade if needed.

PR #552, opened 2024-07-11
feat: add support for RIP-7212

### LINKED ISSUE #523: Add support for RIP-7212
### Describe the feature

RIP-7212 provides a precompile, at address(0x100) for secp256r1 verification. This is already live on multiple networks, including Matic (polygon).

It is currently not possible to test code that relies on this precompile using hardhat.

Some pointers:
- [REVM has this precompile implemented, but only loads it for FJORD](https://github.com/bluealloy/revm/blob/main/crates/revm/src/optimism/handler_register.rs#L148-L153), this would require some change so that it can be activated manually through the config
- This option will then have to be exposed through the hardhat config.

### Search terms

secp265r1, p256, precompile, RIP-7212, EIP-7212

Wodann:
Resolves #523 

A quick (and dirty) implementation to add support for RIP-7212.

I discussed with @fvictorio whether we expect future requests for optional precompiles, but it's unsure what the requirements would be for that at this time. For that reason, I chose a simple design that only assumes a single boolean for configuring a precompile.

As this adds a new field to the `ProviderConfig`, it's a breaking change and requires a minor version bump.

**To Do**
This PR depends on changes in https://github.com/NomicFoundation/hardhat/pull/5512 but that in turns depends on this PR. How should situations like this be resolved?


agostbiro:
> As this adds a new field to the ProviderConfig, it's a breaking change and requires a minor version bump.

We could potentially default to false to avoid the breaking change.

crates/edr_napi/src/result.rs
```diff
@@ -113,7 +113,7 @@ impl From<edr_evm::HaltReason> for ExceptionalHalt {
         match halt {
             edr_evm::HaltReason::OutOfGas(..) => ExceptionalHalt::OutOfGas,
             edr_evm::HaltReason::OpcodeNotFound => ExceptionalHalt::OpcodeNotFound,
-            edr_evm::HaltReason::InvalidEFOpcode => ExceptionalHalt::InvalidEFOpcode,
+            edr_evm::HaltReason::InvalidEFOpcode => ExceptionalHalt::InvalidFEOpcode,
```
agostbiro, line 116: Should `edr_evm::HaltReason::InvalidEFOpcode` be renamed as well?

agostbiro (approved):
Hey, thanks for implementing this! I don't think this use case warrants a const generic bool, since the branch is not in a performance sensitive place, but it's not a merge blocker either, so LGTM overall.

fvictorio:
> We could potentially default to false to avoid the breaking change

I thought the same thing, but we also have some (very minor) changes related to the REVM upgrade which are also technically breaking changes, so I guess it makes sense to err on the side of bumping the minor. I should've done that in the REVM upgrade PR, but I didn't notice it because of the whole `index.d.ts` thing.

fvictorio:
Hmm, but thinking in terms of API, maybe it makes sense to make it optional anyway? We already have some optional configs (the fork config, `initialBaseFeePerGas`, etc.), and this seems to be that kind of optional configuration too.

crates/edr_provider/tests/rip7212.rs
```diff
@@ -0,0 +1,49 @@
...
+use edr_provider::{
+    test_utils::create_test_config, time::CurrentTime, MethodInvocation, NoopLogger, Provider,
+    ProviderRequest,
+};
+use edr_rpc_eth::CallRequest;
+use revm_precompile::secp256r1;
+use tokio::runtime;
+
+#[tokio::test(flavor = "multi_thread")]
+async fn rip_7212() -> anyhow::Result<()> {
```
fvictorio, line 14: Can we add another test that checks that the response is `0x` for the same calldata when the flag is false?

crates/edr_napi/src/result.rs
Wodann: The `edr_evm::HaltReason::InvalidEFOpcode` is a re-export from `revm`. They renamed it, released, and then [reverted it](https://github.com/bluealloy/revm/commit/81818a9fd088de7a3df65b9f9a70a180d5a0fe79) based on feedback from Franco.

As they haven't done a new release, that change will have to wait until `revm` releases. I just wanted to affecting the `index.d.ts` bindings unnecessarily.

Wodann:
> I don't think this use case warrants a const generic bool

I should have mentioned this in the original post. I attempted to implement this using a runtime argument alone, but it failed due to lifetime errors when calling `append_handler_register`. It was impossible to resolve those, so I had to revert to compile time arguments to be able to pass them along in the `DebugContext` - which expects a function pointer.

I checked the API again and noticed a `fn append_handler_register_box`. Potentially that circumvents the lifetime issue. I can give that a try still.

Wodann:
> We already have some optional configs (the fork config, initialBaseFeePerGas, etc.), and this seems to be that kind of optional configuration too.

@fvictorio the optional parameters are settings that are actually optional:
- `forkConfig` signals either local or forked mode
- `initialBaseFeePerGas` is only present post-Merge
- same for the others.

agostbiro:
> I should have mentioned this in the original post. I attempted to implement this using a runtime argument alone, but it failed due to lifetime errors when calling `append_handler_register`. It was impossible to resolve those, so I had to revert to compile time arguments to be able to pass them along in the `DebugContext` - which expects a function pointer.
> 
> I checked the API again and noticed a `fn append_handler_register_box`. Potentially that circumvents the lifetime issue. I can give that a try still.

Ah ok thanks for explaining, that makes sense. If it's not too much trouble, it'd be nice to see if using `fn append_handler_register_box` allows for circumventing the issue.

crates/edr_provider/tests/rip7212.rs
Wodann: Addressed in https://github.com/NomicFoundation/edr/pull/552/commits/25120887c4dfd21d494fce6b121d8c14310c8ef7

Wodann:
> Ah ok thanks for explaining, that makes sense. If it's not too much trouble, it'd be nice to see if using `fn append_handler_register_box` allows for circumventing the issue.

Addressed in https://github.com/NomicFoundation/edr/pull/552/commits/c42d8600483b6ba9ccacde6de9e5811bbd4043f5

crates/edr_napi/src/result.rs
Wodann: As part of the refactor to remove "const generic bool" I had to upgrade to the latest revm, which also resolves this comment.

agostbiro (approved):
Thanks LGTM

PR #560, opened 2024-07-22
chore: move hardhat-solidity-tests to the EDR repo

fvictorio:


agostbiro (changes requested):
Thanks looks good overall, but could you also include the usage instructions (and update if necessary) from the [original repo?](https://github.com/NomicFoundation/hardhat-solidity-tests?tab=readme-ov-file#how-to-use)

fvictorio:
Good catch, thanks!

While writing the readme I also realized that it makes sense to have the example in this repo too, so no `pnpm link` is needed at all for running it.

hardhat-solidity-tests/README.md
```diff
@@ -0,0 +1,10 @@
+Temporary directory to experiment with a Hardhat plugin that uses EDR's solidity tests feature.
+
+This directory can be removed once the plugin is handed over to the Hardhat team.
+
+## How to use
+
+1. Clone this repo and [the example](https://github.com/NomicFoundation/hardhat-solidity-tests-example)
```
agostbiro, line 7:
```suggestion
1. Clone this repo and [the usage example](https://github.com/NomicFoundation/hardhat-solidity-tests-example) to `./example`
```

agostbiro: Note that the example repo is private, we should make that public probably

agostbiro (approved):
Thanks LGTM with one suggestion for improvement

hardhat-solidity-tests/README.md
fvictorio: No, that was my bad. There's no need to clone that repo anymore. I simplified the steps now.

agostbiro: Ah ok cheers! Good to merge from my side

PR #562, opened 2024-07-25
refactor: Port ContractsIdentifier to Rust

Xanewok:
Based on #531, should be merged after that.

This naively ports the `ContractsIdentifier` trie. While some of the logic already lives in `crates/edr_napi/src/trace/contracts_identifier.rs`, I plan to, like with the rest of the related PRs, first faithfully migrate what we have on the edr_napi side as until it does not use any JS-facing types and only then collapse and simplify the types once they can use pure Rust.


fvictorio:
@Xanewok I changed the base branch to `refactor/port-vm-tracer`, to make this easier to review. Once #531 is merged, GitHub should automatically change the base branch of this PR to `refactor/port-solidity-stack-traces` (I think! At least for the `main` branch, that works, but I'm pretty sure I've seen that working for feature branches.)

fvictorio:
Just FYI, this can be dangerous in some situations:

https://github.com/NomicFoundation/hardhat/commit/dd34257cec2ceddc4e7065aa9147378d8f75da98#diff-ff73b14c16c5b75145e083f5574c6a8e9d8cb91ee67359dea27d9de1c8b238d5L142-R125

Using `instanceof` instead of a structural check (like `isBytecodeTrie`) can sometimes be problematic, but IIRC it's only relevant if there is inheritance involved, which is not the case here. Just sharing in case this causes an issue for you down the road, although it shouldn't. But I understand that you needed to do this because `BytecodeTrie` was removed.

fvictorio (approved):
Thanks, LGTM

PR #568, opened 2024-07-26
refactor: Port the `VmTraceDecoder` from Hardhat

Xanewok:
This ports of `VmTraceDecoder` from Hardhat; what's left now is the error inferrer and finally solidity stack trace logic :partying_face: 

What's more interesting, however, is the fix for the `MessageTrace` and others:
1. There's a `napi-rs` bug that incorrectly validates and converts an object that has a required property but can be explicitly undefined, e.g. `deployedContract: Bytecode | undefined`. That's the case for `CreateMessageTrace` and so this is worked around using an optional property, however...
2. With this, the `CallMessageTrace` is a strict superset of `CreateMessageTrace`. We need to be careful with the type order when using the N-API type `Either` to always attempt converting and validating the most specific type first, otherwise we may incorrectly convert into a less specific type, skipping some additional properties it may have.

Link to the companion commit: https://github.com/nomicFoundation/hardhat/commit/137cfb33a8868329ae4d3d3903f4c7bf8838a50f

### DIFF
--- a/crates/edr_napi/src/trace/message_trace.rs
+++ b/crates/edr_napi/src/trace/message_trace.rs
@@ -27,6 +27,10 @@ pub struct PrecompileMessageTrace {
     pub calldata: Uint8Array,
 }
 
+// NOTE: Because of the hack below for `deployed_contract`, now the `CallMessageTrace`
+// is a strict superset of `CreateMessageTrace`, so we need to take care to keep
+// the order consistent from most-specific to least-specific in the `Either{3,4}` type
+// when converting to or from N-API.
 #[napi(object)]
 pub struct CreateMessageTrace {
     // `BaseMessageTrace`
@@ -37,16 +41,21 @@ pub struct CreateMessageTrace {
     pub depth: u32,
     // `BaseEvmMessageTrace`
     pub code: Uint8Array,
-    pub steps: Vec<Either4<EvmStep, PrecompileMessageTrace, CreateMessageTrace, CallMessageTrace>>,
+    pub steps: Vec<Either4<EvmStep, PrecompileMessageTrace, CallMessageTrace, CreateMessageTrace>>,
     // TODO: Will be later filled on the JS side but we should port to ContractsIdentifier in Rust
     // This is explicitly `any` on the JS side to side-step the type-checking until we port
     #[napi(ts_type = "any")]
     pub bytecode: Option<Object>,
     pub number_of_subtraces: u32,
     // `CreateMessageTrace`
-    // NOTE: we can't use Option<T> as this is converted to an optional property,
-    // which is not backwards compatible with the current JS interface
-    pub deployed_contract: Either<Uint8Array, Undefined>,
+    // HACK: It seems that `Either<Uint8Array, Undefined>` means exactly what we
+    // want (a required property but can be explicitly `undefined`) but internally
+    // the napi-rs treats an encountered `Undefined` like a missing property
+    // and it throws a validation error. While not 100% backwards compatible, we
+    // work around using an optional type.
+    // See https://github.com/napi-rs/napi-rs/issues/1986 for context on the PR
+    // that introduced this behavior.
+    pub deployed_contract: Option<Either<Uint8Array, Undefined>>,
 }
 
 #[napi(object)]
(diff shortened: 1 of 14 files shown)


crates/edr_napi/src/trace/message_trace.rs
```diff
@@ -37,15 +40,19 @@ pub struct CreateMessageTrace {
...
-    pub bytecode: Option<Value>,
+    pub steps: Vec<Either4<EvmStep, PrecompileMessageTrace, CallMessageTrace, CreateMessageTrace>>,
+    #[napi(ts_type = "Bytecode")]
+    pub bytecode: Option<Object>,
     pub number_of_subtraces: u32,
     // `CreateMessageTrace`
-    // NOTE: we can't use Option<T> as this is converted to an optional property,
-    // which is not backwards compatible with the current JS interface
-    pub deployed_contract: Either<Uint8Array, Undefined>,
+    // HACK: It seems that `Either<Uint8Array, Undefined>` means exactly what we
```
Wodann, line 48: FYI, backwards compatibility for Hardhat's internals is not a requirement, so if it's easier, you can just adopt `Option<Uint8Array>` and change the Hardhat code.

crates/edr_napi/src/trace/message_trace.rs
```diff
@@ -58,10 +65,9 @@ pub struct CallMessageTrace {
     pub depth: u32,
     // `BaseEvmMessageTrace`
     pub code: Uint8Array,
-    pub steps: Vec<Either4<EvmStep, PrecompileMessageTrace, CreateMessageTrace, CallMessageTrace>>,
-    // TODO: Will be later filled on the JS side but we should port to ContractsIdentifier in Rust
-    // This is explicitly `any` on the JS side to side-step the type-checking until we port
-    pub bytecode: Option<Value>,
+    pub steps: Vec<Either4<EvmStep, PrecompileMessageTrace, CallMessageTrace, CreateMessageTrace>>,
+    #[napi(ts_type = "Bytecode")]
```
Wodann, line 69: If this is now filled on the Rust side, is there a reason we can't define an N-API `Bytecode` type?

Wodann (commented):
Overall LGTM. Some questions and comments

crates/edr_napi/src/trace/message_trace.rs
Xanewok: We do define and use `Bytecode` elsewhere already but there's some subtle nuance here:

1. We can't use `ClassInstance` here because it internally holds a `&'static mut T` so that's UB because we probably will often have traces that point to the same bytecode
2. I introduced the `ClassInstanceRef` that guards the access with a `RefCell`, however it's not possible to directly use it here as it'd require `ToNapiValue` and `FromNapiValue` implementation...
3. and while it might be feasible to implement it using `ToNapiValue` (it can pass the underlying object), however...
4. implementing `FromNapiValue` would be very tricky:
  a. the instances are shared, so on the Rust side are wrapped in `Rc` and internally uses `napi::Ref` for napi-side ref-counting of the underlying object, so...
  b. converting at the FFI boundary without a side-table would have to create new Rc'd objects; ignoring the potential perf drawbacks, it would also be incorrect because it would lead to multiple `RefCells` guarding the same memory access and...
  c. introducing the side-table means introducing a static map per JS class on the Rust side, hooking into the napi lifecycle to manage the objects on the Rust side once they're freed on the JS side, modifying the `ClassInstanceRef` helper to do the same on the Rust side...
  d. ...at which point we'd be reimplementing quite a bit of higher-level convenience code that arguably should be on the napi-rs side itself; hopefully, we won't need as much JS glue anymore once the port is complete.
  
Let me know if the context makes sense; I'm more than happy if we simplified this somehow but I didn't manage to find a way to make it both safe and convenient :cry:

fvictorio (approved):
LGTM, but let's wait for @Wodann's approval

crates/edr_napi/src/trace/message_trace.rs
Wodann: It sounds very complicated. I'll trust your judgement!

Xanewok: Thank you! I'll try to embed a bit more context and/or more explain these things more concisely next time :pray:

PR #569, opened 2024-07-29
fix: prevent crash when returning large JSON responses

### LINKED ISSUE #543: "Error: Failed to convert rust `String` into napi `string`" on `debug_traceTransaction`
Hi guys. I'm getting this error when I'm trying to get traces for a big transaction.

### Steps to represent
Install the latest hardhat
```shell
npm i hardhat@2.22.6
```
Run hardhat node as a mainnet fork. Config:
```ts
const config: HardhatUserConfig = {
  solidity: "0.8.23",
  networks: {
    hardhat: {
      initialBaseFeePerGas: 0,
      accounts: {
        count: 10,
      },
      chainId: 1,
      forking: {
        url: https://mainnet.infura.io/v3/<INFURA_TOKEN>,
      },
    },
  },
};
```
Run:
```shell
npx hardhat node
```

Make request
```shell
curl --location 'localhost:8545' \
--header 'Content-Type: application/json' \
--data '{
  "jsonrpc": "2.0",
  "method": "debug_traceTransaction",
  "params": ["0x7e460f200343e5ab6653a8857cc5ef798e3f5bea6a517b156f90c77ef311a57c"],
  "id": 1
}'
```
Response:
```json
{
    "jsonrpc": "2.0",
    "id": 1,
    "error": {
        "code": -32603,
        "message": "Error: Failed to convert rust `String` into napi `string`",
        "data": {
            "message": "Error: Failed to convert rust `String` into napi `string`"
        }
    }
}
```
When I'm trying to use a smaller transaction, everything works fine.

Works with `0xce8f47c3c6e0dc401dbbd6ad640bd039362eded9d2a6767f03c7229118eb66bd`
Fails with `0x7e460f200343e5ab6653a8857cc5ef798e3f5bea6a517b156f90c77ef311a57c`

Wodann:
Resolves #543

### DIFF
--- a/crates/edr_napi/index.d.ts
+++ b/crates/edr_napi/index.d.ts
@@ -418,7 +418,8 @@ export class Provider {
   setVerboseTracing(verboseTracing: boolean): void
 }
 export class Response {
-  get json(): string
+  /** Returns the response data as a JSON string or a JSON object. */
+  get data(): string | any
   get solidityTrace(): RawTrace | null
   get traces(): Array<RawTrace>
 }
--- a/crates/edr_napi/test/issues.ts
+++ b/crates/edr_napi/test/issues.ts
@@ -1,0 +1,109 @@
+import chai, { assert } from "chai";
+import { JsonStreamStringify } from "json-stream-stringify";
+
+import {
+  ContractAndFunctionName,
+  EdrContext,
+  MineOrdering,
+  Provider,
+  SpecId,
+  SubscriptionEvent,
+} from "..";
+import { ALCHEMY_URL, isCI } from "./helpers";
+
+describe("Provider", () => {
+  const context = new EdrContext();
+  const providerConfig = {
+    allowBlocksWithSameTimestamp: false,
+    allowUnlimitedContractSize: true,
+    bailOnCallFailure: false,
+    bailOnTransactionFailure: false,
+    blockGasLimit: 300_000_000n,
+    chainId: 1n,
+    chains: [],
+    coinbase: Buffer.from("0000000000000000000000000000000000000000", "hex"),
+    enableRip7212: false,
+    genesisAccounts: [
+      {
+        secretKey:
+          "0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80",
+        balance: 1000n * 10n ** 18n,
+      },
+    ],
+    hardfork: SpecId.Latest,
+    initialBlobGas: {
+      gasUsed: 0n,
+      excessGas: 0n,
+    },
+    initialParentBeaconBlockRoot: Buffer.from(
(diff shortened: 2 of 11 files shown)


agostbiro:
Thanks for this fix, the code looks good, but unfortunately it looks like [there is](https://github.com/NomicFoundation/edr/actions/runs/10152890796/job/28075167130?pr=569) an 115% slowdown over all and 120% slowdown on the `synthetix` benchmark. I reran the benchmarks just in case, but we probably have to come up with a different approach. The simplest that comes to mind is instead of making a JSON-RPC call for debug trace transaction, expose a method through NAPI that returns the `DebugTraceResult` that is still in the repo.

crates/tools/js/benchmark/index.js
```diff
@@ -380,6 +379,7 @@ function preprocessConfig(config) {
   }
 
   config.providerConfig.minGasPrice = BigInt(config.providerConfig.minGasPrice);
+  config.providerConfig.enableRip7212 = false;
```
Wodann, line 382: All other changes apart from this line were a result of the formatter. I needed to make this change to allow old snapshots to continue working with the newly added configuration field.

cc: @agostbiro

agostbiro (approved):
Thanks, LGTM with one question

crates/edr_napi/test/issues.ts
```diff
@@ -0,0 +1,109 @@
...
+    let responseData;
+
+    if (typeof response.data === "string") {
+      responseData = JSON.parse(response.data);
+    } else {
+      responseData = response.data;
+    }
+
+    // Validate that we can query the response data without crashing.
+    const _json = new JsonStreamStringify(responseData);
```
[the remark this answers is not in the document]
Wodann: `JsonStreamStringify` is used to piecemeal convert a JSON object to a string, for objects that are too large to convert into a single String allocation.

There is one place in the Hardhat codebase that requires conversion to String and this illustrates a possibility of doing so without crashing due to an OOM error.

crates/edr_napi/index.d.ts
```diff
@@ -418,7 +418,8 @@ export class Provider {
   setVerboseTracing(verboseTracing: boolean): void
 }
 export class Response {
-  get json(): string
+  /** Returns the response data as a JSON string or a JSON object. */
+  get data(): string | any
```
fvictorio, line 422: In TypeScript, any type in a union with `any` becomes `any` (like multiplying by 0 or, more precisely, doing a set union with the universal set). So this is just `any` and not ideal.

I'm not 100% what the right type would be here. I think `string | object` would work. But if that's not easily doable with N-API, then it's fine for this to be `any` for now.

Wodann: I can look into it but the `serde_json::Value` translates to any

PR #574, opened 2024-08-01
ci: update collaborator check in the benchmarks workflow

galargh:
The updated check will evaluate true on:
1. Pushes to the default branch (this is unchanged)
2. Pull Requests where the head branch is in the same repository as the base branch (which implies the author's access write access to the repository)


agostbiro (approved):
Thanks looks good to me! I've rebased on top of the latest in `main` as there was a change there that changed the interface and it caused the CI run to fail. Please wait for the new CI run to finish successfully before merging.

PR #576, opened 2024-08-01
bump: revm with upstream changes

Wodann:
This incorporates Dragan's changes to my REVM PR

### DIFF
--- a/crates/edr_eth/src/chain_spec.rs
+++ b/crates/edr_eth/src/chain_spec.rs
@@ -9,7 +9,7 @@ use crate::{
 #[derive(Clone, Copy, Debug, Default, PartialEq, Eq, RlpEncodable)]
 pub struct L1ChainSpec;
 
-impl revm::primitives::ChainSpec for L1ChainSpec {
+impl revm::primitives::EvmWiring for L1ChainSpec {
     type Block = revm_primitives::BlockEnv;
 
     type Hardfork = revm_primitives::SpecId;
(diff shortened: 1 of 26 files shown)


crates/edr_eth/src/chain_spec.rs
```diff
@@ -9,7 +9,7 @@ use crate::{
 #[derive(Clone, Copy, Debug, Default, PartialEq, Eq, RlpEncodable)]
 pub struct L1ChainSpec;
 
-impl revm::primitives::ChainSpec for L1ChainSpec {
+impl revm::primitives::EvmWiring for L1ChainSpec {
```
agostbiro, line 12: It doesn't have to be addresses in this PR, just for the future: should we adapt our type names to reflect the name change in `revm`? In this case it'd be `L1EvMWiring` which sounds really weird, so I don't have a strong opinion on this.

agostbiro (approved):
Thanks, LGTM

crates/edr_eth/src/chain_spec.rs
Wodann: I don't really like the naming choice that REVM went with as I find it to be an opaque name (not clear what it means). I think `ChainSpec` better reflects that this is the spec of particular chain, so I was going to leave our names as they are.

agostbiro: By default, I'd try to match theirs to simplify things, but I agree that their choice isn't great, so I'd leave as is too 👍

PR #581, opened 2024-08-02
refactor: remove verbosity config

agostbiro:
I broke out this change from a larger PR coming to expose Solidity test config options to JS through the ffi.

The [verbosity config](https://book.getfoundry.sh/reference/config/testing#verbosity) in Foundry controls the level of detail displayed to the user, but it's also used internally to determine what info to collect. Since displaying info to the user will be the responsibility of Hardhat, this PR removes the verbosity config and replaces it with a boolean flag whether to enable tracing. This is sufficient to preserve previous functionality.

### DIFF
--- a/crates/foundry/evm/traces/src/decoder/mod.rs
+++ b/crates/foundry/evm/traces/src/decoder/mod.rs
@@ -72,13 +72,6 @@ impl CallTraceDecoderBuilder {
         self.with_known_contracts(identifier.contracts())
     }
 
-    /// Sets the verbosity level of the decoder.
-    #[inline]
-    pub fn with_verbosity(mut self, level: u8) -> Self {
-        self.decoder.verbosity = level;
-        self
-    }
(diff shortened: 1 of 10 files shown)


crates/foundry/evm/traces/src/decoder/mod.rs
agostbiro, line 1: This isn't actually used currently.

Wodann (approved):

fvictorio:
This PR seems to have unrelated changes that come from `main` :confused:

agostbiro:
> This PR seems to have unrelated changes that come from `main` 😕

Uhh thanks for catching it, I screwed up the merge. Fixed now, please re-review

agostbiro:
The failing tests are failing in the base branch as well. Will look into it tomorrow

fvictorio (approved):

PR #584, opened 2024-08-05
breaking change: rename response.json to response.data

fvictorio:


agostbiro (approved):
Thanks, LGTM

Wodann (approved):

.changeset/sour-donkeys-draw.md
```diff
@@ -0,0 +1,5 @@
+---
+"@nomicfoundation/edr": minor
+---
+
+Renamed `json` field in response to `data`
```
Wodann, line 5: Apart from this, the minor version change should also signal the introduction of `string | any` as a return type

PR #589, opened 2024-08-08
refactor: Port `ReturnData` and `StackTraceEntryType`

Xanewok:
This ports the `StackTraceEntryType` and related `...StackTraceEntry` interfaces from JS to Rust, which are the main external used after being processed by error inferrer et al. to be later encoded into a Node.js call stack trace.

Few things to note:
1. To use the ABI decoding originally pulled from `ethereumjs-abi`, I used `alloy-sol-types` and the macro that defines and generates corresponding error types at compile-time
1. since the ported types are just JS `interface`s, they have to be marked as `#[napi(object)]`, which means that they are passed as value types across the napi boundary
2. they originally used constant value members for the type interfaces as discriminants in JS. To simulate that, we introduce Rust `StackTraceEntryTypeConst` ZST that has the according value set as a const param and which implements `ToNapiValue` and `FromNapiValue` using the underlying const param value at run-time.
  The final JS type being a sum of all of these variants, we use a napi `Either24` (ugh) type to achieve the same thing. This retains the compatibility at the JS interface level.
  
@fvictorio please take a look; this is self-contained and acts as a base for the following error-inferrer/solidity-tracer port.

Pulled these commits for the local Hardhat patch (the companion branch is at https://github.com/NomicFoundation/hardhat/tree/refactor/port-stack-trace-entry-data):
- https://github.com/nomicFoundation/hardhat/commit/dc28ac9fcfa1f7096677f800226bad5f801fe89a
- https://github.com/nomicFoundation/hardhat/commit/b65376b10590219711309df2aabc9f02e62a9b50


fvictorio (approved):
LGTM, but most things in the description went over my head.

@Wodann would you mind taking a quick look? No need to review the code, just a light, conceptual review.

crates/edr_napi/src/trace/solidity_stack_trace.rs
```diff
@@ -0,0 +1,398 @@
...
+    )]
+    pub type_:
+        StackTraceEntryTypeConst<{ StackTraceEntryType::CONTRACT_CALL_RUN_OUT_OF_GAS_ERROR as u8 }>,
+    pub source_reference: Option<SourceReference>,
+}
+
+#[allow(dead_code)]
+// NOTE: This ported directly from JS for completeness, however the type must be
+// used verbatim in JS definitions because napi-rs does not store not allows to
+// reuse the same type unless fully specified at definition site.
```
Wodann, line 364: Typo? The comment isn't clear to me

Wodann (approved):
The code seems quite complicated, but based on the PR description that was necessary.

Hopefully an idiomatic rewrite will allow us to avoid the complexity in the future 🙂 

One comment, but otherwise LGTM

hardhat-tests/test/internal/hardhat-network/stack-traces/test.ts
```diff
@@ -280,9 +281,15 @@ function compareStackTraces(
 
   // if IR is enabled, we ignore callstack entries in the comparison
   if (isViaIR) {
-    trace = trace.filter(
-      (frame) => frame.type !== StackTraceEntryType.CALLSTACK_ENTRY
```
Xanewok, line 284: From my experience during this rewrite it was often the case that when the same type was removed and re-exported under the same name it was not recognized at run-time, even if the types were correct (meaning `tsc` was happy).

This was true for functions, classes and enums. What often helped was importing it under a different alias and then re-exporting it/using it with the same alias, removing the old code and then replacing with the old/actual name. Whether that's a problem with build cache or something I do not know (I tried cleaning node_modules, doing `pnpm install` etc. to no avail).

Rather than figuring out the underlying cause I decided not to do any further detective work and just apply the blunt workaround for now. Let's see if this is fixed with a subsequent change; I'd rather not spend more time here and focus on completing the porting work, instead.

crates/edr_napi/src/trace/solidity_stack_trace.rs
Xanewok: Yeah, I could have done better job at explaining this. I tried to be more in-depth about the context for this: the gist is that the type alias is mostly for the Rust side of the bindings. To actually type-check we would have to not use the type alias but the entire type as is, which is very bad, considering it takes 26 lines, so we accept the trade-off and use the type alias (which type-checks to `any` atm).

When we will limit the exposed API interface, we might specify the type literally in few select places if we deem the trade-off in readability to be acceptable :ok_hand:

Hope that makes sense and it's okay.

Xanewok:
Thanks for the reviews! I'll take the liberty and merge it after pushing a small fix and clarifying the note for the comment made. The next PR will be more substantial :grimacing:

hardhat-tests/test/internal/hardhat-network/stack-traces/test.ts
fvictorio: I didn't look into this, but @Xanewok did a very thorough research and couldn't find a good solution. [A follow-up PR](https://github.com/NomicFoundation/edr/pull/594) uses a different workaround, but it's still a workaround.

I don't think this is super important though, since it only affects tests' code.

PR #592, opened 2024-08-08
feat: solidity test runner config

agostbiro:
## Config Options

This PR exposes Solidity test runner config options to TypeScript from Rust through NAPI-RS.

The basis of the config options are the Foundry testing configuration options documented [here.](https://book.getfoundry.sh/reference/config/testing) There were some additional config options not covered there that are needed to construct a test runner. These are exposed based on the [Foundry config struct](https://github.com/NomicFoundation/edr/blob/103f3b12f73621867983cc58893ec5a8d212dc28/crates/foundry/config/src/lib.rs) that contains all configuration options (including build and deployment options which we don't need).

The following Foundry testing options are not exposed:

- [verbosity](https://book.getfoundry.sh/reference/config/testing#verbosity)
    - This controls the CLI output verbosity which is out of scope for EDR.
- [etherscan_api_key](https://book.getfoundry.sh/reference/config/testing#etherscan_api_key)
    - This is used for verifying deployed contracts and identifying unknown contracts in traces in fork mode. Both are out of scope for the EDR Solidity tests effort.
- [threads](https://book.getfoundry.sh/reference/config/testing#threads)
    - Number of logical cores to use for parallel test execution.
    - Not yet supported by forked Foundry version.
- [show_progress](https://book.getfoundry.sh/reference/config/testing#show_progress)
    - Whether to show test execution progress.
    - Not yet supported by forked Foundry version and probably out of scope for EDR.
- [names](https://book.getfoundry.sh/reference/config/testing#names)
    - Print compiled contract names.
    - Build-related, out of scope for EDR.
- [sizes](https://book.getfoundry.sh/reference/config/testing#sizes)
    - Print compiled contract sizes.
    - Build-related, out of scope for EDR.
- [gas_reports](https://book.getfoundry.sh/reference/config/testing#gas_reports)
    - Specifies the contracts to print gas reports for.
    - Displaying info to the user is Hardhat’s responsibility, so this is out of scope for EDR.
    - How to expose data for gas reporting is a separate issue
- [match-test, no-match-test, match-contract, no-match-contract, match-path, no-match-path](https://book.getfoundry.sh/reference/config/testing#match-test)
    - Only run test methods matching the above specifiers
    - Selecting which contracts to run is the responsibility of Hardhat, so these config options are out of scope for EDR.

## Error Callback

In addition to the config options, this PR brings another change to the Solidity test runner interface: an error callback. 

The reason for introducing this error callback is that building a test runner is async now to resolve forked EVM parameters, but the JS interface is sync and returns results with a progress callback. If we want to be able to propagate errors encountered while building a test runner we have two options:

1. Make the test runner async. This is possible, but we'd need to create a test runner class that takes the progress callback as argument first and then give it an async method for reasons explained in https://github.com/NomicFoundation/edr/pull/563. 
2. Expose an error callback.

The second option is simpler and composes well with the `Promise` wrapper on the TS-side (just have to pass the `reject` arg of the promise as the error callback), so I went with this option.

### DIFF
--- a/crates/edr_napi/src/solidity_tests.rs
+++ b/crates/edr_napi/src/solidity_tests.rs
@@ -19,7 +19,8 @@ use napi::{
 use napi_derive::napi;
 
 use crate::solidity_tests::{
-    artifact::ArtifactId, runner::build_runner, test_results::SuiteResult,
+    artifact::ArtifactId, config::SolidityTestRunnerConfigArgs, runner::build_runner,
+    test_results::SuiteResult,
 };
 
 /// Executes Solidity tests.
@@ -28,20 +29,28 @@ use crate::solidity_tests::{
 /// The progress callback will be called with the results of each test suite.
 /// It is up to the caller to track how many times the callback is called to
 /// know when all tests are done.
+/// The error callback is called if an invalid configuration value is provided.
 // False positive from Clippy. The function is exposed through the FFI.
 #[allow(dead_code)]
 #[napi]
 pub fn run_solidity_tests(
     artifacts: Vec<Artifact>,
     test_suites: Vec<ArtifactId>,
-    gas_report: bool,
+    config_args: SolidityTestRunnerConfigArgs,
     #[napi(ts_arg_type = "(result: SuiteResult) => void")] progress_callback: JsFunction,
+    #[napi(ts_arg_type = "(error: Error) => void")] error_callback: JsFunction,
 ) -> napi::Result<()> {
     let results_callback_fn: ThreadsafeFunction<_, ErrorStrategy::Fatal> = progress_callback
         .create_threadsafe_function(
             // Unbounded queue size
             0,
             |ctx: ThreadSafeCallContext<SuiteResult>| Ok(vec![ctx.value]),
+        )?;
+    let error_callback_fn: ThreadsafeFunction<_, ErrorStrategy::Fatal> = error_callback
+        .create_threadsafe_function(
+            // Unbounded queue size
+            0,
+            |ctx: ThreadSafeCallContext<napi::Error>| Ok(vec![ctx.value]),
         )?;
 
     let known_contracts: ContractsByArtifact = artifacts
@@ -55,8 +64,6 @@ pub fn run_solidity_tests(
         .map(TryInto::try_into)
         .collect::<Result<Vec<_>, _>>()?;
 
-    let runner = build_runner(&known_contracts, test_suites, gas_report)?;
-
     let (tx_results, mut rx_results) = tokio::sync::mpsc::unbounded_channel::<(
         foundry_common::ArtifactId,
         forge::result::SuiteResult,
@@ -64,6 +71,28 @@ pub fn run_solidity_tests(
 
     let runtime = runtime::Handle::current();
     runtime.spawn(async move {
+        let runner = match build_runner(&known_contracts, test_suites, config_args).await {
+            Ok(runner) => runner,
+            Err(error) => {
+                let call_status =
+                    error_callback_fn.call(error, ThreadsafeFunctionCallMode::Blocking);
+                // This should always succeed since we're using an unbounded queue. We add an
+                // assertion for completeness.
+                assert!(
+                    matches!(call_status, napi::Status::Ok),
+                    "Failed to call callback with status {call_status:?}"
+                );
+                return;
+            }
+        };
+
+        // Returns immediately after test suite execution is started
+        runner.test_hardhat(
+            Arc::new(known_contracts),
+            Arc::new(EverythingFilter),
+            tx_results,
+        );
+
         while let Some(name_and_suite_result) = rx_results.recv().await {
             let callback_arg = name_and_suite_result.into();
             // Blocking mode won't block in our case because the function was created with
@@ -78,13 +107,6 @@ pub fn run_solidity_tests(
             );
         }
     });
-
-    // Returns immediately after test suite execution is started
-    runner.test_hardhat(
-        Arc::new(known_contracts),
-        Arc::new(EverythingFilter),
-        tx_results,
-    );
 
     Ok(())
 }
--- a/crates/edr_napi/src/solidity_tests/config.rs
+++ b/crates/edr_napi/src/solidity_tests/config.rs
@@ -1,21 +1,691 @@
-use std::{collections::HashMap, path::PathBuf};
+use std::{collections::HashMap, fmt::Debug, path::PathBuf};
 
-use alloy_primitives::{address, Address};
+use alloy_primitives::{address, hex, Address, B256};
 use edr_eth::U256;
 use forge::{
+    fork::CreateFork,
     inspectors::cheatcodes::CheatsConfigOptions,
     opts::{Env as EvmEnv, EvmOpts},
 };
-use foundry_config::{
-    cache::StorageCachingConfig, FsPermissions, FuzzConfig, GasLimit, InvariantConfig, RpcEndpoint,
-    RpcEndpoints,
+use foundry_config::{FsPermissions, FuzzConfig, InvariantConfig, RpcEndpoint, RpcEndpoints};
+use napi::{
+    bindgen_prelude::{BigInt, Buffer},
+    Either, Status,
 };
+use napi_derive::napi;
+
+use crate::cast::TryCast;
+
+/// Default address for `tx.origin` for Foundry
+///
+/// `0x1804c8AB1F12E6bbf3894d4083f33e07309d1f38`
+const DEFAULT_SENDER: Address = address!("1804c8AB1F12E6bbf3894d4083f33e07309d1f38");
+
+/// Solidity test runner configuration arguments exposed through the ffi.
+/// Docs based on https://book.getfoundry.sh/reference/config/testing
+#[napi(object)]
+#[derive(Clone)]
+pub struct SolidityTestRunnerConfigArgs {
+    /// The absolute path to the project root directory.
+    /// Relative paths in cheat codes are resolved against this path.
+    pub project_root: String,
+    /// Configures the permissions of cheat codes that access the file system.
+    pub fs_permissions: Option<Vec<PathPermission>>,
+    /// Whether to collect traces. Defaults to false.
+    pub trace: Option<bool>,
+    /// Whether to collect debug info. Defaults to false.
+    pub debug: Option<bool>,
+    /// Address labels for traces. Defaults to none.
+    pub labels: Option<Vec<AddressLabel>>,
+    /// Whether to enable isolation of calls. In isolation mode all top-level
+    /// calls are executed as a separate transaction in a separate EVM
+    /// context, enabling more precise gas accounting and transaction state
+    /// changes.
+    /// Defaults to false.
+    pub isolate: Option<bool>,
+    /// Whether or not to enable the ffi cheatcode.
+    /// Warning: Enabling this cheatcode has security implications, as it allows
+    /// tests to execute arbitrary programs on your computer.
+    /// Defaults to false.
+    pub ffi: Option<bool>,
+    /// The value of `msg.sender` in tests as hex string.
+    /// Defaults to `0x1804c8AB1F12E6bbf3894d4083f33e07309d1f38`.
+    pub sender: Option<Buffer>,
+    /// The value of `tx.origin` in tests as hex string.
+    /// Defaults to `0x1804c8AB1F12E6bbf3894d4083f33e07309d1f38`.
+    pub tx_origin: Option<Buffer>,
+    /// The initial balance of the sender in tests.
+    /// Defaults to `0xffffffffffffffffffffffff`.
+    pub initial_balance: Option<BigInt>,
+    /// The value of `block.number` in tests.
+    /// Defaults to `1`.
+    pub block_number: Option<BigInt>,
+    /// The value of the `chainid` opcode in tests.
+    /// Defaults to `31337`.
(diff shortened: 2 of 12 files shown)


crates/edr_napi/src/solidity_tests/config.rs
```diff
@@ -1,21 +1,697 @@
...
+    }
+}
+
+/// Settings to configure caching of remote
+#[napi(object)]
+#[derive(Clone, Debug)]
+pub struct StorageCachingConfig {
+    /// Chains to cache. Either all or none or a list of chain names, e.g.
+    /// ["optimism", "mainnet"].
+    pub chains: Either<CachedChains, Vec<String>>,
```
agostbiro, line 540: Either is a workaround for lack of enums with fields in JS

agostbiro:
The test failures are unrelated and fixed in https://github.com/NomicFoundation/edr/pull/598

fvictorio (approved):

crates/edr_napi/src/solidity_tests.rs
```diff
@@ -55,15 +64,35 @@ pub fn run_solidity_tests(
...
+                    error_callback_fn.call(error, ThreadsafeFunctionCallMode::Blocking);
+                // This should always succeed since we're using an unbounded queue. We add an
+                // assertion for completeness.
+                assert!(
+                    matches!(call_status, napi::Status::Ok),
+                    "Failed to call callback with status {call_status:?}"
+                );
+                return;
+            }
+        };
```
Wodann, line 87: > The reason for introducing this error callback is that building a test runner is async now to resolve forked EVM parameters, but the JS interface is sync and returns results with a progress callback. If we want to be able to propagate errors encountered while building a test runner we have two options:
>
> 1. Make the test runner async. This is possible, but we'd need to create a test runner class that takes the progress callback as argument first and then give it an async method for reasons explained in https://github.com/NomicFoundation/edr/pull/563.

I don't understand how the problem described in the other pull request applies here. To me it seems like a different use case, which wouldn't suffer from the same problem as describe in PR #563.

We have many examples of `async fn` in our N-API that have `.await` calls that work just fine. Can't this function also be made async for the construction of the build runner and only afterwards spawn the background process for parallelisation - which does suffer from the issue in PR #563?

That seems more idiomatic than the error callback.

Am I overlooking something?

crates/edr_napi/src/solidity_tests/config.rs
```diff
@@ -1,21 +1,697 @@
...
+    /// Warning: Enabling this cheatcode has security implications, as it allows
+    /// tests to execute arbitrary programs on your computer.
+    /// Defaults to false.
+    pub ffi: Option<bool>,
+    /// The value of `msg.sender` in tests as hex string.
+    /// Defaults to `0x1804c8AB1F12E6bbf3894d4083f33e07309d1f38`.
+    pub sender: Option<Buffer>,
+    /// The value of `tx.origin` in tests as hex string.
+    /// Defaults to `0x1804c8AB1F12E6bbf3894d4083f33e07309d1f38`.
+    pub tx_origin: Option<Buffer>,
```
Wodann, line 56: What does origin mean in this context?

I see that there is a separate `sender` config option, so it must be something distinct 🤔

crates/edr_napi/src/solidity_tests/config.rs
```diff
@@ -1,21 +1,697 @@
...
+    /// The memory limit of the EVM in bytes.
+    /// Defaults to 33_554_432 (2^25 = 32MiB).
+    pub memory_limit: Option<BigInt>,
+    // TODO https://github.com/NomicFoundation/edr/issues/591
+    /// If set to true, then block data from RPC endpoints in tests will not be
+    /// cached. Otherwise, the data is cached to
+    /// `$HOME/.foundry/cache/<chain id>/<block number>`.
+    /// Defaults to false.
+    pub no_storage_caching: Option<bool>,
+    /// If set, all tests are ran in fork mode using this url or remote name.
```
Wodann, line 99:
```suggestion
    /// If set, all tests are run in fork mode using this url or remote name.
```

Wodann (changes requested):
Overall looks good.

I primarily would like to better understand why we cannot use an `async fn run_solidity_tests` for calling an async constructor, as this pattern is used in our other N-API bindings.

crates/edr_napi/src/solidity_tests/config.rs
fvictorio: `origin` and `sender` are different things in the EVM.

Let's say that you use account `A` to call contract `Foo`, which in turn calls contract `Bar`. In the context of `Bar`'s execution, the sender is `Foo`, but the origin of the transaction is `A`.

In solidity, these values can be obtained with `tx.origin` and `msg.sender`. In the EVM, the opcodes are `0x32 ORIGIN` and `0x33 CALLER`.

crates/edr_napi/src/solidity_tests.rs
agostbiro: The issue is that `run_solidity_tests` takes a progress callback argument as a `napi::JsFunction` and `napi::JsFunction` is `!Send`, so it cannot be passed into an async function as argument on the Rust side. 

So our options are either working around this by introducing a test runner class that takes the progress callback as argument in a sync constructor and exposes an async method to start running the tests that'd throw if there is an error when constructing the test runner. Or alternatively add an error callback argument which is quite common in NodeJS callback-based interfaces. Exposing an error callback lead to simpler code on both sides of the FFI, so I went with that.

crates/edr_napi/src/solidity_tests/config.rs
Wodann: Thanks!

crates/edr_napi/src/solidity_tests.rs
Wodann: It's not self-evident but N-API provides a mechanism to work around the `!Send` issue. We use it for [the constructor of the `Provider`](https://github.com/NomicFoundation/edr/blob/main/crates/edr_napi/src/provider.rs#L32), which is also `async`.

To validate that my idea actually works, I implemented it [here](https://github.com/NomicFoundation/edr/pull/600).

The `run_solidity_tests` function is a function that "spawns a Solidity test runner". In line with the rest of our API, this change ensures that the function will either spawn the test runner or throw an exception. All of the additional progress then works with the callback.

agostbiro: Thanks that's good to know! I'm on mobile, could you also update the js test in edr_napi in your PR to see how that'd look like?

Wodann: Done https://github.com/NomicFoundation/edr/pull/600/commits/6d597cf2c604dc965d3529828496270a3569cffd

agostbiro: Thanks! @fvictorio could you review the JS changes? Is the promise going to be rejected if the async executor throws?

fvictorio: I don't think I can answer this just by looking at the changes in that commit, since it only has changes in the `.d.ts` file.

The way javascript works here is a bit weird. If you have this code:

```js
function foo() {
  throw new Error("foo")
  return Promise.resolve(42);
}

async function main() {
  const answer = foo().catch(() => {});
}

main()
```

Then `main` will throw, because `foo()` doesn't return a rejected promise. But if you add `async` to `foo`'s declaration:

```js
async function foo() {
  throw new Error("foo")
  return Promise.resolve(42);
}

async function main() {
  const answer = foo().catch(() => {});
}

main()
```

then the error will be converted into a rejected promise, and `main` won't throw. So whether or not the function will return a rejected promise or throw an error is hard to answer without knowing what N-API does under the hood. The fact that there weren't any changes in the `index.js` file makes me think that it _might_ throw (but, again, hard to know; it might as well return a rejected promise).

But this should be relatively easy to manually test to see what's the behavior, right?

fvictorio:
+1 to renaming the function to `spawnSolidityTestRunner` btw. I was thinking "startSolidityTestRunner", but I think I like "spawn" more.

crates/edr_napi/src/solidity_tests.rs
agostbiro: Thanks, I added a test for it in https://github.com/NomicFoundation/edr/commit/94ca87a63a0b482630b9d9446d3f7691dfac50ac and it looks like errors aren't thrown with the async executor unfortunately. Using a sync executor and passing `reject` to `catch` works, but that's just a more complicated way of passing an error callback, so I'd respectfully decline this change request.

agostbiro:
> +1 to renaming the function to spawnSolidityTestRunner btw. I was thinking "startSolidityTestRunner", but I think I like "spawn" more.

Ok, I'll do that in a follow up PR.

agostbiro:
Actually on a second thought, `spawnSolidityTestRunner` is kinda misleading, because I'd expect based on this name that I spawn something and then I make requests to it to execute tests. I don't think it makes sense to spend more time on this now, and we'll want to do an interface review anyway with the Hardhat team, so I think we can come up with a better name then.

crates/edr_napi/src/solidity_tests.rs
fvictorio: > error aren't thrown with the async executor

Do you mean that the function doesn't throw an error but the promise rejects, the opposite, or that neither happens?

In any case, I'm ok with keeping the error callback approach, following previous decisions about not caring too much about the API here as long as Hardhat is the only consumer, and the fact that at some point we'll offer a wrapper with a nicer API unconstrained by `napi-rs` limitations.

fvictorio:
Sure, I'm ok with postponing that discussion!

crates/edr_napi/src/solidity_tests.rs
agostbiro: > Do you mean that the function doesn't throw an error but the promise rejects, the opposite, or that neither happens?

The promise just hangs if there is an error: https://github.com/NomicFoundation/edr/actions/runs/10416449901/job/28849161857#step:8:42

And this is how the napi function is invoked: https://github.com/NomicFoundation/edr/commit/94ca87a63a0b482630b9d9446d3f7691dfac50ac#diff-ed50e3e633d981c1abfaec950d3a22878e0f1a4ee8a83722fd1d86f5bb001616R13

> In any case, I'm ok with keeping the error callback approach, following previous decisions about not caring too much about the API here as long as Hardhat is the only consumer, and the fact that at some point we'll offer a wrapper with a nicer API unconstrained by napi-rs limitations.

Ok cheers

Wodann: > Thanks, I added a test for it in https://github.com/NomicFoundation/edr/commit/94ca87a63a0b482630b9d9446d3f7691dfac50ac and it looks like errors aren't thrown with the async executor unfortunately.

That's concerning for all of our other async code though, isn't it?

I'm curious to try one thing. In this code I'm explicitly rejecting the `Promise` instead of resolving it with an error. I'll test whether that makes a difference, because otherwise all of our async code might be at risk of this problem.

Wodann: So I tried a `try..catch` block around the `await runSolidityTests` and logged the error. It was in fact correctly returning the error.

The way that `new Promise` works is that unless you are explicitly forwarding the thrown error to the newly defined promises `reject` function, it will not reject the outer Promise when an inner `Promise` is rejected. In the sync code, you are explicitly doing this using the `.catch(reject)`. Adding that to the async code ensures that the promise is forwarded.

If you want to use `await runSolidityTests`, you have to define an `async fn` which does the piping of inner `Promise` rejections for you.

I found these two articles helpful:
- https://javascript.info/promise-error-handling#unhandled-rejections
- https://stackoverflow.com/questions/45348955/using-await-within-a-promise

agostbiro: > That's concerning for all of our other async code though, isn't it?

Normally you wouldn't wrap a JS function that returns a promise in another promise. The issue here is that you can pass an async executor to a  `Promise` and call `await runSolidityTests(...)` in the executor of the promise, but if that call throws, the `Promise` won't be rejected. 

I think it's quite clear now that making `runSolidityTests` async is not a good idea, because it doesn't compose well with its usage. Could you approve please so that I can merge it and then you can run experiments as a separate issue?

Wodann: > I think it's quite clear now that making runSolidityTests async is not a good idea, because it doesn't compose well with its usage.

In my [previous comment](https://github.com/NomicFoundation/edr/pull/592#discussion_r1719777489) I explained why it was failing. It wasn't a result of `runSolidityTests` being async, but instead an error in the piping of the promises rejection; something that's part of Javascript's design of the `Promise` API.

Wodann (approved):
I proposed a solution to the mentioned `async` problem with N-API [here](https://github.com/NomicFoundation/edr/pull/592#discussion_r1719777489). I'm not sure whether this also circumvents the original issue that we had (about a month ago) relating to `process.exit()`.

I had a discussion with @fvictorio about Javascript's best practices around callbacks. If an API is explicitly designed to receive callbacks to track progress, it should only use callbacks. No promises.

In that case, could you maybe just rename the `errorCallback` to indicate that it will not be called for all errors but just for the spawning error?

I'm a proponent of renaming the `runSolidityTests` function to `spawnSolidityTestRunner` to make that clearer as well.

_edited based on discussion with Franco_

agostbiro:
> In that case, could you maybe just rename the errorCallback to indicate that it will not be called for all errors but just for the spawning error?

The doc says explicitly: "The error callback is called if an invalid configuration value is provided."

PR #596, opened 2024-08-09
refactor: Port stack-traces/debug.ts

Xanewok:
Based on #593

This re-implements the structured console.log printing from the original Hardhat.ts file. While not insanely useful on its own, it is the last file that references the underlying compiler model classes, and so this is the last brick that's necessary to cut and reduce the JS API exposure.

To test it, I modified the corresponding `test.ts` file in the Hardhat repo (using local EDR package link) to print both versions one after another (both message and stack traces) for select tests (It's useful to specify `"only": true"` and `"print": true` in the select `test.json` files). They're mostly the same with the exception of not bolding the JUMPs anymore but I didn't want to pull a new dependency for something so small.

To implement printing the entries as JSON in `printStackTrace` I had to add some `Serialize` impls - these are just for the debugging purposes and we do not aim to make this roundtrippable at the moment as there's no real use case for it yet.

### DIFF
--- a/crates/edr_napi/index.d.ts
+++ b/crates/edr_napi/index.d.ts
@@ -387,6 +387,8 @@ export interface SourceMap {
   location: SourceMapLocation
   jumpType: JumpType
 }
+export function printMessageTrace(trace: PrecompileMessageTrace | CallMessageTrace | CreateMessageTrace, depth: number): void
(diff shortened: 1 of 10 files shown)


fvictorio (approved):
This is just for internal use so I'm ok with merging it without other reviews :+1:

PR #599, opened 2024-08-14
refactor: decouple Logger and Provider generics

Wodann:
This change decouples the logger's error type from the provider. This prevents the necessity of knowing the `Logger`'s error type when implementing traits for the provider-related types (errors, data, interval miner, etc.)

### DIFF
--- a/crates/edr_napi/src/logger.rs
+++ b/crates/edr_napi/src/logger.rs
@@ -128,8 +128,6 @@ impl Logger {
 impl edr_provider::Logger<L1ChainSpec> for Logger {
     type BlockchainError = BlockchainError<L1ChainSpec>;
 
-    type LoggerError = LoggerError;
-
     fn is_enabled(&self) -> bool {
         self.collector.is_enabled
     }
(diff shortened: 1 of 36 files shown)


agostbiro (approved):
Thanks, LGTM overall.

As a note, I'm not that big on `Box<dyn std::error::Error + Send + Sync>>` in libraries due to type erasure, but here the only consumer is the JS interface and there is type erasure anyway due to the FFI, so I don't think it's worth spending time on having richer error type in the logger.

Wodann:
> Thanks, LGTM overall.
> 
> As a note, I'm not that big on `Box<dyn std::error::Error + Send + Sync>>` in libraries due to type erasure, but here the only consumer is the JS interface and there is type erasure anyway due to the FFI, so I don't think it's worth spending time on having richer error type in the logger.

I agree with you that type erasure should be avoided.

With a refactor I did yesterday and an idea I had overnight, there is one alternative approach that I could try to retain the `LoggerErrorT`. It will be especially beneficial if in Hardhat v3 we allow richer error types to be returned instead of strings.

I'm happy to give that a try now or to create an issue for it. Up to you.

agostbiro:
I'm ok with it as is

Wodann:
I created a [follow-up issue](https://github.com/NomicFoundation/edr/issues/602).

PR #600, opened 2024-08-14
refactor: make run_solidity_tests async

Wodann:


crates/edr_napi/test/solidity-tests.ts
```diff
@@ -18,10 +18,10 @@ describe("Solidity Tests", () => {
     // All artifacts are test suites.
     const testSuites = artifacts.map((artifact) => artifact.id);
 
-    const results: Array<SuiteResult> = await new Promise((resolve, reject) => {
+    const results: Array<SuiteResult> = await new Promise(async (resolve) => {
       const resultsFromCallback: Array<SuiteResult> = [];
 
-      runSolidityTests(
+      await runSolidityTests(
```
Wodann, line 24: As mentioned on a previous PR, I'd rename this function to `spawnSolidityTestRunner` to make it clear that this function doesn't actually finish when the tests have run but merely spawns the test runner.

PR #608, opened 2024-08-19
feat: Optimism provider

Wodann:
This PR adds all required trait implementations to be able to use Optimism's chain spec for a provider.

> NOTE: To expedite exposing Optimism for testing in Hardhat, I didn't spend time on adding Optimism-specific provider tests yet.


agostbiro (approved):
Thanks, LGTM

PR #617, opened 2024-08-22
chore: Bump @napi-rs/cli to fix duplicated napi typedefs

Xanewok:
See <https://github.com/napi-rs/napi-rs/pull/2088>

We encountered that in Slang - sometimes when you hop between branches and do rebuilds of `napi build` you get stale index.d.ts info due to cached typedefs between macro invocations. That's not guaranteed to work and backfires once you change the order of exports (e.g. changing the module structure, order etc.) as that gives duplicated type definitions.

The 2.18.4 also contains the missing `declare` attribute when exporting but I didn't pull that in to reduce churn across our feature branches; I assume that all of them modifies the index.d.ts substantially and so let's do that separately at our leisure.

### DIFF
--- a/crates/edr_napi/package.json
+++ b/crates/edr_napi/package.json
@@ -2,7 +2,7 @@
   "name": "@nomicfoundation/edr",
   "version": "0.5.2",
   "devDependencies": {
-    "@napi-rs/cli": "^2.18.1",
+    "@napi-rs/cli": "^2.18.3",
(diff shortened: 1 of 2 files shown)


Xanewok:
I'm not a TS expert but I think it's slightly more correct to export with `declare` as this informs TS that a given definition has been defined elsewhere already (i.e. the native addon). I wouldn't worry about it but I can open an issue to not forget about it.

PR #619, opened 2024-08-22
refactor: remove Foundry `test-utils` crate

agostbiro:
Remove the Foundry `test-utils` crate and merge functionality into the `edr_test_utils` crate. There are a lot of changes, so I tried to split it up into individually reviewable commits. Hopefully it's accessible this way.


Cargo.toml
```diff
@@ -160,6 +159,8 @@ wildcard-imports = "warn"
 zero_sized_map_values = "warn"
 
 [workspace.dependencies]
+edr_test_utils = { path = "crates/edr_test_utils" }
```
Wodann, line 162: Shouldn't this always be a locally specified dev dependency?

To the best of my knowledge, [workspace dependencies are non-optional](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-dependencies-table). Making this a dependency requires compilation of `edr_test_utils` for non-test configurations.

crates/edr_test_utils/Cargo.toml
```diff
@@ -6,6 +6,14 @@ edition.workspace = true
...
 serde.workspace = true
 serde_json.workspace = true
+foundry-common.workspace = true
+foundry-compilers = { workspace = true, features = ["project-util"] }
+
+fd-lock = "4.0.0"
+once_cell = "1"
+regex = "1"
+tracing-subscriber = { workspace = true, features = ["env-filter", "fmt"] }
+rand.workspace = true
```
Wodann, line 16:
```suggestion
rand.workspace = true
regex = "1"
tracing-subscriber = { workspace = true, features = ["env-filter", "fmt"] }
```

alphabetic ordering

Wodann (changes requested):
Thank you for splitting it into separate commits. It made reviewing a lot easier! ❤️ 

I have one small suggestion and one concern

Cargo.toml
agostbiro: > To the best of my knowledge, [workspace dependencies are non-optional](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-dependencies-table)

I think that just means that you cannot mark it as optional, e.g. `edr_test_utils = { workspace = true, optional = true }` doesn't work. But dev dependencies cannot be marked as optional either and `edr_test_utils` is always used as a dev dependency, so I don't think this is a problem.

> Making this a dependency requires compilation of `edr_test_utils` for non-test configurations.

I've tried this in the branch by doing `cd crates/edr_napi && cargo clean && pnpm build` and it didn't compile `edr_test_utils`, so I think it's ok.

> Shouldn't this always be a locally specified dev dependency?

As it seems it won't force compilation of `edr_test_utils` for non-test builds, and it saves typing out the path where it's used, I kinda like this pattern (borrowed it from Foundry).

crates/edr_test_utils/Cargo.toml
agostbiro: Thanks, fixed in [32ab634](https://github.com/NomicFoundation/edr/pull/619/commits/32ab634bfb021e5404c0bbb63c096935ebc28900)

Wodann (approved):

PR #620, opened 2024-08-23
fix: build edr_napi with test profile when running hardhat tests

fvictorio:
For some reason, stack traces test don't work when `edr_napi` is built with the debug profile. Since we were using the release profile before anyway, I think we can just continue using it to get this fixed.

### DIFF
--- a/hardhat-tests/package.json
+++ b/hardhat-tests/package.json
@@ -53,7 +53,7 @@
   "repository": "github:NomicFoundation/edr",
   "scripts": {
     "build": "tsc --build --force --incremental . && pnpm run build:edr",
-    "build:edr": "cd ../crates/edr_napi && pnpm build:debug",
+    "build:edr": "cd ../crates/edr_napi && pnpm build:test",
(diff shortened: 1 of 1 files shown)


Xanewok:
Could you explain a bit? The tests in general seem to be working as per latest CI run: https://github.com/NomicFoundation/edr/actions/runs/10503186671/job/29096060736.

Do you mean that this may help catch the spurious errors in CI or something different?

agostbiro (approved):

agostbiro:
> For some reason, stack traces test don't work when `edr_napi` is built with the debug profile. Since we were using the release profile before anyway, I think we can just continue using it to get this fixed.

This is just in the `feat/solidity-tests` branch: https://github.com/NomicFoundation/edr/pull/524

fvictorio:
Yes, what Agost said. This is something that broke in the solidity tests feature branch. `main` is working fine.

Xanewok:
> > Could you explain a bit? The tests in general seem to be working as per latest CI run: https://github.com/NomicFoundation/edr/actions/runs/10503186671/job/29096060736.
> > Do you mean that this may help catch the spurious errors in CI or something different?
> 
> This is just in the `feat/solidity-tests` branch: #524

Ah, didn't see that. Thanks and sorry for the noise!

PR #626, opened 2024-08-26
disregard: fail benchmark on purpose

agostbiro:


crates/tools/js/benchmark/src/index.ts
```diff
@@ -320,7 +320,8 @@ async function benchmarkScenario(
     }
   }
 
-  const timeMs = performance.now() - start;
+  // TODO remove: artificially inflate measurements
+  const timeMs = (performance.now() - start) * 1.2;
```
agostbiro, line 324: Inflated measurement

PR #629, opened 2024-08-26
feat: rpc cache path for Solidity tests

agostbiro:
- Removes the `no_storage_caching` config option.
- Replace it with optional RPC cache path config option. If this is none, then no RPC calls will be cached, otherwise data is cached to `<rpc_cache_path>/<chain id>/<block number>`.
- Removes code that became unused due to the changes


crates/edr_napi/src/solidity_tests/config.rs
```diff
@@ -104,6 +98,11 @@ pub struct SolidityTestRunnerConfigArgs {
     /// Map of RPC endpoints from chain name to RPC urls for fork cheat codes,
     /// e.g. `{ "optimism": "https://optimism.alchemyapi.io/v2/..." }`
     pub rpc_endpoints: Option<HashMap<String, String>>,
+    /// Optional RPC cache path. If this is none, then no RPC calls will be
+    /// cached, otherwise data is cached to `<rpc_cache_path>/<chain
+    /// id>/<block number>`. Caching can be disabled for specific chains
+    /// with `rpc_storage_caching`.
+    pub rpc_cache_path: Option<String>,
```
Wodann, line 105: I followed the instantiations of this member variable and it seems that this path does not include the RPC endpoint url.

In EDR we explicitly separate cache entries between different endpoint urls (e.g. providers). Is that not a concern for the test runner's caching mechanism?

Wodann (approved):
Overall LGTM. One question

crates/edr_napi/src/solidity_tests/config.rs
fvictorio: I'd guess that's not super important while we are using Foundry's EVM, and that when we replace it with EDR's EVM we'll get that behavior automatically?

Is that the case @agostbiro?

fvictorio (approved):

crates/edr_napi/src/solidity_tests/config.rs
Wodann: Thanks for the explanation. Makes sense!

PR #633, opened 2024-08-27
refactor: chain-specific N-API

Wodann:
This PR makes most of the N-API code chain-agnostic by introducing runtime polymorphism for providers.

A notable exception is the `ExecutionResult`, which only supports the `L1ChainSpec`. Going forward, this will be converted to support the `GenericChainSpec`. This is only used in the `RawTrace` for solidity stack traces and community plugins. For now, we've opted to not support these features for other chain types.

This PR introduces breaking changes in the N-API:
- The `ProviderConfig::hardfork` and `ProviderConfig::chains` variables were changed to use a `string` instead of a `SpecId` to be chain-agnostic.
- The constructor of `Provider` was removed. This has been replaced with with functions on the `EdrContext`.
- The `Provider::setCallOverrideCallback` and `Provider::setVerboseTracing` functions were made `async` to ensure that blocking execution happens on a background thread.


crates/edr_napi/src/spec.rs
```diff
@@ -0,0 +1,139 @@
...
+pub struct Response {
+    // N-API is known to be slow when marshalling `serde_json::Value`s, so we try to return a
+    // `String`. If the object is too large to be represented as a `String`, we return a `Buffer`
+    // instead.
+    data: Either<String, serde_json::Value>,
+    /// When a transaction fails to execute, the provider returns a trace of the
+    /// transaction.
+    solidity_trace: Option<RawTrace>,
+    /// This may contain zero or more traces, depending on the (batch) request
+    traces: Vec<RawTrace>,
```
agostbiro, line 23: It'd be good to have notes here that traces are only returned for L1.

agostbiro (changes requested):
Thanks, LGTM overall and congrats on reaching this milestone! I have two concerns:

- Some nice to have improvements for the documentation. I marked these with comments.
- It looks like there is a [regression](https://github.com/NomicFoundation/edr/actions/runs/10586580928/job/29335656501?pr=633) on the `synthetix` scenario in the benchmarks. I think this needs to be fixed. See more details below.


Error in the benchmark run:

```
Scenario synthetix_9a3a109f should not fail at indexes 1202223,1204446,1206829,1206985,1208654,1209510,1211270,1212167,1214012,1214956
```

Looking at the requests, all of them are `eth_getBlockByNumber`:

```
{"method":"eth_getBlockByNumber","params":["0x114b",false]}
{"method":"eth_getBlockByNumber","params":["0x114d",false]}
{"method":"eth_getBlockByNumber","params":["0x114d",false]}
{"method":"eth_getBlockByNumber","params":["latest",false]}
{"method":"eth_getBlockByNumber","params":["0x114e",false]}
{"method":"eth_getBlockByNumber","params":["latest",false]}
{"method":"eth_getBlockByNumber","params":["0x1151",false]}
{"method":"eth_getBlockByNumber","params":["latest",false]}
{"method":"eth_getBlockByNumber","params":["0x1150",false]}
{"method":"eth_getBlockByNumber","params":["latest",false]}
```

(Generated with `zcat < crates/tools/scenarios/synthetix_9a3a109f.jsonl.gz| sed -n '1202223p;1204446p;1206829p;1206985p;1208654p;1209510p;1211270p;1212167p;1214012p;1214956p'`)

I've reran the benchmark job just in case it's a transient failure.

agostbiro (approved):

Xanewok (approved):
Changes as described in the OP make sense at a cursory glance but I'd probably have to get more context on the underlying feature to do a more thorough review. LGTM!

crates/edr_napi/src/spec.rs
Wodann: Addressed in https://github.com/NomicFoundation/edr/pull/633/commits/00209047e80270572b0e84be664fd175a14d8b65

Wodann:
> Edit: sed line numbers are 1-indexed, but the failures reported are 0-indexed. Fixed this.

@agostbiro This might be useful to know in the future. It turned out to be off by 2. 1 extra because the first line contains the config.

agostbiro:
> @agostbiro This might be useful to know in the future. It turned out to be off by 2. 1 extra because the first line contains the config.

Right, that's good to know!

PR #638, opened 2024-08-30
ci: add workflow for testing Optimism every night

Wodann:
This PR adds support for replaying Optimism on-chain blocks to the CLI app and adds a GitHub workflow for nightly replays of the latest Optimism block.

This PR contains a breaking change to the CLI:
- `--chain-id` was removed as it can be retrieved from the remote
- `--chain-type` was added, replacing the meaning of `-c` from `--chain-id` to `--chain-type`

Once merged, this PR commit should be cherry-picked and applied to `main` to ensure that the `on.schedule` works.
> Scheduled workflows will only run on the default branch.
>
> https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#schedule

### DIFF
--- a/.github/workflows/test-recent-mainnet-block.yml
+++ b/.github/workflows/test-recent-mainnet-block.yml
@@ -35,7 +35,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: replay-block
-          args: -u ${{ secrets.ALCHEMY_URL }} --chain-id 1 --chain-type l1
+          args: -u ${{ secrets.ALCHEMY_URL }} -c l1
 
       - name: Notify failures
         if: failure()
--- a/.github/workflows/test-recent-optimism-block.yml
+++ b/.github/workflows/test-recent-optimism-block.yml
@@ -25,7 +25,7 @@ jobs:
       - uses: Swatinem/rust-cache@v2
 
       - name: Cache EDR RPC cache
-        uses: actions/cache@v2
+        uses: actions/cache@v4
         with:
           path: |
             **/edr-cache
--- a/.github/workflows/test-recent-optimism-block.yml
+++ b/.github/workflows/test-recent-optimism-block.yml
@@ -4,7 +4,6 @@ on:
   schedule:
     - cron: "0 */8 * * *"
   workflow_dispatch:
-  push:
 
 defaults:
   run:
--- a/.github/workflows/test-recent-optimism-block.yml
+++ b/.github/workflows/test-recent-optimism-block.yml
@@ -1,4 +1,4 @@
-name: Run a recent full block in the Hardhat Network
+name: Run a recent full Optimism block in the Hardhat Network
 
 on:
   schedule:
@@ -10,8 +10,8 @@ defaults:
     working-directory: packages/hardhat-core
 
 jobs:
-  test-recent-mainnet-block:
-    name: Test recent mainnet block
+  test-recent-optimism-block:
+    name: Test recent Optimism block
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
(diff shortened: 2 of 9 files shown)


.github/workflows/test-recent-optimism-block.yml
```diff
@@ -0,0 +1,50 @@
...
+defaults:
+  run:
+    working-directory: packages/hardhat-core
+
+jobs:
+  test-recent-optimism-block:
+    name: Test recent Optimism block
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
```
galargh, line 17:
```suggestion
      - uses: actions/checkout@v4
```

.github/workflows/test-recent-mainnet-block.yml
```diff
@@ -33,11 +33,9 @@ jobs:
 
       - name: Run test
         uses: actions-rs/cargo@v1
```
galargh, line 35: FYI, https://github.com/actions-rs has been archived last year and the actions repos haven't been updated for quite a while. I think it'd be nice to replace them, but maybe best to do it in a separate PR - I could propose it some time this week if you like the idea.

.github/workflows/test-recent-optimism-block.yml
```diff
@@ -0,0 +1,50 @@
...
+      - name: Install Rust (stable)
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          override: true
+
+      - uses: Swatinem/rust-cache@v2
+
+      - name: Cache EDR RPC cache
+        uses: actions/cache@v2
```
galargh, line 28:
```suggestion
        uses: actions/cache@v4
```

galargh (commented):
I just had a quick look at the GHA part of the PR, looks nice :)

.github/workflows/test-recent-optimism-block.yml
Wodann: Thanks for detecting these!

Do you have an automated way of linting GitHub actions for outdated versions while editing? My VSCode extension didn't highlight these issues

.github/workflows/test-recent-mainnet-block.yml
Wodann: As we use this across all of our GitHub actions, it would be good to do it in a follow-up task.

I created one in our backlog: https://github.com/NomicFoundation/edr/issues/642

Wodann:
Thank you for the suggestions, @galargh! ❤️

.github/workflows/test-recent-mainnet-block.yml
Xanewok: On a somewhat related note, do we need `actions-rs/cargo` (cargo specifically) at all? It seem to be just running `cargo` and we don't seem to need any cross-compilation or other target specs in general (which could be noisy, I guess), so this might be a regular `run` step, instead.

crates/tools/src/remote_block.rs
```diff
@@ -1,9 +1,62 @@
...
+pub enum SupportedChainTypes {
+    L1,
+    Optimism,
+}
+
+pub async fn replay(
+    chain_type: SupportedChainTypes,
+    url: String,
+    block_number: Option<u64>,
+    chain_id: u64,
```
fvictorio, line 23: I know this was already here, but I hadn't noticed it before. Do we need this parameter?

`edr_evm::test_utils::run_full_block` could uses the `chain_id` after it created the `rpc_client`, so it could be just fetched.

`chain_type`, on the other hand, seems like it can't be avoided, or at least it shouldn't.

Wodann: That's a very good question!

In other places we explicitly override the chain ID with a local chain ID - for security reasons. Here, I don't think that's necessary.

Wodann: Addressed in https://github.com/NomicFoundation/edr/pull/638/commits/5805442fae323eb07bfd7aaa0f51044dde199a6e

fvictorio (approved):

PR #640, opened 2024-09-02
[multi-chain]: Drive-by clarify the required traits in `ChainSpec` and decouple them a bit

Xanewok:
When I started implementing the `ChainSpec` trait I found it difficult to navigate all of the required traits, their associated types and how they all relate to each other.

Since the [book chapter](https://github.com/NomicFoundation/edr/blob/feat/multichain/book/src/04_multichain/01_chain_spec.md) reads like the traits should be increasing in scope depending on the impls, I tried to decouple them a bit to make them more additive and to clarify which trait is responsible for what.

Firstly, instead of requiring the `EvmWiring::Transaction` to be convertible to its RPC counterpart directly in the trait bound, I moved the fallible conversion down to the RpcSpec. While strictly less powerful (does not impl TryFrom anymore but also was not used) I feel that it makes it clear that the `RpcSpec` add strictly more logic, and decouples the traits, based on the core EVM (wiring) types above.

Secondly, while also changing the semantics a bit, I extracted and grouped the conversions between the RPC types and the internal/EVM ones. I hope that makes sense as I don't know the code well enough but it made sense to always require these conversions despite which logs (execution/filter) we further require and this also groups the relevant errors together and provides some more symmetry with the core (wiring) Block/Transaction types.

I don't feel so strongly about formatting but not indenting so much makes it a bit easier for me to follow but I'm happy to bring the old back.

### DIFF
--- a/crates/edr_evm/src/chain_spec.rs
+++ b/crates/edr_evm/src/chain_spec.rs
@@ -26,6 +26,7 @@ use crate::{
 #[allow(clippy::trait_duplication_in_bounds)]
 pub trait ChainSpec:
     alloy_rlp::Encodable
+    // Defines the chain's internal types like blocks/headers or transactions
     + EthHeaderConstants
     + EvmWiring<
         Block: BlockEnvConstructor<Self, block::Header> + BlockEnvConstructor<Self, PartialHeader>,
@@ -36,12 +37,11 @@ pub trait ChainSpec:
                          + Eq
                          + ExecutableTransaction
                          + Transaction
-                         + TransactionType
-                         + TryFrom<
-            <Self as RpcSpec>::RpcTransaction,
-            Error = Self::RpcTransactionConversionError,
-        >,
-    > + RpcSpec<
+                         + TransactionType,
+    >
+    // Defines an RPC spec and a conversion between RPC <-> EVM types
+    + RpcSpec<RpcTransaction: TryInto<<Self as revm::primitives::EvmWiring>::Transaction, Error = Self::RpcTransactionConversionError>>
+    + RpcSpec<
         ExecutionReceipt<FilterLog>: Debug,
         RpcBlock<<Self as RpcSpec>::RpcTransaction>: EthRpcBlock
                                                          + TryInto<
@@ -96,7 +96,7 @@ pub trait ChainSpec:
     fn chain_name(chain_id: u64) -> Option<&'static str>;
 }
 
-/// A trait for constructing a block a [`PartialHeader`] into an EVM block.
+/// A trait for constructing a (partial) block header into an EVM block.
 pub trait BlockEnvConstructor<ChainSpecT: ChainSpec, HeaderT> {
     /// Converts the instance into an EVM block.
     fn new_block_env(header: &HeaderT, hardfork: ChainSpecT::Hardfork) -> Self;
--- a/crates/edr_evm/src/chain_spec.rs
+++ b/crates/edr_evm/src/chain_spec.rs
@@ -31,35 +31,29 @@ pub trait ChainSpec:
     + EvmWiring<
         Block: BlockEnvConstructor<Self, block::Header> + BlockEnvConstructor<Self, PartialHeader>,
         Transaction: alloy_rlp::Encodable
-                         + Clone
-                         + Debug
-                         + PartialEq
-                         + Eq
-                         + ExecutableTransaction
-                         + Transaction
-                         + TransactionType,
+          + Clone
+          + Debug
+          + PartialEq
+          + Eq
+          + ExecutableTransaction
+          + Transaction
+          + TransactionType,
     >
-    // Defines an RPC spec and a conversion between RPC <-> EVM types
-    + RpcSpec<RpcTransaction: TryInto<<Self as revm::primitives::EvmWiring>::Transaction, Error = Self::RpcTransactionConversionError>>
+    // Defines an RPC spec and conversion between RPC <-> EVM types
     + RpcSpec<
-        ExecutionReceipt<FilterLog>: Debug,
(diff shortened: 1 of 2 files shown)


crates/edr_evm/src/chain_spec.rs
```diff
@@ -26,40 +26,34 @@ use crate::{
...
-        RpcReceipt: Debug
-                        + RpcTypeFrom<BlockReceipt<Self>, Hardfork = Self::Hardfork>
-                        + TryInto<BlockReceipt<Self>, Error = Self::RpcReceiptConversionError>,
+          + TryInto<EthBlockData<Self>, Error = Self::RpcBlockConversionError>,
         RpcTransaction: EthRpcTransaction
-                            + RpcTypeFrom<TransactionAndBlock<Self>, Hardfork = Self::Hardfork>,
-    > + RpcSpec<
+          + RpcTypeFrom<TransactionAndBlock<Self>, Hardfork = Self::Hardfork>
+          + TryInto<<Self as revm::primitives::EvmWiring>::Transaction, Error = Self::RpcTransactionConversionError>,
+        RpcReceipt: Debug
```
Wodann, line 49: What's the reason for inverting `RpcReceipt` and `RpcTransaction`? With this change it's no longer following the trait's (alphabetic) order

crates/edr_evm/src/chain_spec.rs
```diff
@@ -26,40 +26,34 @@ use crate::{
...
-                         + ExecutableTransaction
-                         + Transaction
-                         + TransactionType
-                         + TryFrom<
-            <Self as RpcSpec>::RpcTransaction,
-            Error = Self::RpcTransactionConversionError,
-        >,
-    > + RpcSpec<
-        ExecutionReceipt<FilterLog>: Debug,
+          + Clone
```
Wodann, line 34: I like the alignment better. I'm curious how you achieved it?

I'm using `cargo +nightly fmt` but it formatted it the way it was before.

Wodann (changes requested):

crates/edr_evm/src/chain_spec.rs
Xanewok: I did not know that we should follow the alphabetic order, I can revert if that's the case, but this reads slightly better for me, as this follows the order from smaller to bigger building blocks (block -> tx -> receipt) and it only adds to the existing order from the `EvmWiring`, so kind of retains the symmetry.

crates/edr_evm/src/chain_spec.rs
Xanewok: I think we should pin the nightly version used for formatting for consistency. I noticed that formatting wasn't working for me when I tried formatting with `cargo +nightly fmt` (was using `rustc 1.82.0-nightly (e57f3090a 2024-08-05)` directly or with the git pre-commit hook script but nothing worked, so I formatted it manually. I now bumped the nightly version to `rustc 1.83.0-nightly (d6c8169c1 2024-09-03)` and still can't force rustfmt to reformat this.

What exactly is your process to format these?

crates/edr_evm/src/chain_spec.rs
fvictorio: In the Hardhat repo we had a de facto policy¹ I like a lot: if a formatting-related thing (or any opinion-prone subject) can't be automatically enforced (with a formatter or a linter, for example), then that thing is not modified nor discussed. If we apply that here, then we have two options:

1) Configure `rustfmt` to sort this kind of thing alphabetically
2) If that's not possible, keep it as it was originally written

Notice this _technically_ means that adding a new element here and not inserting it in alphabetically order would be fair game. There's always nuance though: if something is already sorted, it probably makes sense to keep it that way.

I'm not (yet) saying we should follow this policy in EDR. But, in my experience, it's a sensible thing to do and it ends up saving a lot time and mental energy.

¹ AFAIK this is not documented anywhere.

crates/edr_evm/src/chain_spec.rs
Wodann: I'm using VSCode's "format on save" and have it configured to use the nightly formatter.

crates/edr_evm/src/chain_spec.rs
Wodann: I'm happy to follow that policy. Agost and I agreed to that in the past too. We leaving the heavy lifting to the linters and rustfmt, but otherwise leave the code as it was done by the implementer. Otherwise it will be an endless (subjective) discussion about preferences. That's why I was asking why the ordering is being changed in this case.

As for code I write, I follow alphabetical order in ADTs and follow that order when using traits or initializing instances of ADTs (unless moving a value).

fvictorio: In that case, yes, let's continue to do that. I'll add a topic for the technical sync just in case we want to discuss this a bit further.

Xanewok: In general, I agree with this de facto policy, especially deferring to tools whenever possible to avoid needless discussions.

What I also found helpful in my experience is the argument that code is read more than it's written and that it's more valuable to explicitly make crucial/common concepts (like `ChainSpec` here) easier to grok when there's no tool or automatism to fall back on (or sometimes even going against it if it's makes it particularly hard to reason about the resulting code; not this case, though).

While I don't have a very strong preference here, it's just that I was having a hard time deciphering what is required/defined here, so I tried to clear it up and show some symmetry even if it meant going against the existing order.

I restored the order in https://github.com/NomicFoundation/edr/pull/640/commits/aca0abb6b623856b3a0278c861ce5a2b9d540ab4, let me know if I should change anything here.

Wodann (approved):

Xanewok:
Thanks for the review! I'll merge this, since I'll open a PR soon that bases on this.

PR #644, opened 2024-09-05
test: use correct cheatcode interface in tests

agostbiro:
Previously an outdated VM interface was used in the cheatcode integration tests due to a path issue. This PR fixes that + removes some unused scripting related Solidity test files (scripting is out of scope for EDR).

Note: base branch should be merged first.


crates/foundry/cheatcodes/spec/src/lib.rs
```diff
@@ -162,6 +162,7 @@ interface Vm {{
 
     #[test]
     fn iface_up_to_date() {
+        dbg!(&IFACE_PATH);
```
fvictorio, line 165: Was this left intentionally?

fvictorio (approved):
LGTM, just one comment.

crates/foundry/cheatcodes/spec/src/lib.rs
agostbiro: That's an oversight, thanks!

PR #646, opened 2024-09-05
ci: measure and upload code coverage

Wodann:
This PR adds measuring of code coverage statistics while executing Rust tests and uploads the results to codecov.io


agostbiro (approved):
Thanks, LGTM

PR #649, opened 2024-09-06
ci: add workflow for testing Optimism every night

Wodann:
This workflow needs to be added to main as the schedule trigger only works on the default branch.

Original PR: https://github.com/NomicFoundation/edr/pull/638

### DIFF
--- a/.github/workflows/test-recent-optimism-block.yml
+++ b/.github/workflows/test-recent-optimism-block.yml
@@ -1,4 +1,4 @@
-name: Run a recent full block in the Hardhat Network
+name: Run a recent full Optimism block in the Hardhat Network
 
 on:
   schedule:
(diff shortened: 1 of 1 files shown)


.github/workflows/test-recent-optimism-block.yml
```diff
@@ -0,0 +1,52 @@
+name: Run a recent full Optimism block in the Hardhat Network
+
+on:
+  schedule:
+    - cron: "0 */8 * * *"
+  workflow_dispatch:
+
+defaults:
+  run:
+    working-directory: packages/hardhat-core
```
fvictorio, line 10: I didn't notice this in the previous PR, but is this correct? Seems like a copy-paste leftover.

Wodann: Good catch. Addressed in follow-up commits

fvictorio (approved):

PR #652, opened 2024-09-09
test: add isolate mode JS integration test

agostbiro:
There is a Rust Solidity test runner integration test that tests that the isolate flag works by checking the gas report. I copied the same test i  to the JS integration test to make sure that the enabling isolate mode works from JS as well.

I've also abstracted out the JS Solidity test runner integration test context to be able to be used from multiple test source files.


js/integration-tests/solidity-tests/test/testContext.ts
```diff
@@ -0,0 +1,98 @@
...
+
+export class TestContext {
+  readonly rpcUrl = process.env.ALCHEMY_URL;
+  readonly rpcCachePath: string = "./edr-cache";
+  readonly fuzzFailuresPersistDir: string = "./edr-cache/fuzz";
+  readonly invariantFailuresPersistDir: string = "./edr-cache/invariant";
+  readonly artifacts: Artifact[];
+  readonly testSuiteIds: ArtifactId[];
+
+  constructor(artifacts: Artifact[], testSuiteIds: ArtifactId[]) {
```
fvictorio, line 22: Normally a static async constructor means a private constructor, unless you want to allow creating instances directly too.

fvictorio (commented):
I understand that the test contract is a copy of an existing solidity test, but I think we can do something much simpler, if we only care about testing that isolate mode is enabled:

```solidity
contract IsolateTest is Test {
    StorageLib storageLib;

    function setUp() public {
        storageLib = new StorageLib();
    }

    function testIsolateTest() public {
        // tstore key: 1 with value :2
        storageLib.tstore(1, 2);

        // toload key: 1
        uint256 val = storageLib.tload(1);

        // If the test is run with `--isolate` flag, the value should be 0
        // as --isolate run each top level call as seperate transaction, so tload will return 0
        assertEq(val, 0, "did you forget to use --isolate flag for 'forge test'?");
    }
}

contract StorageLib {
    function tstore(uint256 key, uint256 val) public {
        assembly {
            tstore(key, val)
        }
    }

    function tload(uint256 key) public view returns (uint256 val) {
        assembly {
            val := tload(key)
        }
        return val;
    }
}
```

(copied from https://github.com/pancakeswap/pancake-v4-core/blob/857b4bf8055e74465900b5c15997fa100ee3044a/test/Isolate.t.sol)

I'd rather use this simpler version because it's easier to understand what is testing and why it works.

js/integration-tests/solidity-tests/test/testContext.ts
agostbiro: Thanks fixed in [4f96f26](https://github.com/NomicFoundation/edr/pull/652/commits/4f96f26ee817db63b623f52fa5baf9d0efc4c73c)

js/integration-tests/solidity-tests/contracts/IsolateTest.sol
```diff
@@ -0,0 +1,40 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+pragma solidity ^0.8.24;
+
+import "./test.sol";
+import "./Vm.sol";
+
+contract IsolateModeTest is DSTest {
```
fvictorio, line 7: Let's add a mention here that this was taken from the Pancakeswap repo.

fvictorio (approved):
Minor comment, feel free to merge after fixing that.

PR #658, opened 2024-09-10
test: add JS fuzz/invariant integration tests

agostbiro:
Add JS fuzz/invariant integration tests and fix the following issues:

- Make the failure persist dir optional for fuzz config as intended originally to be able to disable persisting results.
- Disable collecting gas report samples during fuzz/invariant testing for performance as we don't yet support gas reports.
- Remove unwrap on `FuzzConfig::failure_persist_file`. Instead reflect in the type system that it's not optional.
- Mitigate leaking the fuzz failure persist path by making sure that one particular path is only leaked once.


crates/foundry/forge/src/lib.rs
```diff
@@ -38,19 +46,46 @@ impl TestOptions {
...
-        )
+
+        if let Some(failure_persist_dir) = fuzz_config.failure_persist_dir {
+            let failure_persist_path = failure_persist_dir
+                .join(fuzz_config.failure_persist_file)
+                .into_os_string()
+                .into_string()
+                .expect("path should be valid UTF-8");
+
+            // HACK: We need to leak the path as
```
Wodann, line 57: For correctness, I believe we should still drop the leaked strings upon static deinitialization. That can be achieved by adding a local `struct SomeName { inner: HashSet<&'static str> }` which has a `impl Drop`. This destructor would be called when the `static FAILURE_PATHS` is dropped, guaranteeing that our application doesn't report any memory leaks in tooling.

crates/foundry/forge/src/runner.rs
```diff
@@ -795,13 +745,18 @@ impl<'a> ContractRunner<'a> {
...
                                 // Persist error in invariant failure dir.
-                                if let Err(err) = foundry_common::fs::create_dir_all(failure_dir) {
-                                    error!(%err, "Failed to create invariant failure dir");
-                                } else if let Err(err) = foundry_common::fs::write_json_file(
-                                    failure_file.as_path(),
-                                    &call_sequence,
-                                ) {
-                                    error!(%err, "Failed to record call sequence");
+                                if let Some(failure_dir) = failure_dir {
+                                    let failure_file = failure_file.expect("failure file ");
```
Wodann, line 749:
```suggestion
                                    let failure_file = failure_file.expect("failure file");
```

Typo?

Wodann (changes requested):
One concern w.r.t. the leaking of static strings. Otherwise LGTM!

fvictorio:
I pushed a commit removing `chai-as-promised` and checking if a directory exists using `fs.existsSync`.

The previous version wasn't bad or anything, but I've had so many bad experiences with `chai-as-promised` that I'd rather avoid it.

js/integration-tests/solidity-tests/test/fuzz.ts
```diff
@@ -0,0 +1,105 @@
...
+    assert.equal(result2.totalTests, 1);
+
+    // The fuzz failure directory should now be created
+    assert.isTrue(existsSync(failureDir));
+  });
+
+  // One test as steps should be sequential
+  it("FailingInvariant", async function () {
+    const failureDir = testContext.invariantFailuresPersistDir;
+    const defaultConfig = {
```
fvictorio, line 45: I would rename this to `invariantConfig`

js/integration-tests/solidity-tests/test/fuzz.ts
```diff
@@ -0,0 +1,105 @@
...
+      }
+    );
+    assert.equal(result2.failedTests, 1);
+    assert.equal(result2.totalTests, 1);
+    const secondDuration = performance.now() - secondStart;
+
+    // The invariant failure directory should still be there
+    assert.isTrue(existsSync(failureDir));
+    // The second run should be faster than the first because it should use the persisted failure.
+    assert.isBelow(secondDuration, firstDuration * 0.25);
```
fvictorio, line 103: I think there is a more robust way to check this. The `TestResult` objects have a `kind.runs` number for invariant tests which will be 1 when there is a cached seed, and it will almost surely be above 1 when there's not (and this can be done even more unlikely by changing the `% 13` part of the solidity test to a bigger value).

The problem is that `runTestsWithStats` only returns the number of tests and failures. Maybe we can just add a `testResults` property to its returned value, that way there's no need to refactor all the existing usages of that helper.

This is just a suggestion btw, feel free to keep it as is if you think that this approach won't have frequent false positives or negatives.

agostbiro: Thanks, good idea, switched to checking runs. The assertion is that there are more than 1 runs on an uncached run and there is 1 run on a cached one. I didn't want to increase the `13` to a larger value to reduce the chance of the invariant testing not catching the bug.

[ab94d0b](https://github.com/NomicFoundation/edr/pull/658/commits/ab94d0b77756dbc8b1caa6fdadb7db918b2fe2f6)

js/integration-tests/solidity-tests/test/fuzz.ts
agostbiro: Thanks, fixed in [ab94d0b](https://github.com/NomicFoundation/edr/pull/658/commits/ab94d0b77756dbc8b1caa6fdadb7db918b2fe2f6)

js/integration-tests/solidity-tests/contracts/Invariant.t.sol
```diff
@@ -0,0 +1,39 @@
...
+
+    function addToB(uint256 amount) external {
+        val2 += amount;
+        val3 += amount;
+    }
+}
+
+// Test that the invariant testing works correctly by catching a bug in the contract.
+contract FailingInvariantTest is DSTest {
+    StochasticWrongContract val;
```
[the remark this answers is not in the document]
agostbiro: Thanks, fixed in [ab94d0b](https://github.com/NomicFoundation/edr/pull/658/commits/ab94d0b77756dbc8b1caa6fdadb7db918b2fe2f6)

js/integration-tests/solidity-tests/contracts/Invariant.t.sol
```diff
@@ -0,0 +1,39 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.24;
+
+import "./test.sol";
+import "./Vm.sol";
+
+contract StochasticWrongContract {
+    uint256 public val1;
+    uint256 public val2;
+    uint256 public val3;
```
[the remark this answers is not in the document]
agostbiro: Thanks, fixed in [ab94d0b](https://github.com/NomicFoundation/edr/pull/658/commits/ab94d0b77756dbc8b1caa6fdadb7db918b2fe2f6). I named `sum` `both` instead as both method calls increment it.

crates/foundry/forge/src/runner.rs
agostbiro: Thanks, missed some text there after the space. Fixed in [ab94d0b](https://github.com/NomicFoundation/edr/pull/658/commits/ab94d0b77756dbc8b1caa6fdadb7db918b2fe2f6)

crates/foundry/forge/src/lib.rs
agostbiro: Hey, I tried this out with LLVM leak sanitizer and it reports errors in the base branch, but not in this one, so I think it's ok as is.

Test command (tried on ARM linux in Docker): `RUSTFLAGS="-Z sanitizer=leak" cargo +nightly test -p forge test_fuzz`

agostbiro: (I couldn't get Miri working on the `forge` integration tests)

Wodann: Thanks for validating with leak sanitizer.

[This](https://stackoverflow.com/a/71419608/4944080) is the reason why Leak Sanitizer would not be able to detect a "memory leak" of this kind. It depends on your definition of memory leak, though. Leak sanitizer doesn't consider it a leak because a static variable is still referring to the address of the originally allocated strings.

Wodann (approved):

PR #659, opened 2024-09-10
feat: add Optimism N-API package

Wodann:
This PR restructures the `edr_napi_core` and `edr_napi` to be able to share code with the new `edr_napi_optimism` package.

I ran into one problem that I was able to solve. Namely, when a crate that generates NAPI bindings is added as a dependency to another crate that generates NAPI bindings, the test target of the dependant crate will throw linker errors. This is a [known issue](https://github.com/napi-rs/napi-rs/issues/2130). I was only able to work around it by adding the recommended rustflags to both the workspace- and project-level `.cargo/config.toml`. If I only added it to the project, the `cargo test --all-features --all-targets --workspace` command would still throw the linker errors.

There is currently still a bug in the `edr_napi_optimism` JS tests. In the background, the `hyper` Client is sporadically trying to execute a future using its executor. By default, this executor uses `tokio::spawn` but since the `edr_optimism` NPM package doesn't have a tokio runtime, this panics.

> thread '<unnamed>' panicked at /home/vscode/.cargo/registry/src/index.crates.io-6f17d22bba15001f/hyper-0.14.28/src/common/exec.rs:49:21:
> there is no reactor running, must be called from the context of a Tokio 1.x runtime
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

The solution is to specify a custom executor, but [reqwest always uses the Tokio executor](https://github.com/seanmonstar/reqwest/blob/09884ed0a09d43ebd5c67491caf4ad5683fba995/src/async_impl/client.rs#L684C73-L684C86). An [issue](https://github.com/seanmonstar/reqwest/issues/557) exists to add this executor, but a previous PR was rejected because it was deemed unnecessary.

For now, my proposal would be to merge the changes to the project layout and move forward with the release of the `GenericChainSpec`, as `edr_optimism` won't be exposed anyway.

A future issue can then:
1. fork `reqwest`
2. add setting of the executor
3. set the executor with the `tokio::Runtime` of our `edr` NPM package's runtime


crates/edr_napi/Cargo.toml
```diff
@@ -44,11 +41,8 @@ openssl-sys = { version = "0.9.93", features = ["vendored"] }
 napi-build = "2.0.1"
 
 [features]
-tracing = ["edr_evm/tracing", "edr_napi_core/tracing", "edr_provider/tracing"]
 scenarios = ["rand"]
-
-[profile.release]
```
Wodann, line 50: `cargo` was reporting that this setting was ignored. Only the workspace `Cargo.toml`'s profile are respected.

crates/edr_napi/Cargo.toml
```diff
@@ -4,17 +4,15 @@ version = "0.3.5"
 edition = "2021"
 
 [lib]
-crate-type = ["cdylib"]
+crate-type = ["cdylib", "lib"]
```
Wodann, line 7: This is required for `edr_napi_optimism` to use this crate as a dependency

crates/edr_napi_core/src/logger.rs
```diff
@@ -0,0 +1,1193 @@
+use core::fmt::Debug;
```
Wodann, line 1: File moved from `edr_napi`, with a modification to the config. This removes the dependency on `napi` types.

.github/workflows/edr-ci.yml
```diff
@@ -10,7 +10,6 @@ on:
   workflow_dispatch:
 
 env:
-  RUSTFLAGS: -Dwarnings
```
Wodann, line 13: This should not have any adverse side-effects, as we also manually pass this argument to `cargo clippy` which should fail if there are any warnings.

For reference:

`cargo clippy --workspace --all-targets --all-features -- -D warnings`

PR #661, opened 2024-09-11
feat: disable fuzz fixtures in Solidity

agostbiro:
Foundry supports specifying fuzz inputs in Solidity. They call this feature “fixtures”. This seems to follow from the “do everything in Solidity” philosophy and would probably make more sense to get as an input in EDR. The values could be then specified in a config in Hardhat or generated dynamically with JS.  

For this reason, we don't want to let users specify fuzz inputs with Solidity, but fixtures are integral to the fuzz executor implementation, so while it’s possible to rip it out, it’d be a big effort to put it back later on. And we will want to add support for fuzz fixtures later on, just not via Solidity code.

So I added a `solidity_fuzz_fixtures` config option which is always false for the JS runner and it turns off loading fixtures from Solidity, but the Rust integration test runner can turn it on to keep the tests covering this feature.

This way users won’t be able to depend on Foundry-style fuzz fixtures by accident, but we can add a fuzz fixtures config option to the JS runner interface in the future with little effort.


crates/foundry/forge/src/runner.rs
```diff
@@ -278,51 +284,62 @@ impl<'a> ContractRunner<'a> {
...
-            .filter(|func| func.is_fixture())
-            .for_each(|func| {
-                if func.inputs.is_empty() {
-                    // Read fixtures declared as functions.
+            .filter(|func| func.is_fixture());
+
+        // No-op if the feature is disabled
+        if !self.solidity_fuzz_fixtures {
+            fixture_funcs.for_each(|func| {
+                log::warn!("Possible fuzz fixture usage detected: '{}', but solidity fuzz fixtures are disabled.", &func.name);
```
agostbiro, line 296: This is a noop unless a consumer is configured which isn't currently

Wodann (approved):

fvictorio (approved):

PR #663, opened 2024-09-13
fix: Allow block time offset to be negative

### LINKED ISSUE #588: Allow forking blocks with future timestamps
If EDR forks a block whose timestamp is in "the future", then it panics with an error like this:

```
thread 'tokio-runtime-worker' panicked at /build/crates/edr_provider/src/data.rs:2564:18:
current time must be after fork block: SystemTimeError(3340.313588137s)
```

The easiest way to reproduce this manually is to just set the system clock back one hour and then fork mainnet.

I don't see any good reason to disallow this and, as far as I can tell, pre-EDR Hardhat didn't have this behavior.

See https://github.com/zemse/hardhat-tracer/issues/74#issuecomment-2271140832 for more context.

Xanewok:
Closes #588

This uses a simple approach of first converting to `i128`, which we know will be lossless, to only then fallibly convert it to a smaller type of `i64`.

We do that for simplicity: one could convert a tuple of (sign, value) to shave some bytes and cycles but we would have to explicitly handle the case of `i64::MIN` (its absolute value is greater by 1 than i64::MAX), which would be a bit noisy, whereas not handling it and naively using a `if sign { -1 } else { 1 } * i64::try_from(value)...` would not fully handle that edge case.

I tested locally the reproduction steps from #588 and can both confirm that the error happens using the steps and the problem disappears after applying the fix here. Let me know if you feel that we should add a regression test and if so, how would you prefer to proceed with testing this.

### DIFF
--- a/.changeset/hungry-cats-enjoy.md
+++ b/.changeset/hungry-cats-enjoy.md
@@ -1,0 +1,5 @@
+---
+"@nomicfoundation/edr": patch
+---
+
+Allow forked block time offset to be negative
--- a/crates/edr_provider/src/data.rs
+++ b/crates/edr_provider/src/data.rs
@@ -2692,8 +2692,8 @@ fn create_blockchain_and_state(
                 );
 
             let elapsed = match timer.since(fork_block_timestamp) {
-                Ok(elapsed) => i128::from(elapsed),
-                Err(forward_drift) => -i128::from(forward_drift.duration().as_secs()),
+                Ok(elapsed) => -i128::from(elapsed),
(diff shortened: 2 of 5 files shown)


Wodann:
Yes, for bugs we always add a regression test, ideally on the EDR side, using typed method invocations. The integration test folder of the provider already has a bunch of examples.

Xanewok:
Pushed a fix and added a regression test for this; verifying that the test fails without the fix and succeeds afterwards, so this should be ready for review now.

Wodann (approved):
Thanks for adding the test. LGTM

PR #664, opened 2024-09-13
feat: add Optimism bindings in N-API

Wodann:
This PR exposes Optimism through the N-API bindings, toggled using the `optimism` feature flag. It's a modified version of #659.

It also incorporates a fix for linker failure on Linux when building the test target and creates a clear split between internal (Rust) and external (N-API) types. The internal types live in in `edr_napi_core`.

I [noticed](https://github.com/NomicFoundation/edr/pull/664/files#r1759415675) that we need to be aware of mixed usage of `pnpm build` and `pnpm build:optimism`, as it will likely result in the Optimism bindings being committed by accident. This mistake will be caught by CI.


crates/edr_napi_core/src/logger.rs
```diff
@@ -0,0 +1,1193 @@
...
+    SyncBlock,
+};
+use edr_provider::{
+    time::CurrentTime, CallResult, DebugMineBlockResult, EstimateGasFailure, ProviderError,
+    ProviderSpec, TransactionFailure,
+};
+use itertools::izip;
+
+/// Trait for a function that decodes console log inputs.
+pub trait DecodeConsoleLogInputsFn: Fn(Vec<Bytes>) -> Vec<String> + Send + Sync {}
```
Wodann, line 21: We now deal with Rust `Fn`s instead of the `ThreadSafeFunction` and `JsFunction` in this file. This creates a clear demarcation between marshalled types and internal types.

crates/edr_napi/Cargo.toml
```diff
@@ -45,11 +43,9 @@ openssl-sys = { version = "0.9.93", features = ["vendored"] }
 napi-build = "2.0.1"
 
 [features]
-tracing = ["edr_evm/tracing", "edr_napi_core/tracing", "edr_provider/tracing"]
+optimism = ["dep:edr_optimism"]
 scenarios = ["rand"]
-
-[profile.release]
```
Wodann, line 51: `cargo` was reporting that this setting was ignored. Only the workspace `Cargo.toml`'s profile are respected.

.github/workflows/edr-ci.yml
```diff
@@ -10,7 +10,6 @@ on:
   workflow_dispatch:
 
 env:
-  RUSTFLAGS: -Dwarnings
```
Wodann, line 13: This should not have any adverse side-effects, as we also manually pass this argument to `cargo clippy` which should fail if there are any warnings.

For reference:

`cargo clippy --workspace --all-targets --all-features -- -D warnings`

agostbiro: Thanks for the explanatory comment! I'm wondering if removing it has effect on the `check-edr` and `test-edr-rs`? E.g. if there are some unused imports with some feature combination, would the CI not fail now? It's not a merge blocker for me if it doesn't, just want to make sure we consider it.

.cargo/config.toml
```diff
@@ -3,3 +3,7 @@ gen-execution-api       = "run --bin tools -- gen-execution-api"
 scenario                = "run --release --bin tools -- scenario"
 scenario-with-tracing   = "run --release --features tracing --bin tools -- scenario"
 replay-block            = "run --release --bin tools -- replay-block"
+
+# To be able to run unit tests on Linux, support compilation to 'x86_64-unknown-linux-gnu'.
+[target.'cfg(target_os = "linux")']
+rustflags = ["-C", "link-args=-Wl,--warn-unresolved-symbols"]
```
agostbiro, line 9: Do we have any unit tests that need this now? If not, I'd rather not add it tbh, because if I understand correctly, this will turn linker errors into warnings for all targets which means we might have build errors and not notice it.

agostbiro (commented):
Thanks for the PR! I left some questions + there seems to be a lot of CI errors

crates/edr_napi/.cargo/config.toml
```diff
@@ -6,3 +6,7 @@ target-dir = "./target"
 [target.aarch64-unknown-linux-musl]
 linker = "aarch64-linux-musl-gcc"
 rustflags = ["-C", "target-feature=-crt-static"]
+
+# To be able to run unit tests on Linux, support compilation to 'x86_64-unknown-linux-gnu'.
+[target.'cfg(target_os = "linux")']
+rustflags = ["-C", "link-args=-Wl,--warn-unresolved-symbols"]
```
agostbiro, line 12: Same comment: https://github.com/NomicFoundation/edr/pull/664/files#r1759191016

.github/workflows/edr-ci.yml
Wodann: Good catch! Given that we use `cargo hack` in `check-edr`, we would lose some functionality - I think. I've re-added the environment variables for that specific command [here](https://github.com/NomicFoundation/edr/pull/664/commits/65a545eb8047400d473b883db7c68d2d524bf043).

Clippy catches all warnings that `cargo check|build|test` would catch, as it is a superset of the check command - under the hood. Since we run `cargo clippy` with `--all-targets`, the test target is also run.

This means that it's possible for the `test-edr-rs` step to succeed with warnings, but the step running `cargo clippy` would fail the CI.

To validate this, I just added two simple functions (dead code) to `lib.rs`:

```rs
fn foo() {
    println!("foo");
}

#[cfg(test)]
fn bar() {
    prinltn!("bar");
}
```

Running `cargo check --all-targets` gives the following warnings:

```
warning: function `foo` is never used
  --> crates/edr_eth/src/lib.rs:62:4
   |
62 | fn foo() {
   |    ^^^
   |
   = note: `#[warn(dead_code)]` on by default

warning: `edr_eth` (lib) generated 1 warning
warning: function `bar` is never used
  --> crates/edr_eth/src/lib.rs:67:4
   |
67 | fn bar() {
   |    ^^^

warning: `edr_eth` (lib test) generated 2 warnings (1 duplicate)
```

These become errors when running `cargo clippy --all-targets -- -D warnings` (failing one at a time):

```
error: function `foo` is never used
  --> crates/edr_eth/src/lib.rs:62:4
   |
62 | fn foo() {
   |    ^^^
   |
   = note: `-D dead-code` implied by `-D warnings`
   = help: to override `-D warnings` add `#[allow(dead_code)]`

error: could not compile `edr_eth` (lib) due to 1 previous error

error: function `bar` is never used
  --> crates/edr_eth/src/lib.rs:63:4
   |
63 | fn bar() {
   |    ^^^
   |
   = note: `-D dead-code` implied by `-D warnings`
   = help: to override `-D warnings` add `#[allow(dead_code)]`
```

.cargo/config.toml
Wodann: FYI, it sounds like `napi_build` always does this under the hood because it's linked dynamically.

https://github.com/napi-rs/napi-rs/issues/1005#issuecomment-1011034770
https://github.com/napi-rs/napi-rs/issues/1782#issuecomment-1793980863

I agree that we should avoid setting this for the workspace. I have removed both instances for now, as there are no tests using `edr_napi`

crates/edr_napi/test/optimism.ts
```diff
@@ -0,0 +1,159 @@
...
+import chaiAsPromised from "chai-as-promised";
+
+import {
+  ContractAndFunctionName,
+  EdrContext,
+  L1_CHAIN_TYPE,
+  l1ProviderFactory,
+  MineOrdering,
+  SubscriptionEvent,
+  // HACK: There is no way to exclude tsc type checking for a file from the
```
Wodann, line 11: There is no way to exclude files from `tsc` via the command-line (https://github.com/microsoft/TypeScript/issues/46005)

agostbiro (approved):
Thanks for looking into my comments, LGTM!

PR #673, opened 2024-09-20
fix: Handle nullptr JS TypedArrays by bumping napi to 2.6.11

Xanewok:
This fixes an instance where we roundtrip a zero-length `Uint8Array` in JS code and violate the `slice::from_raw_parts` contract (checked in debug builds at run-time since Rust 1.78) because internally napi calls it with a null pointer, see https://github.com/napi-rs/napi-rs/pull/2258.

Found while running the test suite of https://github.com/ProjectOpenSea/seaport with local, most recent Hardhat+EDR.

This should unblock further testing and releasing a Hardhat version that uses the EDR-powered stack tracing.

### DIFF
--- a/crates/edr_napi/Cargo.toml
+++ b/crates/edr_napi/Cargo.toml
@@ -18,8 +18,8 @@ k256 = { version = "0.13.1", default-features = false, features = ["arithmetic",
 log = { version = "0.4.20", default-features = false }
 # when napi is pinned, be sure to pin napi-derive to the same version
 # The `async` feature ensures that a tokio runtime is available
-napi = { version = "2.16.0", default-features = false, features = ["async", "error_anyhow", "napi8", "serde-json"] }
-napi-derive = "2.16.0"
+napi = { version = "2.16.11", default-features = false, features = ["async", "error_anyhow", "napi8", "serde-json"] }
+napi-derive = "2.16.11"
 edr_defaults = { version = "0.3.5", path = "../edr_defaults" }
 edr_evm = { version = "0.3.5", path = "../edr_evm", features = ["tracing"]}
 edr_eth = { version = "0.3.5", path = "../edr_eth" }
--- a/crates/edr_napi/Cargo.toml
+++ b/crates/edr_napi/Cargo.toml
@@ -18,6 +18,7 @@ k256 = { version = "0.13.1", default-features = false, features = ["arithmetic",
 log = { version = "0.4.20", default-features = false }
 # The `async` feature ensures that a tokio runtime is available
 napi = { version = "2.16.11", default-features = false, features = ["async", "error_anyhow", "napi8", "serde-json"] }
+napi-derive = "2.16.11"
 edr_defaults = { version = "0.3.5", path = "../edr_defaults" }
 edr_evm = { version = "0.3.5", path = "../edr_evm", features = ["tracing"]}
 edr_eth = { version = "0.3.5", path = "../edr_eth" }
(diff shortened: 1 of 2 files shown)


crates/edr_napi/Cargo.toml
```diff
@@ -18,8 +18,8 @@ k256 = { version = "0.13.1", default-features = false, features = ["arithmetic",
 log = { version = "0.4.20", default-features = false }
 # when napi is pinned, be sure to pin napi-derive to the same version
```
Xanewok, line 19: These don't seem to be released in tandem (anymore?). Should we remove the comment or reword it to something like "If you bump the napi version, please also make sure that we use a compatible napi-derive"?.

This should ideally be handled by the upstream package but sometimes mistakes happen and I assume one of these were the reason for this warning, here.

agostbiro (commented):
The code LGTM, but there seems to be a big regression on the `seaport` benchmark: https://github.com/NomicFoundation/edr/actions/runs/10959402320/job/30431668868?pr=673

I reran it just in case it's a false positive.

crates/edr_napi/Cargo.toml
Wodann: Feel free to. I also removed the comment in `feat/multichain`

Xanewok:
> The code LGTM, but there seems to be a big regression on the `seaport` benchmark: https://github.com/NomicFoundation/edr/actions/runs/10959402320/job/30431668868?pr=673
> 
> I reran it just in case it's a false positive.

Good catch, thanks! I will look into this to bisect what introduced a regression. Is that the only test suite that regressed (so much)?

agostbiro:
> Good catch, thanks! I will look into this to bisect what introduced a regression. Is that the only test suite that regressed (so much)?

On the second run `rocketpool` showed a regression as well, but not that much: https://github.com/NomicFoundation/edr/actions/runs/10959402320/job/30441788798

There is an iPython Notebook with a [data analysis](https://github.com/NomicFoundation/edr/blob/main/crates/tools/python/scenarios.ipynb) of the scenarios in the repo that might be helpful. What stands out at first glance is that `seaport` is the only scenario that makes heavy use of `eth_sendRawTransaction`:

Screenshot 2024-09-23 at 11 04 07

Xanewok:
@agostbiro I'm looking at seaport specifically and I'm struggling with locally reproducing the >10% regression.

I ran the benchmarks locally on a Pop_OS 22.04 running the Linux 6.9.3 kernel on an AMD Ryzen 3900X CPU (with "performance" CPU governor set), the only thing running was the X session and the GNOME terminal running the benchmark (no browser, no IMs, no A/V, no network), so the results should be relatively noise-free.

Here's the benchmarking setup that I used: https://gist.github.com/Xanewok/4ff505ef1ecfde910210fd083dbd2dc7. I also modified the iteration count up to 20 from 7, here are the results:

```
napi version 2.16.1: 6795.493899 ms
napi version 2.16.2: 6703.319267 ms                                                                                            
napi version 2.16.4: 6782.881617 ms                          
napi version 2.16.5: 6813.238553 ms                         
napi version 2.16.6: 6875.120218999999 ms                                                                                      
napi version 2.16.7: 6722.393457 ms                           
napi version 2.16.8: 6776.054829 ms       
napi version 2.16.9: 6825.334383 ms                                                                                            
napi version 2.16.10: 6831.796009 ms
napi version 2.16.11: 6868.889158 ms    
```

What is the self-hosted environment that we test on? Do you have an idea what I might do to help reproduce the regression locally?

agostbiro:
Thanks for looking into this @Xanewok and sorry for the false alarm! It turns out we need to bump iterations on the `seaport` scenario. These are the results of running `pnpm run -s benchmark --grep seaport` five times on the benchmark server (ms):

```
main

6966.649748
7101.934028
7365.330143
7404.122042999999
7330.396748

fix/nullptr-js-typed-arrays

7651.774936000001
7756.7813479999995
6889.5786769999995
6949.626776
8270.088566999999
```

agostbiro (approved):

PR #674, opened 2024-09-20
refactor: adapt `isContext` cheatcode

agostbiro:
Foundry has an `isContext` [cheatcode](https://github.com/foundry-rs/foundry/pull/7377) that tells you which `forge` command was used to execute the Solidity code (e.g. `forge test` or `forge coverage` or `forge script`). 

I adapted the `isContext` cheatcode for EDR by renaming the context enum from `ForgeContext` to `ExecutionContext` and removed the scripting related variants.

The Foundry code assumes that Solidity code is ran via the `forge` CLI and that the execution context is set based on the subcommand once per process. This doesn't make sense for us, since EDR provides Solidity tests as a library, so I changed this to getting the execution context as a config value in the cheatcode implementation.

I changed the representation of the execution context in the Rust code. I split the previous `ForgeContext` enum into two enums as it was previously used as both as an argument from Solidity where it can have five variants and as a config value where it can only have four variants (`TestGroup` doesn't make sense as a config value).

I also removed the `PartialEq` implementation for the previous `ForgeContext` and inlined its logic in the cheatcode as the `PartialEq` implementation wasn't symmetric (`Test == TestGroup`, but `TestGroup != Test`) which can be a source of subtle bugs.

### DIFF
--- a/crates/edr_napi/src/solidity_tests/config.rs
+++ b/crates/edr_napi/src/solidity_tests/config.rs
@@ -2,8 +2,10 @@ use std::{collections::HashMap, fmt::Debug, path::PathBuf};
 
 use alloy_primitives::hex;
 use edr_solidity_tests::{
-    executors::invariant::InvariantConfig, fuzz::FuzzConfig,
-    inspectors::cheatcodes::CheatsConfigOptions, SolidityTestRunnerConfig,
+    executors::invariant::InvariantConfig,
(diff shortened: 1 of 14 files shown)


Wodann (approved):
The logic changes look sound to me. I can't really comment on the product design choices outlined in the PR description though.

fvictorio (approved):

PR #675, opened 2024-09-23
chore: format with latest nightly version

fvictorio:


Xanewok (approved):

Xanewok:
We should probably pin this version somewhere to reduce the CI failure. For example we do that [in slang](https://github.com/NomicFoundation/slang/blob/15c437c6c4902cd43e0027f750ba59e8f22f47f9/bin/hermit.hcl#L9) (but using Hermit would be an overkill).

PR #677, opened 2024-09-24
bump: REVM to latest version with minimal forking

Wodann:
Upgrade REVM to the [latest version that requires minimal changes](https://github.com/NomicFoundation/revm/commit/4722477). At this point we only need a modification to revert to Rust v1.80.

A large restructuring of crates was performed in REVM that required changes to our re-exports. I tried to structure it in a way that would minimise future impact of reorganisations in REVM.


agostbiro (approved):
Thanks, LGTM

PR #678, opened 2024-09-24
ci: Replace actions-rs actions in GitHub workflows

### LINKED ISSUE #642: Replace `actions-rs` actions in GitHub workflows
**Definition of Done**
Replace all occurrences  of the deprecated `actions-rs` GitHub actions with a maintained version

Xanewok:
Closes #642

This gets rid of the redundant `actions-rs/cargo` and moves to the maintained action https://github.com/actions-rust-lang/setup-rust-toolchain instead of `actions-rs/toolchain`.


.github/workflows/test-recent-optimism-block.yml
```diff
@@ -14,13 +14,7 @@ jobs:
...
           ref: feat/multichain
 
-      - name: Install Rust (stable)
-        uses: actions-rs/toolchain@v1
-        with:
-          profile: minimal
-          override: true
-
-      - uses: Swatinem/rust-cache@v2
+      - uses: ./.github/actions/setup-rust
```
Xanewok, line 17: `main` was already merged to `feat/multichain`, so this should be safe to do, i.e. the action should already be there.

.github/workflows/edr-ci.yml
```diff
@@ -84,43 +71,24 @@ jobs:
...
-
-      - name: Install stable toolchain
-        uses: actions-rs/toolchain@v1
-        with:
-          profile: minimal
-          override: true
-          components: clippy
-
-      - uses: Swatinem/rust-cache@v2
+        run: rustup toolchain install nightly --profile minimal --component rustfmt
```
Wodann, line 88: Why not use inputs to the `setup-rust` action?

https://github.com/NomicFoundation/edr/blob/feat/multichain/.github/actions/setup-node/action.yml#L4

Xanewok: I found it to be a lot simpler - we just execute `setup-rust` as our baseline "setup Rust toolchain as expected" and this is the only exception to this. Since it's a one-liner and used only here, I found it simpler to just call this directly and keep the main action lean.

Wodann: You can have default values for all inputs of an action, which would keep the default action lean too, while allowing configuration.

I'll leave it up to you to decide whether to take that approach or not.

Wodann (approved):

.github/workflows/edr-ci.yml
Xanewok: I tried to find a middle ground in https://github.com/NomicFoundation/edr/pull/678/commits/e7796f73696342cf6ef8fc91d73bf832ac1ff4df, hope that's okay with you.

Wodann: Thanks! That's exactly what I had in mind :)

galargh (commented):
Sorry for only getting to it now. I really like how you used `cargo` directly ❤️ I only have one comment about potentially pinning action versions.

PR #680, opened 2024-09-24
fix: Only add unmapped instructions at the opcode boundary

Xanewok:
This fixes the adding unmapped instruction logic started from a misaligned opcode, i.e. in the middle of a PUSHx opcode. The bug was present in the original code but it did not verify the opcode for correctness, which is why we're only seeing it now.

@fvictorio this fixes the Chainlink's issue that's been reported lately.

### DIFF
--- a/.changeset/dirty-phones-vanish.md
+++ b/.changeset/dirty-phones-vanish.md
@@ -1,0 +1,5 @@
+---
+"@nomicfoundation/edr": patch
+---
+
+fix(tracing): Decode unmapped instructions only at the opcode boundary
--- a/crates/edr_solidity/src/source_map.rs
+++ b/crates/edr_solidity/src/source_map.rs
@@ -209,3 +209,35 @@ pub fn decode_instructions(
 
     instructions
 }
+
+#[cfg(test)]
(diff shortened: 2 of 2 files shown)


Wodann (approved):
Having a reproducible test case for this would be ideal, but if I understood correctly that's hard to achieve in this case, right?

Xanewok:
> Having a reproducible test case for this would be ideal, but if I understood correctly that's hard to achieve in this case, right?

Yes, the test case relies on a specific behaviour, where the last instruction in the mapping is a PUSH: the exact test where I'm reproducing it had a 55 MB build info JSON; I have stripped it to 1.8 MB, which was only the `CompilerOutputContract` for the offending contract.

 I could craft a synthetic unit test that exercises the new logic, for example calling the procedure with a last instruction being a PUSH but I'm not sure how much value would there be in that, as to introduce a regression one would have to modify the now documented code and there's not a lot of different ways to do that, anyway.

fvictorio (approved):

Xanewok:
I've added a simple unit test that will prevent from us regressing in this case.

PR #683, opened 2024-09-25
chore: use @ignored npm namespace

fvictorio:
This PR mainly replaces `@nomicfoundation/edr` with `@ignored/edr` wherever it makes sense. Besides that:

- It uses the `IGNORED_ORG_NPM_TOKEN` secret for the release workflow.
- It uses optional dependencies to publish, like we used to do. This simplifies things locally, at least until we actually make a release in npm. Notice that the comment that explains this mentions `@ignored`; my hope here is that we'll remember to do this when we grep `@ignored` to remove all occurrences in the project.

### DIFF
--- a/crates/edr_napi/scripts/prepublish.sh
+++ b/crates/edr_napi/scripts/prepublish.sh
@@ -5,5 +5,6 @@ set -o pipefail
 
 pnpm napi prepublish -t npm --skip-gh-release
 
+# temporarily commented to be able to publish in the @ignored org locally with verdaccio
 # rename optionalDependencies key to dependencies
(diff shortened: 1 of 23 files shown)


agostbiro (approved):
Thanks, LGTM pending successful CI.

crates/edr_napi/scripts/prepublish.sh
```diff
@@ -5,5 +5,6 @@ set -o pipefail
 
 pnpm napi prepublish -t npm --skip-gh-release
 
+# temporarily commented to be able to publish in the @ignored org locally with verdaccio
```
Wodann, line 8: It is still commented

PR #686, opened 2024-09-26
ci: solidity test benchmarks

agostbiro:
This PR adds automated regression checks for the JS Solidity test runner. The setup is similar to how the RPC benchmark that benchmarks Hardhat Node-style workloads works.

## Benchmark Suite

The benchmark runs the test harness of `forge-std`. This has two advantages:
- Largest coverage of cheatcodes in a single test harness which is helpful to catch regressions in all features.
- Makes sure our implementation works with `forge-std` which we intend to use.
 
The disadvantage is that `forge-std` is not representative of how users are writing Solidity tests. But this is not a concern as long as we’re not using the benchmark to guide optimizations, only to catch regressions.

There are multiple benchmark scenarios. The first one is called `Total` which measures the entire test harness execution of the `forge-std` test harness. The purpose of this scenario is to make sure that there are no regressions in the parallel test suite executor.

In addition to test suites, individual tests within test suites are also parallelized. This means that if we want to catch regressions in features covered by a certain test suite, we have to execute that test suite in isolation. For this reason, in addition to the `Total` scenario, we execute the following test suites as well:

- `StdCheatsTest`: [environment](https://book.getfoundry.sh/cheatcodes/environment)-related cheatcode coverage + fuzzing.
- `StdStorageTest`: storage-related cheatcode coverage + fuzzing.
- `StdMathTest`: math-related utilities + it's very quick (~15ms on my machine), so regressions from FFI overhead should show up on this.
- `StdUtilsForkTest`: read-type programmatic fork tests.
- `StdCheatsForkTest`: write + fuzz tests for programmatic forking.

The rest of the test suites in `forge-std` test things implemented in Solidity or test file-system utilities, so they're less interesting for us.

## Benchmark Metric

The metric for all scenarios is the wall clock time measured from JS in milliseconds. The measurement excludes loading files from disk as that's out of scope for EDR. If there is a 10% regression on any of the scenarios above, the benchmark job fails (same as the RPC benchmark).

## Baseline 

The baseline is executing the same tests with `forge test`. Instructions to run the baseline are in the [crates/tools/js/benchmark/src/solidity-tests.ts](https://github.com/NomicFoundation/edr/blob/e5eb6201ff6fede9373e08fc87c7e90ffb77878b/crates/tools/js/benchmark/src/solidity-tests.ts) file. 

I tested manually that performance is identical to `forge test` on my machine, so we don't have to keep running the baseline.

## Benchmark Jobs

There are two benchmark workflows in the `feat/solidity-tests` branch and three benchmark jobs:

- `EDR Benchmarks`:
  - `Run JS scenario runner benchmark for Hardhat Node style workload`
    - Stores results for `main` and compares PRs against against those results
    - Dashboard: https://nomic-foundation-automation.github.io/edr-benchmark-results/bench/
  - `Run JS Solidity test runner benchmark`
    - This is the new job
    - Stores results for `feat/solidity-tests` and compares PRs targeting `feat/solidity-tests` against those results
      - Once `feat/solidity-tests` is merged to `main`, we'll switch to storing results for `main`
    - Dashboard will be live here once this PR is merged: https://nomic-foundation-automation.github.io/edr-benchmark-results/soltests/

- `EDR Benchmarks for feat/solidity-tests`
  - `Run JS scenario runner benchmark for Hardhat Node style workload`
    - Stores results for `feat/solidity-tests` and compares PRs against against those results
    - Once `feat/solidity-tests` is merged to `main`, we'll remove this

I reorganized the workflow files a bit to share Rust compilation cache + start caching the RPC cache for the Solidity test runner benchmarks.

### DIFF
--- a/.github/workflows/edr-benchmark-feat-solidity-tests.yml
+++ b/.github/workflows/edr-benchmark-feat-solidity-tests.yml
@@ -29,13 +29,13 @@ jobs:
       - uses: actions/checkout@v3
       - uses: ./.github/actions/setup-node
       - uses: ./.github/actions/setup-rust
-        
+
       - name: Install packages
         run: pnpm install --frozen-lockfile --prefer-offline
 
       - name: Run benchmark tests
         run: pnpm test
-  
+
   js-scenario-benchmark:
     name: Run JS scenario runner benchmark for Hardhat Node style workload
     environment: github-action-benchmark
@@ -61,7 +61,7 @@ jobs:
 
       - name: Install packages
         run: pnpm install --frozen-lockfile --prefer-offline
-        
+
       - name: Run benchmark
         run: pnpm run benchmark
 
--- a/.github/workflows/edr-benchmark-feat-solidity-tests.yml
+++ b/.github/workflows/edr-benchmark-feat-solidity-tests.yml
@@ -47,7 +47,7 @@ jobs:
       - uses: ./.github/actions/setup-rust
 
       - name: Rust cache
-        uses: actions/cache@v3
+        uses: actions/cache@v4
         with:
           path: |
             ~/.cargo/registry/index/
@@ -55,7 +55,9 @@ jobs:
             ~/.cargo/git/db/
             .cargo-cache
             target/
-          key: benchmark-rs-cache-v1
+          save-always: true
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
+          restore-keys: ${{ runner.os }}-cargo-
 
       - name: Install packages
         run: pnpm install --frozen-lockfile --prefer-offline
--- a/.github/workflows/edr-benchmark-feat-solidity-tests.yml
+++ b/.github/workflows/edr-benchmark-feat-solidity-tests.yml
@@ -24,7 +24,7 @@ jobs:
     environment: github-action-benchmark
     runs-on: self-hosted
     # Only run for trusted collaborators since third-parties could run malicious code on the self-hosted benchmark runner.
-    if: github.ref == 'refs/heads/main' || github.repository == github.event.pull_request.head.repo.full_name
+    if: github.ref == 'refs/heads/feat/solidity-tests' || github.repository == github.event.pull_request.head.repo.full_name
     steps:
       - uses: actions/checkout@v3
       - uses: ./.github/actions/setup-node
--- a/.github/workflows/edr-benchmark-feat-solidity-tests.yml
+++ b/.github/workflows/edr-benchmark-feat-solidity-tests.yml
@@ -1,4 +1,4 @@
-name: EDR Benchmark for feat/solidity-tests
+name: EDR Benchmarks for feat/solidity-tests
 
 on:
   push:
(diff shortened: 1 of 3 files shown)


crates/tools/js/benchmark/src/solidity-tests.ts
```diff
@@ -39,65 +74,202 @@ export async function setupForgeStdRepo() {
...
+    rpcEndpoints: {
+      // These are hardcoded in the `forge-std` foundry.toml
+      mainnet:
+        "https://eth-mainnet.alchemyapi.io/v2/WV407BEiBmjNJfKo9Uo_55u0z0ITyCOX",
+      optimism_sepolia: "https://sepolia.optimism.io/",
+      arbitrum_one_sepolia: "https://sepolia-rollup.arbitrum.io/rpc/",
+    },
     fuzz: {
-      failurePersistDir: path.join(forgeStdRepoPath, "failures"),
+      seed: "0x1234567890123456789012345678901234567890",
```
Wodann, line 180: Maybe add a comment here to indicate that we are using a seed to ensure deterministic execution.

.github/workflows/edr-benchmark-feat-solidity-tests.yml
```diff
@@ -19,50 +19,63 @@ concurrency:
...
 jobs:
-  js-benchmark:
-    name: Run JS scenario runner benchmark
+  benchmarks-test:
+    name: Benchmarks test
     environment: github-action-benchmark
     runs-on: self-hosted
     # Only run for trusted collaborators since third-parties could run malicious code on the self-hosted benchmark runner.
-    if: github.ref == 'refs/heads/feat/solidity-tests' || (github.event.pull_request.author_association == 'OWNER' || github.event.pull_request.author_association == 'MEMBER' || github.event.pull_request.author_association == 'COLLABORATOR')
+    if: github.ref == 'refs/heads/main' || github.repository == github.event.pull_request.head.repo.full_name
```
Wodann: `main` looks like it might be changed by mistake

Wodann (approved):
Overall looks good. There is one erroneous change that you might want to look into before merging. The other comment is a suggestion

crates/tools/js/benchmark/src/solidity-tests.ts
agostbiro: Thanks, fixed in [08dacd2](https://github.com/NomicFoundation/edr/pull/686/commits/08dacd20152819aa9b765ffca7af96865aff498f)

agostbiro:
There is a failing benchmark indicating a performance regression compared to `main`: https://github.com/NomicFoundation/edr/actions/runs/11121119547/job/30899561763?pr=686

This is not actually a regression, what happened was that EDR got faster in `main` and those changes are not yet in this branch as merging `main` into `feat/solidity-tests` is blocked by independent JS package resolution issues.

.github/workflows/edr-benchmark-feat-solidity-tests.yml
agostbiro: Apologies, I got confused a bit. It actually needs to be `feat/solidity-tests` here to make it run in the branch, but we want to change it back to `main` before we merge `feat/solidity-tests` into main. Fixed it here with a TODO comment: [Fix precondition for soltests benchmark](https://github.com/NomicFoundation/edr/pull/524/commits/ae8625fa3457eb5d4bd1f402959881bc91c2015f)

PR #687, opened 2024-09-27
chore: update hardhat tests

fvictorio:
Sync `hardhat-tests` with the latest changes in the Hardhat repository. This also sets up things to change solc's CI to use our repo instead of Hardhat's.


agostbiro (approved):
Thanks, LGTM

PR #688, opened 2024-09-30
fix: Make sure to uniformly 0x-prefix hex strings in custom error message

Xanewok:
Fixes https://github.com/NomicFoundation/hardhat/issues/5793

The address case slipped through the cracks - I adapted the stack trace test and added a unit test for that.

cc @fvictorio


agostbiro (approved):
Thanks, LGTM

PR #692, opened 2024-10-02
refactor: use expect instead of unwrap

Wodann:
Replaces `unwrap`s with `expect`s to make it easier to debug Solidity Stack trace code.


agostbiro (approved):
Thanks, LGTM

PR #693, opened 2024-10-02
fix: ignore unknown opcode in stack traces

agostbiro:
Fixes https://github.com/NomicFoundation/hardhat/issues/5784


alcuadrado (approved):
LGTM

Issue #700, opened 2024-10-21
Process hangs with `Invalid hex bytecode for contract` error


agostbiro:
Feature: Solidity tests

Process hangs with `Invalid hex bytecode for contract` error

What’s also curious about this error is that it is not advertised on the error callback. It is just displayed and the process hangs.

Reproduction steps:

1. Check out the `wip/testing-solidity-tests-wtih-hardhat` branch of `NomicFoundation/edr`.
2. Check out the `wip/solidity-test-merged` branch of `NomicFoundation/hardhat`
3. Go to `v-next/hardhat` of the Hardhat checkout
4. Build Hardhat with `pnpm build`
5. Go to `crates/edr_solidity_tests/tests/testdata` of the EDR checkout
6. Link Hardhat with `npm link $PATH_TO_HARDHAT/v-next/hardhat`
7. Replace the value of https://github.com/NomicFoundation/edr/blob/wip/testing-solidity-tests-wtih-hardhat/crates/edr_solidity_tests/tests/testdata/hardhat.config.js#L4 with `["default/fork"]` or `["default/linking"]`.
8. Run solidity tests with `npx hardhat3 test:solidity`

agostbiro:
Done in https://github.com/NomicFoundation/edr/pull/704

PR #704, opened 2024-10-21
fix: report config error on invalid artifact

agostbiro:
Make sure the config error callback is called with an error if an invalid artifact is provided. Partially fixes https://github.com/NomicFoundation/edr/issues/700


Wodann (approved):

galargh (approved):
🥇

Issue #702, opened 2024-10-21
Test suite results are missing for empty test artifacts


agostbiro:
We rely on counting how many suite results we received back on the progress callback to find out whether the test run has finished or not. If we pass a test artifact that doesn’t contain any executable tests to the `runSolidityTests` function in the `testSuiteIds` argument, then we, currently, wont’ receive a callback for that artifact.

**Definition of Done**

Issue suite results on the progress callback for all the test suite IDs that EDR receives. In particular, a suite result for a test artifact that doesn’t contain any executable tests would have 0 test results.

agostbiro:
Done in https://github.com/NomicFoundation/edr/pull/705

PR #705, opened 2024-10-21
fix: return empty result for empty test suite

agostbiro:
Make sure we return an empty test suite result for empty test suites. 

The fix changes the semantics of the `TestFilter` slightly, hence the need for the change to `crates/edr_solidity_tests/tests/it/fork.rs`, but since the `TestFilter` isn't exposed (it's only used in integration tests), I think this is ok. 

The change in semantics is that previously we only returned results for a test suite if the test name filter also matched at least one test name in the test suite. Now we return an empty result instead if the path filter and suite name filter match, but the test name filter doesn't match.

Fixes https://github.com/NomicFoundation/edr/issues/702

### DIFF
--- a/crates/edr_solidity_tests/tests/it/fork.rs
+++ b/crates/edr_solidity_tests/tests/it/fork.rs
@@ -11,7 +11,7 @@ async fn test_cheats_fork_revert() {
     let filter = SolidityTestFilter::new(
         "testNonExistingContractRevert",
         ".*",
-        &format!(".*cheats{RE_PATH_SEPARATOR}Fork"),
+        &format!(".*cheats{RE_PATH_SEPARATOR}Fork2"),
     );
     let runner = TEST_DATA_DEFAULT.runner().await;
     let suite_result = runner.test_collect(filter).await;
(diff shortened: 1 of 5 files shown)


crates/edr_solidity_tests/tests/it/fork.rs
```diff
@@ -11,7 +11,7 @@ async fn test_cheats_fork_revert() {
     let filter = SolidityTestFilter::new(
         "testNonExistingContractRevert",
         ".*",
-        &format!(".*cheats{RE_PATH_SEPARATOR}Fork"),
+        &format!(".*cheats{RE_PATH_SEPARATOR}Fork2"),
```
Wodann, line 14: Is this a unique identifier that needs to be upgraded any time the filter changes?

Wodann (approved):

crates/edr_solidity_tests/tests/it/fork.rs
agostbiro: The intention here is to match a single test suite, but there are actually two test suites with the `cheats/Fork` prefix: `cheats/Fork.t.sol` and `cheats/Fork2.t.sol`. Here we want to only match the latter, and this was achieved previously by also filtering the test methods. But with the change in semantics in this PR, if the path/suite name matches, but the test name filter doesn't match, we will return an empty suite result. So in order to only get one test suite result here, I had to make the path filter more specific to only match `Fork2.t.sol`.

Wodann: Gotcha! Thanks for explaining 🙂

galargh (approved):
Great, thanks - this is going to be really helpful on the hardhat side 🙇

PR #706, opened 2024-10-22
fix: remove console.log backwards compatibility tests

Wodann:
This PR fixes broken test files due to the upgrade to the latest Hardhat. These files were already removed in Hardhat in this PR: https://github.com/NomicFoundation/hardhat/pull/5764


agostbiro (approved):
Thanks, LGTM!

Wodann:
I ended up pushing the merge commit separately as the "squash and merge" option has previously caused negative side-effects by erasing the merge commit. As a result, git becomes confused about what the base commit of this branch is.

PR #708, opened 2024-10-23
chore: upgrade soltest crates to Revm 9

agostbiro:
Bump the forked Foundry crates for the Solidity tests feature to use REVM 9. 

Based on https://github.com/foundry-rs/foundry/pull/7934. There was one additional change required: specifying the Cancun hardfork for the Rust integration tests, as they used the default version and in Revm 9 that is Prague which caused problems. (The JS interface specifies Cancun explicitly.)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -69,7 +69,7 @@ dependencies = [
 [[package]]
 name = "alloy-consensus"
 version = "0.1.0"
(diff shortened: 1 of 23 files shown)


agostbiro:
I'm going to merge this without approval to avoid having to juggle too many branches. The CI is passing except for a regression on a Solidity test benchmark that runs in fork mode. I couldn't repro this locally, so I strongly suspect this is just a matter of network conditions.

PR #710, opened 2024-10-24
test: add test to validate HTTP headers

Wodann:
Adds a test to validate the fix in https://github.com/NomicFoundation/hardhat/pull/5846. In the process, I also bumped Hardhat to the latest version, which includes the aforementioned fix.

Bumping Hardhat's version also required the removal of some stack trace tests, which had previously already been removed in Hardhat (https://github.com/NomicFoundation/hardhat/pull/5764)


agostbiro (approved):
Thanks, LGTM

PR #723, opened 2024-10-30
chore: fix setting tracing subscriber

agostbiro:
Previously we were only setting a default tracing subscriber for the current thread in `edr_napi` which precluded logs from other threads being collected.

With this change, the subscriber is set globally for all threads so that we can get tracing output when running JS tests, e.g. from `hardhat-tests`: `RUST_LOG=debug pnpm test`.

### DIFF
--- a/crates/edr_napi/src/context.rs
+++ b/crates/edr_napi/src/context.rs
@@ -1,5 +1,7 @@
 use std::{ops::Deref, sync::Arc};
 
+#[cfg(feature = "tracing")]
+use napi::Status;
 use napi_derive::napi;
 use tracing_subscriber::{prelude::*, EnvFilter, Registry};
 
--- a/crates/edr_napi/src/context.rs
+++ b/crates/edr_napi/src/context.rs
@@ -1,6 +1,5 @@
 use std::{ops::Deref, sync::Arc};
 
-use napi::Status;
 use napi_derive::napi;
 use tracing_subscriber::{prelude::*, EnvFilter, Registry};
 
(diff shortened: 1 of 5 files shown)


Wodann (commented):
The [documentation](https://docs.rs/tracing/latest/tracing/subscriber/fn.set_global_default.html) mentions:

> Note: Libraries should NOT call set_global_default()! That will cause conflicts when executables try to set them later.

It seems like we might want to use [`with_default`](https://docs.rs/tracing/latest/tracing/subscriber/fn.with_default.html) inside each function to prevent causing crashes if other shared libraries are also using `tracing`.

We have a similar pending task for fixing the tokio runtime. That's currently blocked by the `reqwest_middleware` not allowing a custom tokio runtime to be set.

WDYT?

agostbiro:
(Btw I had a quick look and our version of `tracing` doesn't seem to set `no_mangle`, so maybe other shared Rust lib calling it is not an issue?)

Wodann:
> (Btw I had a quick look and our version of `tracing` doesn't seem to set `no_mangle`, so maybe other NodeJS shared Rust lib calling it is not an issue?)

If mangle is used in the same way as it normally is (i.e. mangle a function name to be unique), I'm not sure how it would prevent the issue. The static memory location of the default logger would still be reused, right?

Wodann:
The safest bet might be to log a warning when we can't set a default logger? Then we can at least get debug logs to work for now, without causing crashes.

We can find a permanent solution in a follow-up task.

agostbiro:
> If mangle is used in the same way as it normally is (i.e. mangle a function name to be unique), I'm not sure how it would prevent the issue. The static memory location of the default logger would still be reused, right?

Right, mangling is only relevant if the other shared lib was a C lib, but another Rust shared lib will have the same mangled name and can use that to look up the address of the static that holds the global logger? I'm unsure of the details of how this works exactly, but I'm curious to learn more.

agostbiro:
> The safest bet might be to log a warning when we can't set a default logger? Then we can at least get debug logs to work for now, without causing crashes.

Ok, done in [0f405d2](https://github.com/NomicFoundation/edr/pull/723/commits/0f405d2eed668e98f28dbaf325024f1783463f9a)

Wodann (approved):

PR #724, opened 2024-10-31
refactor: generate non-const enum in d.ts

### LINKED ISSUE #721: Remove `enum` types from N-API
Hardhat is using [isolatedModules](https://www.typescriptlang.org/tsconfig/#isolatedModules) to ensure compatibility with single-file transpilation tools like tsx. With isolatedModules, const enums can’t be inlined, since this setting assumes each file compiles independently rather than together.

Instead of using enums, it would be better to generate constants, similar to what was done for [SpecId](https://github.com/NomicFoundation/edr/blob/feat/multichain/crates/edr_napi/src/chains/l1.rs#L126).

[original discussion](https://github.com/NomicFoundation/hardhat/pull/5840#issuecomment-2437953620)

**Definition of done**
Remove all instances of const enum types from the N-API in the [`feat/multichain` branch](https://github.com/NomicFoundation/edr/tree/feat/multichain)

Wodann:
Resolves #721


schaable (approved):
Looks good! I've tested it by replacing the casts in `v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/stack-traces/stack-trace-solidity-errors.ts` and we no longer get the isolated modules error.

PR #725, opened 2024-11-03
fix: Optimism Sepolia hardfork

Wodann:
This fixes an incorrect chain ID for the OP Sepolia testnet.

It also adds a test. To support this test, I needed to perform a small refactor to expose the `ProviderTestFixture` and make it chain-specific.


agostbiro (approved):
Thanks, LGTM!

PR #726, opened 2024-11-03
fix: chain ID at block number

Wodann:
This fixes a merge error resulting in latest chain ID always being used for `eth_call` instead of using the chain ID at the block number.


agostbiro (approved):
Thanks, LGTM with two small notes:

- There is a merge conflict that needs to be fixed before merging
- Codecov failed the CI, but I think it's fixed already in main?

PR #730, opened 2024-11-18
Update macos-12 to macos-13 (latest macos to still support intel).

nebasuke:
GitHub runner deprecated macos-12 (see https://github.com/actions/runner-images/issues/10721). macos-13 is the latest image to still support x86.

Issue #254, opened 2023-12-13
Test Hardhat with EDR in third-party repositories


Wodann:
**Depends on**
- [x] #273  

**Definition of Done**
For quality assurance, we need to test Hardhat with EDR in the following third-party repositories:

- [x] Test in [OpenZeppelin](https://github.com/OpenZeppelin/openzeppelin-contracts)
- [x] Test in https://github.com/Uniswap/v3-core
- [x] Test in https://github.com/Synthetixio/synthetix (one error in EDR mode which looks like a bug in the test suite)
- [x] Test in https://github.com/NexusMutual/smart-contracts (can't get it running due to various plugin errors)
- [x] Test in https://github.com/ProjectOpenSea/seaport/ (1 test failure due to accessing HH VM private properties)
- [x] Test in https://github.com/smartcontractkit/chainlink/tree/develop/contracts (`hardhat-waffle` errors, but these errors also occur with the latest released HH version, so we ignore them)
- [x] Test in https://github.com/rocket-pool/rocketpool/
- [x] Test in https://github.com/neptune-mutual-blue/protocol (same 29 failures in EJS and EDR mode)
- [x] Test in https://github.com/safe-global/safe-contracts
- [x] Test in https://github.com/taikoxyz/taiko-mono (switched to Foundry)
- [x] Test parallel mode in a sample project
- [x] Decide whether to test more based on https://www.notion.so/nomicfoundation/Testing-with-real-projects-bd896ef0b11c45fbb41eeb6df16e6584

This can be combined with [these performance testing tests](https://github.com/NomicFoundation/edr/issues/33)

agostbiro:
Failures discovered so far:

https://github.com/NomicFoundation/edr/issues/281
https://github.com/NomicFoundation/edr/issues/282
https://github.com/NomicFoundation/edr/issues/283

https://github.com/NomicFoundation/edr/issues/287
https://github.com/NomicFoundation/edr/issues/288

agostbiro:
Rocketpool@6a9dbfd8 is 4x faster in EDR mode than with EthereumJS and all tests pass.

Tested with EDR version https://github.com/NomicFoundation/hardhat/commits/30001589a

agostbiro:
Safe Contracts@914d0f8 is 1.5x faster in EDR mode than with EthereumJS and all tests pass.

Tested with EDR version https://github.com/NomicFoundation/hardhat/commits/954c6a631

agostbiro:
[...]

PR #741, opened 2024-12-10
feat: add chain-specific Block, LocalBlock, and BlockReceipt types

### LINKED ISSUE #735: Optimisim RPC receipts have null in their l1-related fields, and some are missing
**Definition of done**
Optimism RPC receipts return Optimism-specific fields

Wodann:
Resolves #735 

The primary change in this PR is the addition of chain-specific `Block`, `LocalBlock`, and `BlockReceipt` types; along with supporting types, such as a `BlockBuilder` and `BlockReceiptFactory`.

In addition, for parts of the codebase that I touched (or required it), I made the following changes:
- use the newly added types to simplify existing function parameter and struct member types
- remove `ChainSpecT` generics and instead replace them with more granular generics
- add helper/convenience type aliases for chain-specific versions of the changed types

**Follow-up**
- https://github.com/NomicFoundation/edr/issues/742


crates/edr_utils/src/types.rs
```diff
@@ -0,0 +1,4 @@
+// Trait for retrieving a sub-type from a higher-kind type.
+pub trait HigherKinded<ParamT> {
+    type Type;
+}
```
agostbiro, line 4: Hey thanks for the PR this, it's a big achievement! 

I had multiple passes now and I think I'm starting to get it, but tbh this trait still confuses me, so bear with me if I'm off the mark here:

My understanding is that the trait is used to map execution log types to execution receipt types (execution log types are associated types in execution receipt types). Is this correct? If yes, I'd maybe change the name to `InverseAssociatedType` or something like that?

A higher kinded type for me would be something like a functor-like generic map trait that we implement for execution receipt types instead of using the custom `MapReceiptLogs` trait.

Wodann:
The benchmark CI job was failing. I've been investigating it and learned the following (skipping debugging steps for brevity):

1. Previously, both in HH+EthereumJS and HH+EDR, [two test were failing](https://github.com/NomicFoundation/edr/issues/254#issuecomment-1946457716). They failed for different reasons, but we accepted it since there was no regression.
   - The failing test is a test that checks whether a transaction reverts with an ABI encoded error reading `InvalidSigner()`.
2. Scenarios were recorded with these failing tests. The `snapshot.json` expects `305` failures.
3. In my branch, the snapshot reports 306 failures; 1 too many.
   - The extra failure is from the test mentioned in step 1. and is a result of transaction revert with the expected ABI encoded error `InvalidSigner()`
4. I tested the latest `seaport` commit with my version of HH+EDR and all tests are now passing.
   - I ran `yarn test:ref` instead of `yarn test` (which I think was previously used to create the scenario), as `yarn test` is failing with the following error:
   ```
   Error: Command failed: /usr/local/share/nvm/versions/node/v22.11.0/bin/node /workspaces/hardhat/packages/hardhat-core/internal/solidity/compiler/solcjs-runner.js /home/vscode/.cache/hardhat-nodejs/compilers-v2/wasm/soljson-v0.8.24+commit.e11b9ed9.js
   RuntimeError: memory access out of bounds
       at wasm://wasm/053b6712:wasm-function[330]:0x2a475
       at wasm://wasm/053b6712:wasm-function[42946]:0x135e4e6
       at ccall (/home/vscode/.cache/hardhat-nodejs/compilers-v2/wasm/soljson-v0.8.24+commit.e11b9ed9.js:110:28936)
       at /home/vscode/.cache/hardhat-nodejs/compilers-v2/wasm/soljson-v0.8.24+commit.e11b9ed9.js:110:29327
       at runWithCallbacks (/workspaces/hardhat/node_modules/.pnpm/solc@0.8.26_debug@4.3.4/node_modules/solc/bindings/compile.js:169:18)
       at Object.boundFunctionStandard [as compileStandard] (/workspaces/hardhat/node_modules/.pnpm/solc@0.8.26_debug@4.3.4/node_modules/solc/bindings/compile.js:83:20)
       at compileStandardWrapper (/workspaces/hardhat/node_modules/.pnpm/solc@0.8.26_debug@4.3.4/node_modules/solc/wrapper.js:66:24)
       at main (/workspaces/hardhat/packages/hardhat-core/internal/solidity/compiler/solcjs-runner.js:26:23)
       at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
   ```
   [This issue](https://github.com/ethereum/solidity/issues/13966) might be related, but the suggested fix - of upgrading the Solidity version - didn't work for me.
   This might affect benchmark scenario times

crates/edr_utils/src/types.rs
Wodann: I had to look this up myself as I wasn't 100% sure, but my understanding is "Higher kinded types are types which take other types and construct a new type".

In this case, we are indeed constructing the "inverse associated type", but the `trait HigherKinded` can be used to generate any type based on a type parameter.

The `ExecutionReceiptHigherKindedForChainSpec` is an implementation that constructs a chain-specific execution receipt based on the log type.

Wodann:
With the fix for the benchmarks, this PR now only has issues with the Mac CI. I'll merge `main` into `feat/multichain` in a follow-up to fix this, as it has already been [fixed](https://github.com/NomicFoundation/edr/pull/730) in that branch.

agostbiro (approved):
Thanks, LGTM

crates/edr_utils/src/types.rs
Wodann: We discussed offline and came up with the [following changes](https://github.com/NomicFoundation/edr/pull/741/commits/dc4bc44c380dbf568f098299b88a45094556637b) to address the concern.

Issue #740, opened 2024-12-09
Move `cache_dir` config option to `ForkConfig`


Wodann:
**Definition of Done**
- The `cache_dir` member field of `ProviderConfig` should be moved to `ForkConfig`
- Do this in the `feat/multichain` and `feat/solidity-tests` branches

PR #750, opened 2024-12-15
Move cache_dir config option to ForkConfig - feat/solidity-tests

iosh:
-  Rename `ForkConfig` json_rpc_url to url
-  Move  cache_dir to `ForkConfig` 

Relate #740


Wodann (changes requested):
Similar to the other PR, the `cache_dir` should be moved to the `ForkConfig` for `edr_provider` too

PR #752, opened 2024-12-16
Use UInt8Array instead of a Buffer for coinbase in ProviderConfig

### LINKED ISSUE #746: Update ProviderConfig to Use Uint8Array Instead of Buffer for coinbase
This is not a high-priority change, but in the ideal scenario, the `coinbase` field of the `ProviderConfig` interface would accept a `Uint8Array` instead of a `Buffer`.

The motivation is to move away from Node.js-specific types, such as `Buffer`, to facilitate supporting other runtimes in the future. Currently, Hardhat is casting from `Uint8Array` to Buffer when passing the `coinbase` to EDR.

Work should be done in the `feat/multichain` branch.

IftequarIffu:
This PR fixes #746

### DIFF
--- a/crates/edr_napi/index.d.ts
+++ b/crates/edr_napi/index.d.ts
@@ -199,7 +199,7 @@ export interface ProviderConfig {
   /** The configuration for chains */
   chains: Array<ChainConfig>
   /** The address of the coinbase */
-  coinbase: Buffer
+  coinbase: Uint8Array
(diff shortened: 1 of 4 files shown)


Wodann (changes requested):
There are no changes on the Rust side to change the data size which would invalidate code generation of the `index.d.ts` and potentially introduce UB.

The investigation is meant to investigate a refactir on the Rust side

PR #753, opened 2024-12-18
perf: speed up integration test compilation

Wodann:
Every file in the `tests` directory is compiled as a separate binary, which is slow. By using a single entry point, we can speed up compilation.

Inspired by [this blogpost](https://matklad.github.io/2021/02/27/delete-cargo-integration-tests.html).


Wodann:
I merged this to expedite my development, but it would still be helpful to have a look at, @agostbiro

agostbiro:
Thanks, LGTM

PR #755, opened 2024-12-19
refactor: use genesis state instead of genesis accounts

Wodann:
This PR adds support for setting of a full genesis state instead of just genesis accounts. This functionality is now also exposed to Hardhat where it will be used to set the genesis state of a `LocalBlockchain` or to overwrite accounts in the `ForkedBlockchain`.

In the process of updating the `pnpm patch`, I also:
- upgraded to Hardhat 2.22.17
- adopted const constructors for beacon root related constants

**Follow-up work**
- #756


Wodann:
Merging to expedite inclusion in the `release/optimism` branch. I'll handle any requests for change in follow-up PRs 🙂

PR #757, opened 2024-12-20
build: upgrade to Rust 1.83.0

Wodann:
This version of Rust [resolves a bug](https://github.com/rust-lang/rust/pull/131233) that was introduced in 1.81 that broke our usage of native Node modules, inhibiting prior upgrades.


agostbiro:
I reran the benchmark job to see if it was a one-off. The regression on `synthetix` (the largest benchmark suite) is somewhat concerning.

Wodann:
> I reran the benchmark job to see if it was a one-off. The regression on `synthetix` (the largest benchmark suite) is somewhat concerning.

The regression originally happened [here](https://github.com/NomicFoundation/edr/actions/runs/12421733030/job/34704709927?pr=755) in the previous PR.

I still need to investigate why introducing the genesis state introduces the regression, but I expect it either has to do with:
- Changing the `ProviderConfig` in that PR changes the executed code
- The back and forth between Rust and TS introduced additional overhead in construction of the Provider

Wodann:
I added a [follow-up issue](https://github.com/NomicFoundation/edr/issues/761) for the performance regression

PR #764, opened 2025-01-06
fix: ignore unknown opcodes in source maps

### LINKED ISSUE #763: Hardhat Network tracing disabled: VmTraceDecoder failed to be initialized.
### Version of Hardhat

2.22.17

### What happened?

With the latest version of hardhat `2.22.17` and hardhat-tracer `3.1.0` a compiler can generate an opcode that the EDR doesn't recognise. Consequently the tracer is disabled for network requests. 

### Minimal reproduction steps

- Checkout [a repo with minimal reproduction steps](https://github.com/sparrowDom/hardhat-tracer-mre). 
- run the compile & node commands as in the linked repo's Readme
- to make the error go away remove a small function as described in the linked repo's Readme

### Search terms

tracer hardhat network debugging

fvictorio:
Closes #763.

As far as I can tell, this can only happens when solc produces source maps that have an error, so I think just ignoring this is fine. I'm not adding a test for this because it seems to only happen in very special circumstances, with older versions of solc.

### DIFF
--- a/crates/edr_solidity/src/source_map.rs
+++ b/crates/edr_solidity/src/source_map.rs
@@ -159,7 +159,10 @@ pub fn decode_instructions(
         let source_map = &source_maps[instructions.len()];
 
         let pc = bytes_index;
-        let opcode = OpCode::new(bytecode[pc]).expect("Invalid opcode");
+        let opcode = match OpCode::new(bytecode[pc]) {
+            Some(opcode) => opcode,
+            None => continue,
+        };
 
         let push_data = if opcode.is_push() {
             let push_data = &bytecode[bytes_index..][..1 + opcode.info().immediate_size() as usize];


crates/edr_solidity/src/source_map.rs
```diff
@@ -159,7 +159,10 @@ pub fn decode_instructions(
         let source_map = &source_maps[instructions.len()];
 
         let pc = bytes_index;
-        let opcode = OpCode::new(bytecode[pc]).expect("Invalid opcode");
+        let opcode = match OpCode::new(bytecode[pc]) {
```
Wodann, line 162: I don't think it's advisable to fail silently. It opens a whole can of worms when this should actually be panicking.

Both approaches have downsides:
1. Failing silently requires a lot of effort on the user side, which is frustrating
2. Panicking on an invalid opcode is too harsh/restrictive

I'd propose:

Minimal change:
1. Log that we detected an invalid opcode

Ideally, also:
2. Add a boolean in the configuration that sets whether to `ignore_invalid_opcodes`. If that's enabled, we don't panic, but log an error.
3. Add documentation to Hardhat describing this option

Wodann (changes requested):

crates/edr_solidity/src/source_map.rs
nebasuke: Thanks @Wodann, I agree that silent failing could be dangerous. 

Do we have any ability to do debug level logging (to be used when ignore_invalid_opcodes is enabled)? 
Regardless, can we at least at the invalid op code to part of the panic message?

Wodann: We have debug logging using the [`log` crate](https://docs.rs/log/latest/log/). For that you need to run with `RUST_LOG=warning your-command` in the CLI. I don't think we currently enable the logger for Hardhat, though.

We also use a [logger](https://github.com/NomicFoundation/edr/blob/main/crates/edr_napi/src/logger.rs) to print things to Hardhat's logger.

fvictorio: @Wodann I think you might have the wrong impression of what's going on here. I should've explained better the context of the change in the description, sorry about that.

> this should actually be panicking

I'm pretty sure it shouldn't. This code is part of the error inference heuristics, which "fail silently" all the time. From a user perspective, this means that you might get a plain "transaction reverted" instead of, say, "tried to send funds to non-payable function", or a "\<unknown function>" label somewhere in a stack trace. Again, that's much better that a panic.

> Failing silently requires a lot of effort on the user side, which is frustrating

I'm not sure what you have in mind here. Maybe you think that an execution will have the wrong behavior, like an opcode being skipped instead of reverting? But, again, that's not the case here. Users would have less useful information about a revert, but the execution behavior (the fact that a tx reverted, the receipt, etc.) will be exactly the same.

---

I do agree that it's desirable to know that this is happening. I'm not sure that logging a warning is necessarily the best option though; sometimes (like here!) there's not much we can do, and that only leads to non-actionable user reports. I know because it's what happened in Hardhat for a long time, when those kinds of warnings reached diminishing returns after we fixed the actionable stuff.

In any case, I think this a telemetry question, and I wouldn't block this PR on that.

Wodann: Discussed this offline. We'll add a `log::debug` for future reference.

crates/edr_solidity/src/source_map.rs
```diff
@@ -159,7 +160,16 @@ pub fn decode_instructions(
         let source_map = &source_maps[instructions.len()];
 
         let pc = bytes_index;
-        let opcode = OpCode::new(bytecode[pc]).expect("Invalid opcode");
+        let opcode = if let Some(opcode) = OpCode::new(bytecode[pc]) {
+            opcode
+        } else {
+            debug!("Invalid opcode {} at pc: {}", bytecode[pc], pc);
```
agostbiro, line 166: nit: I prefer to not to import these macros from log and just write `log::debug!` so that it's easy to differentiate from `tracing::debug!`

agostbiro (approved):
Thanks LGTM with one nit

crates/edr_solidity/src/source_map.rs
Wodann: +1

Wodann (approved):
Same comment as Agost. Otherwise looks good to me!

PR #770, opened 2025-01-15
test: switch to `forge-std` for JS integration tests

agostbiro:
Switch to `forge-std` for Solidity test JS integration tests.

### DIFF
--- a/js/integration-tests/solidity-tests/contracts/ContractEnvironment.t.sol
+++ b/js/integration-tests/solidity-tests/contracts/ContractEnvironment.t.sol
@@ -1,13 +1,10 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.24;
 
-import "./test.sol";
-import "./Vm.sol";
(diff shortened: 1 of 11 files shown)


js/integration-tests/solidity-tests/contracts/ContractEnvironment.t.sol
```diff
@@ -20,8 +17,4 @@ contract ContractEnvironmentTest is DSTest {
         assertEq(block.number, 23, "block number is incorrect");
         assertEq(block.timestamp, 45, "timestamp is incorrect");
     }
-    
-    function testContextIsTest() public {
-        assertEq(vm.isContext(Vm.ExecutionContext.Test), true);
-    }
 }
```
agostbiro, line 20: I removed this test case, because the execution context doesn't seem to be exposed from `forge-std` and it's unnecessary anyway: there is an identical Rust integration test.

Wodann (approved):

fvictorio (approved):

PR #772, opened 2025-01-17
fix: issue 771

### LINKED ISSUE #771: Multichain EDR crashes if interval mining is non-zero
Reproduction is easy: add `interval: 1n` to the mining configuration here:

https://github.com/NomicFoundation/edr/blob/3415702e7b828ce65c50734d5fcb9f0c0b99f75d/crates/edr_napi/test/optimism.ts#L55-L60

cc @schaable

Wodann:
Fixes #771

### DIFF
--- a/.changeset/rich-parents-shave.md
+++ b/.changeset/rich-parents-shave.md
@@ -1,0 +1,5 @@
+---
+"@ignored/edr-optimism": patch
+---
+
+Fixed exception on termination of provider with interval mining enabled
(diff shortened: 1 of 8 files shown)


agostbiro (approved):
Thanks, the change itself looks good to me, but there are some CI failures that seem to be related to a JS package mismatch. I'm not sure how they could be caused by this PR, so if unrelated feel, free to merge, otherwise fix them please.

PR #774, opened 2025-01-17
chore: remove hardhat-solidity-tests

agostbiro:
As discussed, remove `hardhat-solidity-tests` since it's outdated.


fvictorio:
My only question is why `crates/tools/js/benchmark/buildinfo.json`. But feel free to merge if that was done intentionally.

fvictorio (approved):

PR #775, opened 2025-01-24
fix: remove one way branching in bytecode trie

agostbiro:
This PR fixes stack overflow on drop in the `BytecodeTrie` by eliminating one-way branching which produces very deep tries for large contracts. Care has been taken to preserve the implementation details of the previous implementation which the stack trace heuristics rely on.

I moved the `BytecodeTrie` to a separate module in [7bcb871](https://github.com/NomicFoundation/edr/pull/775/commits/7bcb8712fb70428a6184292a13100daccce06e51). The changes to eliminate one-way branching can be reviewed by reviewing [76840d8](https://github.com/NomicFoundation/edr/pull/775/commits/76840d88c77e6688f616f430f6b50ced1e85bd66).

I also added a [benchmark](https://github.com/NomicFoundation/edr/pull/775/files#diff-f3f8c0e043dcf722c2eb7c98f6ce432856efd6464495c6f54051985243bcdb34) for constructing the `ContractDecoder`, but maybe it'd be better to put this in the `tools` crate, as running the benchmark requires some manual setup.

Eliminating one way branching resulted in a 5x speed up for constructing the `ContractDecoder` for the `forge-std` test suite. 

We are benchmarking the `ContractDecoder` construction instead of the `BytecodeTrie` construction as the former takes care of constructing the data that goes in the trie and I didn't want to spend time on recreating that logic separately. This does mean however that the benchmark also measures parsing and normalizing the contract metadata, I suspect which now takes the majority of the time.

### DIFF
--- a/crates/edr_solidity/benches/contracts_identifier.rs
+++ b/crates/edr_solidity/benches/contracts_identifier.rs
@@ -1,0 +1,71 @@
+//! Benchmark for construcing a contract identifier.
+//!
+//! Steps to run:
+//! 1. Check out https://github.com/NomicFoundation/forge-std/tree/js-benchmark-config
+//!    locally (note the branch).
+//! 2. In the `forge-std` repo root:
+//! 2.1. `npm i`
+//! 2.2. `npx hardhat compile`
+//! 3. In the `crates/edr_solidity` directory:
+//! 3.1. `export FORGE_STD_ARTIFACTS_DIR=/path/to/forge-std/artifacts`
+//! 3.2. `cargo bench contracts_identifier`
+use std::{fs, path::PathBuf, time::Duration};
+
+use criterion::{black_box, criterion_group, criterion_main, Criterion};
+use edr_solidity::{
+    artifacts::BuildInfo,
+    contract_decoder::{BuildInfoConfig, ContractDecoder},
+};
+
+fn load_build_info_config() -> anyhow::Result<BuildInfoConfig> {
+    let artifacts_dir = std::env::var("FORGE_STD_ARTIFACTS_DIR")?;
+    let build_info_dir = PathBuf::from(&artifacts_dir).join("build-info");
+
+    let mut build_infos = Vec::new();
+    for entry in fs::read_dir(build_info_dir)? {
+        let entry = entry?;
+        let path = entry.path();
+
+        if path.is_file() && path.extension().and_then(|ext| ext.to_str()) == Some("json") {
+            let contents = fs::read(&path)?;
+            let build_info = serde_json::from_slice::<BuildInfo>(&contents)?;
+            build_infos.push(build_info);
+        }
+    }
+
+    println!("build infos len: {}", build_infos.len());
+
+    Ok(BuildInfoConfig {
+        build_infos: Some(build_infos),
+        ignore_contracts: None,
+    })
+}
+
+pub fn criterion_benchmark(c: &mut Criterion) {
+    let build_info_config = load_build_info_config().expect("loads build info config");
+    let contracts = &build_info_config
+        .build_infos
+        .as_ref()
+        .expect("loaded build info")
+        .first()
+        .expect("there is at least one build info")
+        .output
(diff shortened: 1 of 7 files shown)


agostbiro:
Do we want a changelog item for this? It's not a user reported issue.

crates/edr_solidity/src/bytecode_trie.rs
```diff
@@ -1,83 +1,417 @@
...
+
+        cursor.match_.as_ref().and_then(|item| {
+            if cursor.prefix.range_end == key.len() {
+                // If the cursor's range ends where the key ends, we have a hit.
+                Some(TrieSearch::ExactHit(item.clone()))
             } else {
-                return Some(TrieSearch::LongestPrefixNode(cursor));
+                // Otherwise the cursor's range is greater than the key's length which means the
+                // key is a prefix of the match.
+                None
```
Wodann, line 136: Is this the case where the key is "foo" and a stored entry is "foobar"?

Wouldn't we still be interested in that case? Or is there a separate function for that?

crates/edr_solidity/src/bytecode_trie.rs
```diff
@@ -1,83 +1,417 @@
...
+    /// Fill the split node from the node with the node
+    /// and the new key as children. If the new key is shorter than the
+    /// node's prefix, the new split node will be a match for the new
+    /// key. Panics if `split_index > new_key.key().len()`
+    fn fill_split_node(&mut self, node_to_split: BytecodeTrie<T>, new_item: T) {
+        // We use the descendants to keep track of insertion order, so it's
+        // important to preserve that order here
+        let mut descendants = Vec::with_capacity(node_to_split.descendants.len() + 1);
+        descendants.extend(node_to_split.descendants.iter().cloned());
+        descendants.push(new_item.clone());
```
Wodann, line 187: I don't think this variable is used in this function

Wodann (approved):
I reviewed the commit mentioned in the description of the PR. Overall LGTM. Just one question and one comment

crates/edr_solidity/benches/contracts_identifier.rs
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    Ok(Some(BuildInfoConfig {
+        build_infos: Some(build_infos),
+        ignore_contracts: None,
+    }))
+}
+
+pub fn criterion_benchmark(c: &mut Criterion) {
+    let Some(build_info_config) = load_build_info_config().expect("loads build info config") else {
```
Wodann, line 48: You might be able to avoid the overhead of loading using this API of criterion:

https://bheisler.github.io/criterion.rs/book/user_guide/benchmarking_with_inputs.html

crates/edr_solidity/src/bytecode_trie.rs
agostbiro: Uhh that's something I broke when fixing the last bug. Thanks for catching it! Fixed + tested in 
[bc378e0](https://github.com/NomicFoundation/edr/pull/775/commits/bc378e065ff3c49e9a69f6c1bbedff4ae690ebca)

crates/edr_solidity/src/bytecode_trie.rs
agostbiro: > Is this the case where the key is "foo" and a stored entry is "foobar"?
Yes, we'd return `None` in this case.

> Wouldn't we still be interested in that case? Or is there a separate function for that?
Yes, normally this is an interesting result for trie lookups, but for the purposes of the contract identifier, it's a no match, so we're following the previous implementation here by returning `None`.

crates/edr_solidity/benches/contracts_identifier.rs
agostbiro: Thanks, could you explain this a bit more? I don't follow.

agostbiro:
The `seaport` benchmark regression is a false positive due to benchmark variability.

PR #778, opened 2025-01-27
fix: statically link Windows C runtime

Wodann:
Fixes https://github.com/NomicFoundation/edr/issues/781

Tested locally on a fresh Windows install.


nebasuke (approved):
Thanks Remco!

PR #780, opened 2025-01-29
fix: pass build info as buffer

agostbiro:
Pass build info in tracing config as buffer to avoid FFI copy overhead.

Also upgraded to Hardhat 2.22.18 and added a patch with the changes needed in Hardhat. 

Draft PR in Hardhat with the required changes: https://github.com/NomicFoundation/hardhat/pull/6200/files


crates/edr_napi/src/provider.rs
```diff
@@ -245,6 +247,31 @@ impl Provider {
     }
 }
 
+/// Tracing config for Solidity stack trace generation.
+#[napi(object)]
+pub struct TracingConfigWithBuffers {
+    /// Build information to use for decoding contracts.
+    pub build_infos: Option<Vec<Uint8Array>>,
```
agostbiro, line 254: Wrapping in an `Option` is a change from the design doc, but it maintains backwards compatibility as the `TracingConfig` in Hardhat has the `buildInfo` as an optional parameter.

fvictorio (approved):

PR #782, opened 2025-01-30
ci: upgrade upload/download artifacts to v4

agostbiro:
Due to deprecation of v3: https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/

### DIFF
--- a/.github/workflows/edr-npm-release.yml
+++ b/.github/workflows/edr-npm-release.yml
@@ -144,12 +144,13 @@ jobs:
         if: ${{ !matrix.settings.docker }}
         shell: bash
       - name: Upload artifact
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
           name: bindings-${{ matrix.settings.target }}
           # The upload artifact action doesn't respect the working directory setting. Unclear if this is a bug or not
           # https://github.com/actions/upload-artifact/issues/294
           path: ./crates/edr_napi/${{ env.APP_NAME }}.*.node
+          overwrite: true
           if-no-files-found: error
   test-macOS-windows-binding:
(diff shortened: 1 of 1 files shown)


.github/workflows/edr-npm-release.yml
```diff
@@ -144,12 +144,13 @@ jobs:
...
         shell: bash
       - name: Upload artifact
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
           name: bindings-${{ matrix.settings.target }}
           # The upload artifact action doesn't respect the working directory setting. Unclear if this is a bug or not
           # https://github.com/actions/upload-artifact/issues/294
           path: ./crates/edr_napi/${{ env.APP_NAME }}.*.node
+          overwrite: true
```
agostbiro, line 153: This is to preserve previous behavior after this breaking change:

> Uploading to the same named Artifact multiple times.
> 
> Due to how Artifacts are created in this new version, it is no longer possible to upload to the same named Artifact multiple times. You must either split the uploads into multiple Artifacts with different names, or only upload once. Otherwise you will encounter an error.

https://github.com/actions/upload-artifact?tab=readme-ov-file#breaking-changes

galargh: Actually, we don't have to add this. We're uploading to each `bindings-${{ matrix.settings.target }}` only once **per workflow**. That's OK with the new actions as well. What's no longer supported is duplicated uploads to the same name. It doesn't hurt to add it, though.

```suggestion
```

galargh (approved):

.github/workflows/edr-npm-release.yml
agostbiro: Ah ok thanks removed it

PR #790, opened 2025-02-05
feat: stack traces for unit tests

agostbiro:
Done in this PR:

- Add stack trace support for unit tests by re-executing deployment, setup and the test for each failing test.
   - This approach avoids performance penalty from stack trace generation on the happy path.
    - The alternative to re-executing deployment and setup even if those didn’t fail would be enabling trace collection by default for all deployment and setup which is expensive.
- Build info is passed in as unparsed buffers to avoid FFI copy overhead. Build info is only parsed if a stack trace needs to be generated as parsing it is an expensive operation.
- The trace option from `SolidityTestRunnerConfigArgs` is removed as stack traces are generated automatically.

TODOS left for stack traces after this PR:

- Re-execution for deployment/setup/fuzz/invariant tests
- Detect if an impure cheatcode was used
- Make sure re-execution is safe in fork mode based on whether the block spec is cacheable

No changeset as this gets released to `@ignored/edr`.


crates/edr_solidity_tests/src/runner.rs
```diff
@@ -95,42 +106,211 @@ impl<'a> ContractRunner<'a> {
...
+
+        let has_invariants = self
+            .contract
+            .abi
+            .functions()
+            .any(foundry_evm::abi::TestFunctionExt::is_invariant_test);
+
+        // Invariant testing requires tracing to figure out what contracts were created.
+        let tmp_tracing = executor.inspector.tracer.is_none() && has_invariants && needs_setup;
+        // let tmp_tracing = true;
```
Wodann, line 178: Leftover?

crates/edr_solidity_tests/src/runner.rs
```diff
@@ -95,42 +106,211 @@ impl<'a> ContractRunner<'a> {
...
-        }
+    fn setup(&self, executor: &mut Executor, needs_setup: bool) -> TestSetup {
+        self._setup(executor, needs_setup)
+            .unwrap_or_else(|err| TestSetup::failed(err.to_string()))
     }
 
-    fn _setup(&mut self, setup: bool) -> Result<TestSetup> {
-        trace!(?setup, "Setting test contract");
+    fn _setup(&self, executor: &mut Executor, needs_setup: bool) -> Result<TestSetup> {
+        trace!(?needs_setup, "Setting test contract");
```
Wodann, line 297: Question: What does `?` mean here?

crates/edr_solidity_tests/src/stack_trace.rs
```diff
@@ -0,0 +1,294 @@
...
+    let mut last_trace = None;
+    for (_, trace) in traces {
+        for node in trace.nodes() {
+            let address = node.trace.address;
+            if node.trace.kind.is_any_create() {
+                address_to_creation_code.insert(address, &node.trace.data);
+                address_to_runtime_code.insert(address, &node.trace.output);
+            }
+        }
+        last_trace = Some(trace);
```
Wodann, line 62: Wouldn't it be more efficient and easy to read if you do a separate `let last_trace = traces.last()`?

Wodann (approved):
I have a couple of questions about the code changes, but overall the changes look good to me. Nice work!

I did notice that there is a regression in the benchmark, there are some failing tests, and the `cargo doc` task is throwing some [warnings](https://github.com/NomicFoundation/edr/actions/runs/13167397162/job/36751040013?pr=790#step:4:1583) in your branch (unrelated to this change).

crates/edr_solidity_tests/src/runner.rs
agostbiro: It's short hand for recording fields for structured tracing. Docs here (need to scroll down a bit): https://docs.rs/tracing/latest/tracing/#recording-fields

crates/edr_napi/src/provider.rs
```diff
@@ -259,6 +259,22 @@ pub struct TracingConfigWithBuffers {
     pub ignore_contracts: Option<bool>,
 }
 
+impl std::fmt::Debug for TracingConfigWithBuffers {
```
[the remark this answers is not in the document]
agostbiro: No, unfortunately `Uint8Array` doesn't have a `Debug` implementation.

crates/edr_solidity_tests/src/stack_trace.rs
agostbiro: Good catch thanks. This is a left-over from an earlier iteration where the function took an `IntoIterator` instead of a slice. Fixed in https://github.com/NomicFoundation/edr/pull/790/commits/6859268e37f49f97d142c846bfb6b94d4ef6c03f

crates/edr_solidity_tests/src/runner.rs
agostbiro: Thanks, fixed in https://github.com/NomicFoundation/edr/pull/790/commits/6859268e37f49f97d142c846bfb6b94d4ef6c03f

agostbiro:
> I have a couple of questions about the code changes, but overall the changes look good to me. Nice work!

Thanks!

> I did notice that there is a regression in the benchmark, there are some failing tests (which I'm re-running to validate), and the `cargo doc` task is throwing some [warnings](https://github.com/NomicFoundation/edr/actions/runs/13167397162/job/36751040013?pr=790#step:4:1583) in your branch (unrelated to this change).

- The regression is on a fork test and hopefully only due to remote latencies. Rerunning to valdiate.
- Fixed failing tests in https://github.com/NomicFoundation/edr/pull/790/commits/18e3abd044e3f3624fa7e77d6c57df679ee9e8ea by passing an empty tracing config for `edr_napi` tests.
- Added an issue to fix doc warnings: https://github.com/NomicFoundation/edr/issues/792

agostbiro:
I identified two sources of performance regressions:
- There are some failing tests in the `forge-std` test suite that we're using as a benchmark suite due to unsupported cheatcodes. Previously we'd special case these in the benchmark reporting, but now stack traces are generated for them which is a performance hit. So I just commented them out in our `forge-std` fork: https://github.com/NomicFoundation/forge-std/commit/19500a0799ac49836d196c9ea9fd3393e43e3b56
- We're temporarily using two different versions of REVM in `edr_solidity_tests` to bridge between EDR's stack trace heuristics and Foundry's traces. This leads to a <5% performance regression probably due to preventing some compiler optimizations.

crates/edr_solidity_tests/src/runner.rs
Wodann: Thanks!

PR #797, opened 2025-02-07
refactor: compile time guarantees to prevent secret key leakage

agostbiro:
Refactoring to provide compile time guarantees to prevent secret key leakage through error reporting (e.g. Sentry).

Normally the secret keys that EDR receives shouldn't hold any funds, but users can make mistakes, so we have to do our best to protect the keys from being accidentally exposed through error reporting.

The compile time guarantee that we need is that types that hold secret keys have no `Display` and `Serialize` implementations and if they have a `Debug` implementation, it's opaque. If this is fulfilled, there is no way to accidentally leak secret keys into error reports, because there is no way to display them.

Prior to this PR there were two areas that were risky:

1. Keys were passed from NodeJS as `String`. Since `String` has `Debug`, `Display,` and `Serialize` implementations, it could be leaked into error reports unintentionally.
2. `edr_provider::AccountConfig` which holds a `k256::SecretKey` had a custom serialization implementation to work around `k256::SecretKey` not having one. `edr_provider::AccountConfig` needed a serialization implementation collect benchmark scenarios.

Issue 1. was solved by switching from `String` to `napi::JsString` in the provider napi config type as `napi::JsString` doesn't have `Debug`, `Display`, or `Serialization` implementation. I also moved helper functions that take native string arguments to `edr_test_utils` to discourage future changes that'd introduce storing secret keys in native strings. Note that if there is a crash in the napi code with a message that leaks data before the buffer is converted to `napi::JsString`, that could still potentially leak to Sentry, but there is nothing we can do about this.

Issue 2. was solved by removing the `Serialize` implementation from `edr_provider::AccountConfig` and introducing an `edr_scenarios` crate that contains provider config types that support serialization. `edr_scenarios` is not used in production code.


crates/edr_napi/src/account.rs
```diff
@@ -17,7 +25,17 @@ impl TryFrom<GenesisAccount> for edr_provider::AccountConfig {
     type Error = napi::Error;
```
agostbiro, line 25: The errors that can be propagated from here are opaque.

crates/edr_napi/src/account.rs
```diff
@@ -17,7 +25,17 @@ impl TryFrom<GenesisAccount> for edr_provider::AccountConfig {
...
-        let secret_key = secret_key_from_str(&value.secret_key)
+        static_assertions::assert_not_impl_all!(JsString: Debug, Display, Serialize);
+        // `k256::SecretKey` has `Debug` implementation, but it's opaque (only shows the
+        // type name)
+        static_assertions::assert_not_impl_any!(k256::SecretKey: Display, Serialize);
+
+        let secret_key = value.secret_key.into_utf8()?;
+        // This is the only place in production code where it's allowed to use
+        // `DangerousSecretKeyStr`.
+        #[allow(deprecated)]
```
Wodann, line 36: I'm wondering whether `deprecated` signals the correct thing. Would this be a good place to use `unsafe`?

An alternative would be to convert `DangerousSecretKeyStr` into an `OpaqueString` type that contains a private string variable and doesn't implement `Debug` or `Display`. Its stored string can be exposed as `str` with a function (potentially using `unsafe`)

crates/edr_eth/src/signature.rs
```diff
@@ -27,6 +27,9 @@ pub enum SignatureError {
     /// Invalid secret key.
     #[cfg_attr(feature = "std", error("Expected 32 byte secret key"))]
     InvalidSecretKeyLength,
+    /// When parsing a secret key from string to hex
+    #[cfg_attr(feature = "std", error("Invalid hex"))]
+    InvalidSecretKeyHex,
     /// When parsing a signature from string to hex
     #[cfg_attr(feature = "std", error(transparent))]
     DecodingError(#[cfg_attr(feature = "std", from)] hex::FromHexError),
```
Wodann, line 35: Should this enum variant be removed now?

agostbiro: I pushed some changes after submitting the PR, so not sure if you're referring to this line:

>     /// When parsing a secret key from string to hex
>     #[cfg_attr(feature = "std", error("Invalid hex"))]
>     InvalidSecretKeyHex,

If yes, we need it as an opaque hex error type (the error from the `hex` crate can contain a character from the parsed string).

crates/edr_napi/src/account.rs
agostbiro: The message I was going for is "don't use this". The best solution would be imo to make the function private in `edr_napi`, but then there is duplication would be required with `edr_test_utils`. I agree that deprecated is a bit awkward, but it's closer to the "don't use this" semantics for me than `unsafe` as that one is tied to memory safety in my mind which is not the case here.

crates/edr_eth/src/signature.rs
Wodann: I'm referring to:

```rs
    DecodingError(#[cfg_attr(feature = "std", from)] hex::FromHexError),
```

crates/edr_napi/src/account.rs
Wodann: > unsafe [..] is tied to memory safety in my mind which is not the case here.

That a valid point.

I was hoping for an alternative to `#[deprecated]` that causes a compiler error instead of a warning. I did some research but didn't come up with anything 😔 There is no way to configure a project-specific clippy error, otherwise that would have been nice

Wodann (approved):
Just [this conversation](https://github.com/NomicFoundation/edr/pull/797#discussion_r1947035187) is open-ended. If the `DecodingError` enum variant is actually still used in the codebase (and that's not a concern), feel free to merge

crates/edr_napi/src/account.rs
agostbiro: Yeah a custom clippy error would be perfect. But we're running clippy with `-D warnings`, so using the struct will result in a CI error unless the usage is marked with `#[allow(deprecated)]`

crates/edr_eth/src/signature.rs
agostbiro: Ah `DecodingError` is still used when decoding signatures from hex. It's ok to propagate the hex error there, since signatures are not secret.

PR #799, opened 2025-02-10
feat: stack trace support for setup/deployment and invariant tests

agostbiro:
Adds stack trace support by re-executing failed calls for

- deployment and setup for all types of tests
- invariant tests

Re-execution on deployment and setup is rather forward, but for invariant tests there are two things to consider:

1. Previously, we only did tracing on setup and deployment if the invariant test had a `setUp` method. It greatly simplifies re-execution if we always do tracing on setup and deployment if there are invariant tests. The performance decrease should be negligible, because most invariant test have a `setUp` method and since they tend to take longer, the overhead is amortized.
2.  The way the revert reason is returned for invariant tests is a bit weird in Foundry. Sometimes it's not returned even if there is a revert and when replaying a recorded invariant failure, the rather obscure "replay failure" is [returned.](https://github.com/NomicFoundation/edr/compare/feat/solidity-tests...feat/stack-trace-replay?expand=1#diff-27fd00367d0897c3a5811c73b8615732cf5286feccc4fa83eec55fb234baff68L1062-L1072) I think the reason for this is that they're doing tracing by default on re-execution and display the revert reason from the traces instead of `TestResult::reason`. I opted to change this behavior so that the revert reason in the test result matches the stack trace which Hardhat relies on.  Note that there are some failing Foundry integration tests related to this which I'll fix if we agree on this point.

TODOS left for stack traces after this PR:

- Re-execution for fuzz tests
- Make sure re-execution is safe in fork mode based on whether the block spec is cacheable
- Detect if an impure cheatcode was used

No changeset as this gets released to @ignored/edr.

### DIFF
--- a/crates/edr_solidity_tests/src/runner.rs
+++ b/crates/edr_solidity_tests/src/runner.rs
@@ -866,16 +866,16 @@ impl<'a, NestedTraceDecoderT: SyncNestedTraceDecoder> ContractRunner<'a, NestedT
                                 }
                                 counterexample =
                                     Some(CounterExample::Sequence(counterexample_sequence));
+                            }
 
-                                // If we can't get a revert reason for the second time, we couldn't
-                                // replay the failure, so keep the original revert reason
-                                // and discard the stack trace as it may be misleading.
-                                if reason.is_some() && revert_reason.is_none() {
-                                    tracing::warn!(?invariant_contract.invariant_function, "Failed to compute stack trace");
-                                } else {
-                                    stack_trace = stack_trace_result;
-                                    reason = revert_reason;
-                                }
+                            // If we can't get a revert reason for the second time, we couldn't
+                            // replay the failure, so keep the original revert reason
+                            // and discard the stack trace as it may be misleading.
+                            if reason.is_some() && revert_reason.is_none() {
+                                tracing::warn!(?invariant_contract.invariant_function, "Failed to compute stack trace");
+                            } else {
+                                stack_trace = stack_trace_result;
+                                reason = revert_reason;
                             }
                         }
                         Err(err) => {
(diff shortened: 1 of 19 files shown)


Wodann:
> The way the revert reason is returned for invariant tests is a bit weird in Foundry. Sometimes it's not returned even if there is a revert and when replaying a recorded invariant failure, the rather obscure "replay failure" is [returned.](https://github.com/NomicFoundation/edr/compare/feat/solidity-tests...feat/stack-trace-replay?expand=1#diff-27fd00367d0897c3a5811c73b8615732cf5286feccc4fa83eec55fb234baff68L1062-L1072) I think the reason for this is that they're doing tracing by default on re-execution and display the revert reason from the traces instead of TestResult::reason. I opted to change this behavior so that the revert reason in the test result matches the stack trace which Hardhat relies on. Note that there are some failing Foundry integration tests related to this which I'll fix if we agree on this point.

Based on this context, it's hard for me to agree or disagree. Maybe we can discuss it tomorrow during our sync?

crates/edr_solidity_tests/src/runner.rs
```diff
@@ -1047,31 +1084,41 @@ fn try_to_replay_recorded_failures(executor: Executor, args: ReplayArgs<'_>) ->
...
                     traces,
                     coverage,
-                    txes,
+                    inputs: txes,
+                    generate_stack_trace: true,
+                    contract_decoder: Some(contract_decoder),
+                    revert_decoder,
+                })
+                .ok()
+                .map_or_else(
```
Wodann, line 1101: Can't these two lines just be a single `map_or_else`?

js/helpers/package.json
```diff
@@ -31,7 +31,7 @@
   "private": true,
   "scripts": {
     "build": "pnpm run build:dev",
-    "build:dev": "tsc --build --incremental .",
```
Wodann, line 34: Was this causing issues?

Wodann (approved):
The code changes look good to me. As I mentioned previously, it would be good to discuss the approach taken during the sync, to help me verify my understanding before I can agree/disagree.

I also left one question and one comment on the code changes themselves.

js/helpers/package.json
agostbiro: Yeah it often fails to build with incremental on. Should've fixed in an other PR, ended up committing here by accident

Wodann (approved):
Final changes look good to me

Issue #804, opened 2025-02-18
Add support for EIP-7702


Wodann:
**Definition of done**
Implement EIP-7702 according to the agreed upon technical design

PR #805, opened 2025-02-18
feat: add EIP-7702 transactions

Wodann:
Implements EIP-7702 transactions according to the agreed upon technical design.

Completes #804

### DIFF
--- a/.changeset/sour-coats-look.md
+++ b/.changeset/sour-coats-look.md
@@ -1,0 +1,5 @@
+---
+"@nomicfoundation/edr": patch
+---
+
+Adds EIP-7702 transactions to eth_call, eth_estimateGas, eth_sendTransaction, eth_sendRawTransaction, and debug_traceCall
--- a/crates/edr_eth/src/signature/y_parity.rs
+++ b/crates/edr_eth/src/signature/y_parity.rs
@@ -1,31 +1,48 @@
 use alloy_rlp::BufMut;
 use k256::SecretKey;
+use revm_primitives::PrimitiveSignature;
 
 use super::{Recoverable, RecoveryMessage, Signature, SignatureError, SignatureWithRecoveryId};
 use crate::{Address, U256};
 
-#[derive(Clone, Copy, Debug, Eq, PartialEq, Hash)]
+#[derive(Clone, Debug, Eq, PartialEq, Hash)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 /// An ECDSA signature with Y-parity.
-pub struct SignatureWithYParity {
-    /// R value
+pub struct SignatureWithYParity(PrimitiveSignature);
+
+/// Arguments for constructing a new `SignatureWithYParity`.
+pub struct Args {
+    /// The `r` value of the signature.
     pub r: U256,
-    /// S value
+    /// The `s` value of the signature.
     pub s: U256,
-    /// Whether the V value has odd Y parity.
+    /// The Y-parity of the signature.
     pub y_parity: bool,
 }
 
 impl SignatureWithYParity {
+    /// Constructs a new instance from the provided `r`, `s`, and `y_parity`
+    /// values.
+    pub fn new(args: Args) -> Self {
+        let Args { r, s, y_parity } = args;
+
+        Self(PrimitiveSignature::new(r, s, y_parity))
+    }
+
     /// Constructs a new instance from a message and secret key.
     ///
     /// To obtain the hash of a message consider
     /// [`crate::utils::hash_message`].
-    pub fn new<M>(message: M, secret_key: &SecretKey) -> Result<Self, SignatureError>
+    pub fn with_message<M>(message: M, secret_key: &SecretKey) -> Result<Self, SignatureError>
     where
         M: Into<RecoveryMessage>,
     {
         SignatureWithRecoveryId::new(message, secret_key).map(SignatureWithYParity::from)
+    }
+
+    /// Returns the inner `PrimitiveSignature`.
+    pub const fn into_inner(self) -> PrimitiveSignature {
+        self.0
     }
 }
 
@@ -33,14 +50,7 @@ impl SignatureWithYParity {
 // list.
 impl alloy_rlp::Decodable for SignatureWithYParity {
     fn decode(buf: &mut &[u8]) -> alloy_rlp::Result<Self> {
-        let decoded = Self {
-            // The order of these fields determines decoding order.
-            y_parity: bool::decode(buf)?,
-            r: U256::decode(buf)?,
-            s: U256::decode(buf)?,
-        };
-
-        Ok(decoded)
+        PrimitiveSignature::decode_rlp_vrs(buf, bool::decode).map(Self)
     }
 }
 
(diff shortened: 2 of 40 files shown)


.changeset/sour-coats-look.md
```diff
@@ -0,0 +1,5 @@
+---
+"@nomicfoundation/edr": patch
+---
+
+Adds EIP-7702 transactions to eth_call, eth_estimateGas, eth_sendTransaction, eth_sendRawTransaction, and debug_traceCall
```
agostbiro, line 5: Question: support for `debug_traceCall` is not in the design doc. Is this intentional? If we support `debug_traceCall` should we support `debug_traceTransaction` as well? cc @fvictorio

crates/edr_eth/src/signature/y_parity.rs
```diff
@@ -1,80 +1,81 @@
...
-    pub s: U256,
-    /// Whether the V value has odd Y parity.
-    pub y_parity: bool,
-}
+pub struct SignatureWithYParity(PrimitiveSignature);
 
 impl SignatureWithYParity {
+    /// Constructs a new instance from the provided `r`, `s`, and `y_parity`
+    /// values.
+    pub fn new(r: U256, s: U256, y_parity: bool) -> Self {
```
agostbiro, line 16: Nit: it's easy to mix up `r` and `s`, because they're consecutive parameters with the same type. For me a struct providing named parameters would be better as argument, but it's not a blocker.

crates/edr_eth/src/transaction/request/eip7702.rs
```diff
@@ -0,0 +1,176 @@
...
+        unsafe { self.sign_for_sender_unchecked(secret_key, caller) }
+    }
+
+    /// Signs the transaction with the provided secret key, belonging to the
+    /// provided caller's address.
+    ///
+    /// # Safety
+    ///
+    /// The `caller` and `secret_key` must correspond to the same account.
+    pub unsafe fn sign_for_sender_unchecked(
```
agostbiro, line 50: Nit: this would be ideally restricted to `pub(super)` as it's dangereous, but I see that the other similar methods are public as well.

crates/edr_eth/src/transaction/signed/eip7702.rs
```diff
@@ -0,0 +1,292 @@
...
+            let request = expectation::request()?;
+
+            let secret_key = expectation::secret_key()?;
+            let signed = request.sign(&secret_key)?;
+
+            Ok(signed.into())
+        }
+
+        pub fn secret_key() -> Result<SecretKey, SignatureError> {
+            secret_key_from_str(SECRET_KEYS[0])
```
agostbiro, line 222: This and other compilation errors would be fixed by using `edr_test_utils::secret_key::secret_key_from_str` from `main`.

crates/edr_provider/tests/eip7702.rs
```diff
@@ -0,0 +1,361 @@
...
+fn sign_authorization(
+    authorization: eip7702::Authorization,
+    secret_key: &SecretKey,
+) -> anyhow::Result<eip7702::SignedAuthorization> {
+    let signature = SignatureWithYParity::with_message(authorization.signature_hash(), secret_key)?;
+
+    Ok(authorization.into_signed(signature.into_inner()))
+}
+
+// #[tokio::test(flavor = "multi_thread")]
```
agostbiro, line 72: Is this commented out on purpose?

crates/edr_provider/src/lib.rs
```diff
@@ -482,6 +482,15 @@ impl<LoggerErrorT: Debug + Send + Sync + 'static, TimerT: Clone + TimeSinceEpoch
     }
 }
 
+#[cfg(feature = "test-utils")]
+impl<LoggerErrorT: Debug + Send + Sync + 'static, TimerT: Clone + TimeSinceEpoch>
+    Provider<LoggerErrorT, TimerT>
+{
+    pub fn data_mut(&self) -> tokio::sync::MutexGuard<'_, ProviderData<LoggerErrorT, TimerT>> {
```
agostbiro, line 489: This doesn't seem to be used.

crates/edr_provider/tests/eip7702.rs
```diff
@@ -0,0 +1,361 @@
+#![cfg(feature = "test-utils")]
```
agostbiro, line 1: It's difficult for me to see if all the test cases from the design doc are covered in this module. Can you please clarify if they are covered?

agostbiro (commented):
Hey, thanks for this PR! Looks good overall, but I have some questions that should be cleared up I think before merging.

.changeset/sour-coats-look.md
Wodann: `debug_traceTransaction` receives the transaction hash of a previously mined transaction, so I didn't mention it explicitly. I tallied it up with methods like `eth_getTransactionByHash` which should also return an EIP-7702 transaction.

I can add some tests to validate that those succeed though.

crates/edr_eth/src/transaction/request/eip7702.rs
Wodann: It's marked as `unsafe` to indicate this danger. The safety doc explains what assumption exists about the relationship between the input arguments.

crates/edr_eth/src/signature/y_parity.rs
Wodann: Thanks for the suggestion. Suggestion applied [here](https://github.com/NomicFoundation/edr/pull/805/commits/ee256e26ea5e9bc73f9b6e8f30265ba47d62c057)

crates/edr_provider/src/lib.rs
Wodann: Good catch. This was leftover from a prior iteration. Resolved in https://github.com/NomicFoundation/edr/pull/805/commits/e41b0a759e8aee62d28366de5588c5e1d11d33e1

.changeset/sour-coats-look.md
Wodann: I added the tests I mentioned in https://github.com/NomicFoundation/edr/pull/805/commits/56a5820188862626d92a325e4c84fb1df869786c

crates/edr_provider/tests/eip7702.rs
Wodann: Resolved by https://github.com/NomicFoundation/edr/pull/805/commits/56a5820188862626d92a325e4c84fb1df869786c

.changeset/sour-coats-look.md
fvictorio: Maybe to expand on what Remco said: the authorizations part of EIP-7702 transactions is pretty much orthogonal to the execution part (similar to blob transactions, although blobs can be observed by the EVM; authorizations can't). So it makes sense for this to come for free without much extra effort.

crates/edr_eth/src/transaction/request/eip7702.rs
agostbiro: Yeah I agree that this is ok if something has to be public, but wanted to point out that it's only used in the module, so it could be just as well `pub(super)`.

agostbiro (approved):
Thanks for the changes and comments! LGTM

PR #808, opened 2025-02-20
feat: add support for EIP-2537

### LINKED ISSUE #807: Validate support for EIP-2537
**Definition of done**
Add tests to validate EIP-2537 is supported by EDR according to the agreed upon technical design

Wodann:
Resolves #807 

As part of this PR, I restructured the integration tests for `edr_provider` to ensure that only one binary is built for the newly added EIP-7702 and EIP-2537 tests.

In the `feat/multichain` branch I already made all changes to ensure that only one binary is built for all integration tests - although those change use the multichain EDR APIs. This change should result in minimal merge conflicts once we merge the two branches together - in the future.


Cargo.toml
```diff
@@ -21,8 +21,8 @@ rpath = true
 rpath = true
 
 [workspace.dependencies]
-alloy-eips = { version = "0.11.0", default-features = false }
-alloy-serde = { version = "0.11.0", default-features = false}
+alloy-eips = { version = "0.8.0", default-features = false }
+alloy-serde = { version = "0.8.0", default-features = false}
```
agostbiro, line 25: I'm wondering why the downgrade here? This can be a problem when merging `main` into `feat/solidity-tests` as the Foundry crates are only tested to work with REVM 19.4 and `alloy-* = 0.11`

Wodann: When enabling the `blst` feature flag, which is required to support this EIP, I was getting `cargo` errors due to incompatible versions of `blst` due to dynamically linking the dependency. I had to revert to the same `alloy` versions REVM is using to guarantee that the code could compile.

agostbiro: Ah I see thanks for the explanation! Could you try to upgrade to `alloy-* = 0.11.1`, `revm = 19.5` and `revm-precompile=16.1.0`? Those should work together based on the Foundry `Cargo.lock`:

https://github.com/foundry-rs/foundry/blob/f3130a52937039a7e82b4aef07143139c322d09f/Cargo.lock

Wodann: I will try; thanks!

Wodann: That worked: https://github.com/NomicFoundation/edr/pull/808/commits/7bb29d226458f38624923bfbc83b76755a70bccf

Thanks for finding a fix that doesn't cause any issues on your branch! :D

PR #810, opened 2025-02-21
feat: add support for EIP-2935

### LINKED ISSUE #809: Add revert reason to signal unsupported EIP-2935
We won’t support [EIP-2935](https://eips.ethereum.org/EIPS/eip-2935) at first. We are assuming most people won’t need this, since it’s more of a forward-looking change for work related to light clients.

Instead of supporting it, we’ll set the bytecode of that system contract so that any calls or transactions sent to it revert with a clear message. This is the contract we’ll use:

```solidity
// SPDX-License-Identifier: Unlicense
pragma solidity ^0.8.0;

contract EIP2935NotAvailable {
  fallback() external payable {
    revert("EIP2935 is not supported in Hardhat yet: https://github.com/NomicFoundation/hardhat/issues/6226");
  }
}
```

Compiling this code with solc 0.8.28 (without optimizations) results in the following runtime bytecode:

```
0x60806040526040517f08c379a0000000000000000000000000000000000000000000000000000000008152600401610036906100e8565b60405180910390fd5b600082825260208201905092915050565b7f45495032393335206973206e6f7420737570706f7274656420696e204861726460008201527f686174207965743a2068747470733a2f2f6769746875622e636f6d2f4e6f6d6960208201527f63466f756e646174696f6e2f686172646861742f6973737565732f3632323600604082015250565b60006100d2605f8361003f565b91506100dd82610050565b606082019050919050565b60006020820190508181036000830152610101816100c5565b905091905056fea26469706673582212209faaba1140034f51b16123b009d46640f0f499cf7ca044d2901caa9cf73048dc64736f6c634300081c0033
```

We should set `HISTORY_STORAGE_ADDRESS` to this bytecode when `prague` is enabled.

### Test cases

- For a local blockchain with the `cancun` hardfork, the `HISTORY_STORAGE_ADDRESS` account is empty.
- For a local blockchain with the `prague` hardfork, the `HISTORY_STORAGE_ADDRESS` account is not empty. Sending a transaction to it results in the expected revert error.
- When forking, the `HISTORY_STORAGE_ADDRESS` shouldn’t be touched, no matter the local nor the remote hardfork.

Wodann:
Resolves #809.

In this PR I also restructured the integration tests of `edr_evm` to use a single binary (instead of one per `*.rs` file). This speeds up compilation times for EDR.

There was one "gotcha" while implementing this. Namely, it seems like the EIP-2935 bytecode has already been deployed to mainnet:

https://etherscan.io/address/0x0000F90827F1C53a10cb7A02335B175320002935#code

This means that - when forking - the address already has bytecode, also for older hardforks.

I added a test to signal this behaviour.


agostbiro (approved):
Thanks, LGTM!

PR #811, opened 2025-02-21
fix: return EIP-7702 properties for transactions

Wodann:
This fixes two issues with EIP-7702 transactions:
- We were never parsing remote transactions to `transaction::Signed`, instead recognising them as EIP-155 transactions
- We weren't returning EIP-7702 properties from the JSON-RPC interface


agostbiro (approved):
Thanks, LGTM

Issue #812, opened 2025-02-21
Add test for RLP-encoding Prague block header


Wodann:
**Definition of done**
A test that validates the RLP encoding and block hash for a Prague block

PR #814, opened 2025-02-21
feat: add EIP-7685 properties to block header

### LINKED ISSUE #813: Add EIP-7685 properties to block header
We won’t support EIP-7685, but we’ll add the new `requestsHash` header field that this EIP introduces.

From the EIP: «For a block with no requests data, the `requests_hash` is simply `sha256("")`.». That is, if `prague` is enabled, then blocks should have a hardcoded `requestsHash` field of `0xe3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855`.

Wodann:
Resolves #813.

I added a [follow-up task](https://github.com/NomicFoundation/edr/issues/812) for once a Prague block has been generated on Holesky.

### DIFF
--- a/.changeset/good-pears-draw.md
+++ b/.changeset/good-pears-draw.md
@@ -1,0 +1,5 @@
+---
+"@nomicfoundation/edr": patch
+---
+
+Adds the EIP-7685 requestsHash field to Block RPC response type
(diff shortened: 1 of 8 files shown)


crates/edr_napi/index.d.ts
```diff
@@ -42,6 +42,12 @@ export interface BlockOptions {
    * block (EIP-4788).
    */
   parentBeaconBlockRoot?: Buffer
+  /**
+   * The commitment hash calculated for a list of [EIP-7685] data requests.
+   *
+   * [EIP-7685]: https://eips.ethereum.org/EIPS/eip-7685
+   */
+  requestsHash?: Buffer
```
Wodann, line 50: This change is forward compatible as it allows being set to `undefined`

agostbiro (approved):
Thanks, LGTM

PR #816, opened 2025-02-21
test: add test for EIP-7691

### LINKED ISSUE #815: Add test for EIP-7691
[This EIP](https://eips.ethereum.org/EIPS/eip-7691) changes the blob target from 3 to 6, and the max number of blobs in a block from 6 to 9. We have [at least one test](https://github.com/NomicFoundation/edr/blob/90e3b1597dd876d8bd33eeff69e6822ec17de7a2/crates/edr_provider/tests/eip4844.rs#L396) around this that should be duplicated and adapted to assert different things before and after Prague.

Wodann:
Resolves #815 .

### DIFF
--- a/.changeset/good-pears-draw.md
+++ b/.changeset/good-pears-draw.md
@@ -1,5 +1,5 @@
 ---
-"@nomicfoundation/edr": patch
+"@nomicfoundation/edr": minor
 ---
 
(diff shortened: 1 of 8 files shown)


agostbiro (approved):
Thanks, LGTM

PR #818, opened 2025-02-21
test: add tests for EIP-7623

### LINKED ISSUE #817: Adds test for EIP-7623
The change in calldata cost introduced by [EIP-7623](https://eips.ethereum.org/EIPS/eip-7623) will come mostly free from REVM, but we should include a couple of smoke tests.

### Test cases

- **Sending data to an EOA**
    
    EIP-7623 increases the cost of sending a transaction with data that won’t involve any execution (the transaction neither creates a contract nor interacts with one).
    
    Sending this transaction should cost 21016 gas before Prague and 21040 after it:
    
    ```json
    {
      "from": "0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266",
      "to": "0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266",
      "data": "0x11"
    }
    ```
    
- **Deploying a contract**
    
    The EIP is designed in a way that means that the price of deploying or interacting with a contract won’t change. We’ll only test the scenario where a contract is deployed, because it’s simpler.
    
    Sending this transaction should cost 53409 both before and after Prague:
    
    ```json
    {
      "from": "0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266",
      "data": "0x600b380380600b5f395ff300"
    }
    ```

Wodann:
Fixes #817

### DIFF
--- a/crates/edr_provider/tests/integration/eip7623/send_data_to_eoa.rs
+++ b/crates/edr_provider/tests/integration/eip7623/send_data_to_eoa.rs
@@ -1,0 +1,54 @@
+use edr_eth::{address, bytes, transaction::EthTransactionRequest, SpecId};
+use edr_rpc_eth::CallRequest;
+
+use super::new_provider;
+use crate::integration::eip7623::assert_transaction_gas_usage;
+
+fn call_request() -> CallRequest {
+    let transaction_request = transaction_request();
+
+    CallRequest {
+        from: Some(transaction_request.from),
+        to: transaction_request.to,
+        data: transaction_request.data,
+        ..CallRequest::default()
+    }
+}
+
+fn transaction_request() -> EthTransactionRequest {
+    EthTransactionRequest {
+        from: address!("0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266"),
+        to: Some(address!("0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266")),
+        data: Some(bytes!("0x11")),
+        ..EthTransactionRequest::default()
+    }
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn estimate_gas() -> anyhow::Result<()> {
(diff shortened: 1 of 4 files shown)


crates/edr_provider/tests/integration/eip7623/send_data_to_eoa.rs
```diff
@@ -0,0 +1,54 @@
...
+    }
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn estimate_gas() -> anyhow::Result<()> {
+    let cancun_provider = new_provider(SpecId::CANCUN)?;
+    assert_eq!(
+        super::estimate_gas(&cancun_provider, call_request()),
+        // NOTE: Our estimate differs from the real cost by 1 gas unit.
+        21_017
```
Wodann, line 33: @fvictorio our gas estimate is off by one. Is that expected?

fvictorio: Could this be caused by https://github.com/NomicFoundation/edr/blob/404821a26234bf67c4772add83c7e593c086ec42/crates/edr_provider/src/data.rs#L788-L790 ?

agostbiro (approved):
Thanks, LGTM

crates/edr_provider/tests/integration/eip7623/send_data_to_eoa.rs
fvictorio: Tbh I don't remember why we added that in the first place, so I can't argue if it's desirable or not. But removing it would be risky unless we are sure it won't have an impact. The semantics of that method are "if you use this gas limit, then your transaction won't OOG, and whatever gas ends up being used is not going to be significantly lower than the one yo uset", so adding one wei is unlikely to be a problem.

Wodann: I'll leave it as is then

PR #820, opened 2025-02-25
misc: make EIP-7702 a breaking change

Wodann:
After discussing, we decided that the changes to the JSON-RPC should be a breaking change.


agostbiro (approved):
Thanks, LGTM

PR #822, opened 2025-02-27
fix: deterministic re-execution for stack traces

agostbiro:
We generate stack traces via re-execution of failing tests for performance reasons (stack traces need expensive EVM step tracing). 

Normally EVM execution is deterministic: if you apply the same calls on top of the same state you get the same results. Solidity tests have some features that break this deterministic behavior:
- Forking a remote with the "latest" block number.[^1] 
- Cheatcodes that are impure in the sense that their output depends on things outside of the state managed by EDR (e.g. the `ffi` cheatcode that executes arbitrary CLI commands or cheatcodes that access the file system).

The changes in this PR ensure that we don't generate stack traces if EVM execution is indeterministic. The current changes don't implement notifying the user about this yet, as how we go about that will need a design doc first to align with the HH team. Instead the `TestResult::stackTrace()` method just returns `None` if we prevented stack trace generation.

If we detect impure cheatcode usage, we collect the Solidity declaration of the cheatcode for use later in warnings/error messages (example declaration: `"function createSelectFork(string calldata urlOrAlias) external returns (uint256 forkId);"`).

[^1]: Technically, forking with unsafe block numbers is inderministic too as they can be reorged out, but we decided to ignore this for now as unsafe block numbers become safe quickly, so if a user experiences problems because of this, they get resolved if they rerun their tests.

### DIFF
--- a/crates/edr_napi/index.d.ts
+++ b/crates/edr_napi/index.d.ts
@@ -669,6 +669,52 @@ export interface AddressLabel {
   /** The label to assign to the address */
   label: string
 }
+/** The stack trace result */
+export interface StackTrace {
+  /** Enum tag for JS. */
+  kind: "StackTrace"
+  /** The stack trace entries */
+  entries: Array<SolidityStackTraceEntry>
+}
+/** We couldn't generate stack traces, because an unexpected error occurred. */
+export interface UnexpectedError {
+  /** Enum tag for JS. */
+  kind: "UnexpectedError"
+  /** The error message from the unexpected error. */
+  errorMessage: string
+}
+/**
+ * We couldn't generate stack traces, because the stack trace generation
+ * heuristics failed due to an unknown reason.
+ */
+export interface HeuristicFailed {
+  /** Enum tag for JS. */
+  kind: "HeuristicFailed"
+}
+/**
+ * We couldn't generate stack traces, because the test execution is unsafe to
+ * replay due to indeterminism. This can be caused by either specifying a fork
+ * url without a fork block number in the test runner config or using impure
+ * cheatcodes.
+ */
+export interface UnsafeToReplay {
+  /** Enum tag for JS. */
+  kind: "UnsafeToReplay"
+  /**
+   * Indeterminism due to specifying a fork url without a fork block number
+   * in the test runner config
+   */
+  globalForkLatest: boolean
+  /**
+   * The list of executed impure cheatcode signatures. We collect function
+   * signatures instead of function names as whether a cheatcode is impure
+   * can depend on the arguments it takes (e.g. `createFork` without a second
+   * argument means implicitly fork from “latest”). Example signature:
+   * `function createSelectFork(string calldata urlOrAlias) external returns
+   * (uint256 forkId);`.
+   */
+  impureCheatcodes: Array<string>
+}
 /**The result of a test execution. */
 export const enum TestStatus {
   /**Test success */
@@ -1041,11 +1087,12 @@ export declare class TestResult {
   readonly durationMs: bigint
   /**
    * Compute the error stack trace.
-   * If the heuristic failed, returns an empty array.
+   * The result is either the stack trace or the reason why we couldn't
+   * generate the stack trace.
    * Returns null if the test status is succeeded or skipped.
-   * Throws if there was an error computing the stack trace.
+   * Cannot throw.
    */
-  stackTrace(): Array<SolidityStackTraceEntry> | null
+  stackTrace(): StackTrace | UnexpectedError | HeuristicFailed | UnsafeToReplay | null
 }
 export declare class Exit {
   get kind(): ExitCode
--- a/crates/edr_napi/index.d.ts
+++ b/crates/edr_napi/index.d.ts
@@ -702,7 +702,7 @@ export interface UnsafeToReplay {
   kind: "UnsafeToReplay"
   /**
    * Indeterminism due to specifying a fork url without a fork block number
-   * in the test runner config
+   * in the test runner config.
    */
   globalForkLatest: boolean
   /**
--- a/crates/foundry/evm/core/src/backend/cow.rs
+++ b/crates/foundry/evm/core/src/backend/cow.rs
@@ -265,6 +265,23 @@ impl<'a> DatabaseExt for CowBackend<'a> {
     fn has_cheatcode_access(&self, account: &Address) -> bool {
         self.backend.has_cheatcode_access(account)
     }
+
+    fn record_cheatcode_purity(&mut self, cheatcode_name: &'static str, is_pure: bool) {
+        // Only convert to mutable if we need to update.
+        if !is_pure
+            && !self
+                .backend
+                .inner
+                .impure_cheatcodes
+                .contains(cheatcode_name)
+        {
+            self.backend
+                .to_mut()
+                .inner
+                .impure_cheatcodes
+                .insert(cheatcode_name);
+        }
+    }
 }
 
 impl<'a> DatabaseRef for CowBackend<'a> {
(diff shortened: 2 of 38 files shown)


crates/foundry/evm/core/src/backend/cow.rs
```diff
@@ -265,6 +265,23 @@ impl<'a> DatabaseExt for CowBackend<'a> {
...
+                .backend
+                .inner
+                .impure_cheatcodes
+                .contains(cheatcode_name)
+        {
+            self.backend
+                .to_mut()
+                .inner
+                .impure_cheatcodes
+                .insert(cheatcode_name);
```
Wodann, line 282: Optional: I'm not sure about the performance implications of this branch, but you could reduce this to just an insert to avoid the branch prediction of the `contains` - if you're worried - especially as the `insert` likely has an `if` inside its implementation to determine whether a previous value existed.

Ideal would have been to use the `entry` API as it abstracts these two calls into one, but it's still nightly.

This suggestion is the same as your implementation in `crates/foundry/evm/core/src/backend/mod.rs`

crates/foundry/evm/core/src/backend/mod.rs
```diff
@@ -696,6 +713,24 @@ impl Backend {
         false
     }
 
+    /// Whether the backend instance only executed pure cheatcodes.
+    /// Returns true if no cheatcodes were executed.
```
Wodann, line 717:
```suggestion
    /// Returns true if no impure cheatcodes were executed.
```
?

crates/foundry/evm/core/src/fork/multi.rs
```diff
@@ -122,19 +122,24 @@ impl MultiFork {
...
         );
         let (sender, rx) = oneshot_channel();
         let req = Request::CreateFork(Box::new(fork), sender);
         self.handler
             .clone()
             .try_send(req)
             .map_err(|e| eyre::eyre!("{:?}", e))?;
-        rx.recv()?
+        let (fork_id, shared_backend, env) = rx.recv()??;
+        Ok((fork_id, shared_backend, env, fork_block_number))
```
Wodann, line 142: Why return `fork_block_number` from this function when it's already part of the input? Can't the caller just re-use the input value?

crates/foundry/evm/core/src/backend/mod.rs
```diff
@@ -1076,12 +1111,15 @@ impl DatabaseExt for Backend {
 
     fn create_fork(&mut self, create_fork: CreateFork) -> eyre::Result<LocalForkId> {
         trace!("create fork");
-        let (fork_id, fork, _) = self.forks.create_fork(create_fork)?;
+        let (fork_id, fork, _, fork_block_number) = self.forks.create_fork(create_fork)?;
```
Wodann, line 1114: It seems like `fork_block_number` can be extracted from `CreateFork`

crates/foundry/evm/core/src/backend/mod.rs
```diff
@@ -479,7 +483,7 @@ impl Backend {
         };
 
         if let Some(fork) = fork {
-            let (fork_id, fork, _) = backend
+            let (fork_id, fork, _, fork_block_number) = backend
```
Wodann, line 486: It seems like `fork_block_number` can be extracted from `CreateFork`

Wodann (commented):
At a high-level, the changes look good to me.

This was quite tricky to review, as I am not familiar with the implementation of the cheatcodes. I went through all of the cheatcodes; based on their name and Foundry's cheatcode reference, I tried to determine whether I'd expect them to be pure or impure. When I had doubts, I commented on the PR. When my intuition agreed with your implementation, I took it to mean correctness, but given my unfamiliarity, this is not a guarantee. As such, it might be good for @fvictorio to also have a look - as he's more familiar with the cheatcodes.

I also had a couple of comments w.r.t code implementation.

crates/foundry/evm/core/src/backend/cow.rs
agostbiro: Thanks for the suggestion! The backend here is `Cow`, so the goal of the check is to avoid upgrading to an owned version if we don't actually have to insert.

crates/foundry/evm/core/src/backend/mod.rs
agostbiro: I meant to clarify with this comment that the function returns true in two cases:
- If cheatcodes were executed and only pure cheatcodes were executed.
- If no cheatcodes were executed.

crates/foundry/evm/core/src/fork/multi.rs
agostbiro: Good catch thanks! Fixed in https://github.com/NomicFoundation/edr/pull/822/commits/74a47d17aa564b3169657b43397285f2ad62c31e

crates/foundry/evm/core/src/backend/mod.rs
agostbiro: Fixed in https://github.com/NomicFoundation/edr/commit/74a47d17aa564b3169657b43397285f2ad62c31e

crates/foundry/evm/core/src/backend/mod.rs
agostbiro: Fixed in https://github.com/NomicFoundation/edr/commit/74a47d17aa564b3169657b43397285f2ad62c31e

crates/foundry/evm/core/src/backend/mod.rs
Wodann: Ah, gotcha. Now it makes sense

Wodann (approved):
Approving as all of my comments were answered and resolved.

Maybe @fvictorio can have a review after merging?

agostbiro:
Thanks @Wodann I ended up pushing another commit here instead of creating a new PR to expose indeterministic execution warnings to Hardhat[^1], as I discovered a bug in this PR while testing that out (which is now fixed).

The new commit: [Expose indeterminism warnings](https://github.com/NomicFoundation/edr/pull/822/commits/a202feeb28efdef642ce1df794c63d58a25e02c2)

The design doc for exposing indeterministic execution warnings to Hardhat: https://www.notion.so/nomicfoundation/Indeterministic-Execution-Warnings-in-Solidity-Tests-from-EDR-1a7578cdeaf58041a237d4331c5a1677#1a7578cdeaf58095b0dcef55c61303e4

crates/edr_napi/index.d.ts
```diff
@@ -669,6 +669,52 @@ export interface AddressLabel {
...
+  globalForkLatest: boolean
+  /**
+   * The list of executed impure cheatcode signatures. We collect function
+   * signatures instead of function names as whether a cheatcode is impure
+   * can depend on the arguments it takes (e.g. `createFork` without a second
+   * argument means implicitly fork from “latest”). Example signature:
+   * `function createSelectFork(string calldata urlOrAlias) external returns
+   * (uint256 forkId);`.
+   */
+  impureCheatcodes: Array<string>
```
Wodann, line 716: This deviates from the definition in the product document. The doc specifies `reasons: Array<GlobalForkLatest | ImpureCheatcode>`.

Was this change decided after defining the doc?

Wodann (approved):
This cleans up the API nicely; well done!

I raised one concern where there seemingly is a discrepancy between the design document and the implementation.

As I'm not sure whether this is a result of the design having changed during implementation, I approved and you can decide how to deal with it before merging.

fvictorio (approved):
I only skimmed the list of cheatcodes, but this looks good to me. We can re-visit this later after we have more experience with these features.

crates/edr_napi/index.d.ts
agostbiro: Hey, thanks for asking, yes this was changed. I left a comment about this in the design doc, but didn't update it. The rationale is that it's simpler this way, because there can be only one `GlobalForkLatest`, so there is no need for it to be in the array and then we don't need the sum type `GlobalForkLatest | ImpureCheatcode` and the associated `kind` tags.

PR #825, opened 2025-03-04
fix: optimize cow backend impure cheatcode collection

agostbiro:
Optimize impure cheatcode collection when using `CowBackend`: store the impure cheatcodes on the `CowBackend` instead of the inner `Backend` to avoid having to clone it to record impure cheatcodes.


Wodann (approved):

agostbiro:
Closing as outdated

PR #839, opened 2025-03-23
Normalise rpc debug trace format

### LINKED ISSUE #783: Normalise JSON-RPC format for rpcDebugTrace
Hardhat is currently reformatting the `rpcDebugTrace` call, see: 
https://github.com/NomicFoundation/hardhat/blob/024d72b09c6edefb00c012e9514a0948c255d0ab/v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/utils/convert-to-edr.ts#L176-L212

It seems that EDR JSON-RPC `rpcDebugTrace` calls return a custom format, possibly due to trying to implement https://eips.ethereum.org/EIPS/eip-3155. This is not the JSON-RPC trace format, it’s intended to be used by EVM testing (as in testing the EVM itself) infrastructure.

We should move the normalisation that Hardhat does back into EDR and make sure we work with most external tools/libs/etc. We can follow geth's approach to do so.

AmosAidoo:
## Which issue does this PR close?

- Closes #783 

## Changes

- Moved normalisation that Hardhat does on debug trace into EDR.
- Added new types `RpcDebugTraceResult` and `RpcDebugTraceLogItem` that conform to expected structure.
- Added tests to validate trace structure against expected normalised structure.

## How was this tested?

- Compared trace manually with output from Geth.
- Wrote test to validate normalised structure.
- Ran tests and verified all tests pass including old ones.

### DIFF
--- a/crates/edr_napi/src/debug_trace.rs
+++ b/crates/edr_napi/src/debug_trace.rs
@@ -36,3 +36,35 @@ pub struct DebugTraceLogItem {
     /// Map of all stored values with keys and values encoded as hex strings.
     pub storage: Option<HashMap<String, String>>,
 }
+
+#[napi(object)]
+pub struct RpcDebugTraceResult {
+    pub failed: bool,
+    pub gas: BigInt,
+    pub return_value: String,
+    pub struct_logs: Vec<RpcDebugTraceLogItem>,
+}
+
+#[napi(object)]
+pub struct RpcDebugTraceLogItem {
+    /// Program Counter
+    pub pc: BigInt,
+    /// Name of the operation
+    pub op: String,
+    /// Gas left before executing this operation as hex number.
+    pub gas: BigInt,
+    /// Gas cost of this operation as hex number.
+    pub gas_cost: BigInt,
+    /// Array of all values (hex numbers) on the stack
+    pub stack: Option<Vec<String>>,
+    /// Depth of the call stack
+    pub depth: BigInt,
+    /// Size of memory array
+    pub mem_size: BigInt,
+    /// Description of an error as a hex string.
+    pub error: Option<String>,
+    /// Array of all allocated values as hex strings.
+    pub memory: Option<Vec<String>>,
+    /// Map of all stored values with keys and values encoded as hex strings.
+    pub storage: Option<HashMap<String, String>>,
+}
--- a/crates/edr_napi/src/provider.rs
+++ b/crates/edr_napi/src/provider.rs
@@ -2,7 +2,10 @@ mod config;
 
 use std::sync::Arc;
 
-use edr_provider::{time::CurrentTime, InvalidRequestReason};
+use edr_evm::DebugTraceResult;
+use edr_provider::{
+    time::CurrentTime, InvalidRequestReason, RpcDebugTraceLogItem, RpcDebugTraceResult,
+};
 use edr_rpc_eth::jsonrpc;
 use edr_solidity::contract_decoder::ContractDecoder;
 use napi::{
@@ -102,6 +105,9 @@ impl Provider {
     #[doc = "Handles a JSON-RPC request and returns a JSON-RPC response."]
     #[napi]
     pub async fn handle_request(&self, json_request: String) -> napi::Result<Response> {
+        let method_name = serde_json::from_str::<JsonRpcMethodExtractor>(&json_request)
+            .map_or_else(|_| "".to_string(), |m| m.method);
+
         let provider = self.provider.clone();
         let request = match serde_json::from_str(&json_request) {
             Ok(request) => request,
@@ -204,16 +210,49 @@ impl Provider {
                 }
             })
             .map_err(|error| napi::Error::new(Status::GenericFailure, error.to_string()))
-            .map(|data| {
+            .and_then(|data| {
                 let solidity_trace = solidity_trace.map(|trace| SolidityTraceData {
                     trace,
                     contract_decoder: Arc::clone(&self.contract_decoder),
                 });
-                Response {
+
+                // Normalise JSON-RPC format for rpcDebugTrace
+                let data = if method_name == "debug_traceTransaction"
+                    || method_name == "debug_traceCall"
+                {
+                    let parsed_data: Result<DebugTraceResult, _> = match &data {
+                        Either::A(json_string) => {
+                            serde_json::from_str::<JsonRpcResponse<DebugTraceResult>>(json_string)
+                                .map(|r| r.result)
+                        }
+                        Either::B(json_value) => serde_json::from_value::<
+                            JsonRpcResponse<DebugTraceResult>,
+                        >(json_value.clone())
+                        .map(|r| r.result),
+                    };
+
+                    let trace_result = parsed_data.map_err(|e| {
+                        napi::Error::new(
+                            Status::GenericFailure,
+                            format!("Failed to parse DebugTraceResult: {e}"),
+                        )
+                    })?;
+
+                    let transformed = normalise_rpc_debug_trace(trace_result)
+                        .map_err(|e| napi::Error::new(Status::GenericFailure, e))?;
+                    let transformed = JsonRpcResponse {
+                        result: transformed,
+                    };
+                    Either::B(serde_json::to_value(transformed)?)
+                } else {
+                    data
+                };
+
+                Ok(Response {
                     solidity_trace,
                     data,
                     traces: traces.into_iter().map(Arc::new).collect(),
-                }
+                })
             })
     }
 
@@ -314,6 +353,16 @@ struct SolidityTraceData {
     contract_decoder: Arc<ContractDecoder>,
 }
 
+#[derive(serde::Deserialize, serde::Serialize)]
+struct JsonRpcResponse<T> {
+    result: T,
+}
+
+#[derive(serde::Deserialize)]
+struct JsonRpcMethodExtractor {
+    method: String,
+}
+
 #[napi]
 pub struct Response {
     // N-API is known to be slow when marshalling `serde_json::Value`s, so we try to return a
@@ -373,4 +422,59 @@ impl Response {
             Ok(None)
         }
     }
+}
+
+// Rust port of https://github.com/NomicFoundation/hardhat/blob/024d72b09c6edefb00c012e9514a0948c255d0ab/v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/utils/convert-to-edr.ts#L176
+fn normalise_rpc_debug_trace(trace: DebugTraceResult) -> Result<serde_json::Value, String> {
+    let mut struct_logs = Vec::new();
+
+    for log in trace.logs {
+        let rpc_log = RpcDebugTraceLogItem {
+            pc: log.pc,
+            op: log.op_name,
+            gas: u64::from_str_radix(log.gas.trim_start_matches("0x"), 16).unwrap_or(0),
+            gas_cost: u64::from_str_radix(log.gas_cost.trim_start_matches("0x"), 16).unwrap_or(0),
+            stack: log.stack.map(|values| {
+                values
+                    .into_iter()
+                    .map(|value| value.trim_start_matches("0x").to_string())
+                    .collect()
+            }),
+            depth: log.depth,
+            mem_size: log.mem_size,
+            error: log.error,
+            memory: log.memory,
+            storage: log.storage.map(|storage| {
+                storage
+                    .into_iter()
+                    .map(|(key, value)| {
+                        let stripped_key = key.strip_prefix("0x").unwrap_or(&key).to_string();
+                        let stripped_value = value.strip_prefix("0x").unwrap_or(&value).to_string();
+                        (stripped_key, stripped_value)
+                    })
+                    .collect()
+            }),
+        };
+
+        struct_logs.push(rpc_log);
+    }
+
+    // REVM trace adds initial STOP that Hardhat doesn't expect
+    if !struct_logs.is_empty() && struct_logs[0].op == "STOP" {
+        struct_logs.remove(0);
+    }
+
+    let return_value = trace
+        .output
+        .map(|b| b.to_string().trim_start_matches("0x").to_string())
+        .unwrap();
+
+    let rpc_debug_trace = RpcDebugTraceResult {
+        failed: !trace.pass,
+        gas: trace.gas_used,
+        return_value,
+        struct_logs,
+    };
+
+    serde_json::to_value(rpc_debug_trace).map_err(|e| e.to_string())
 }
--- a/crates/edr_napi/test/provider.ts
+++ b/crates/edr_napi/test/provider.ts
@@ -5,6 +5,8 @@ import {
   ContractAndFunctionName,
   MineOrdering,
   Provider,
+  RpcDebugTraceLogItem,
+  RpcDebugTraceResult,
   SpecId,
   SubscriptionEvent,
 } from "..";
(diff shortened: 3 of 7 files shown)


AmosAidoo:
@nebasuke This is my PR for the issue.
I am available anytime to make the necessary changes.

nebasuke:
Great, thank you @AmosAidoo, we appreciate the effort! 

It will probably be @Wodann or @agostbiro that can take a look.

AmosAidoo:
`cd hardhat-tests && pnpm test:ci`
Noticed I missed these tests. I'm taking a look.

nebasuke:
> Noticed I missed these tests. I'm taking a look.

Great, thanks! I'll re-run CI when needed.

AmosAidoo:
the tests still expect the `pass`, `gasUsed` and `opName` fields so I added them in addition to the new format. Also, the tests still assume the memory and storage strings will have the leading `0x` string.
So I guess this PR will need a follow-up issue after this is integrated into Hardhat. The hardhat tests will then have to be updated afterwards.

AmosAidoo:
> Great, thanks! I'll re-run CI when needed.

You can re-run the CI

AmosAidoo:
Can you kindly re-run

crates/edr_napi/src/provider.rs
```diff
@@ -204,16 +210,48 @@ impl Provider {
...
+                            let transformed = normalise_rpc_debug_trace(trace_result)
+                                .map_err(|e| napi::Error::new(Status::GenericFailure, e))?;
+
+                            let transformed = JsonRpcResponse {
+                                result: transformed,
+                            };
+                            Either::B(serde_json::to_value(transformed)?)
+                        }
+                        Err(_) => data,
+                    }
```
Wodann, line 245: Could you please split this into a function with a descriptive name?

crates/edr_napi/src/provider.rs
```diff
@@ -204,16 +210,48 @@ impl Provider {
...
+                    let parsed_data: Result<DebugTraceResult, _> = match &data {
+                        Either::A(json_string) => {
+                            serde_json::from_str::<JsonRpcResponse<DebugTraceResult>>(json_string)
+                                .map(|r| r.result)
+                        }
+                        Either::B(json_value) => serde_json::from_value::<
+                            JsonRpcResponse<DebugTraceResult>,
+                        >(json_value.clone())
+                        .map(|r| r.result),
+                    };
```
Wodann, line 232: It is incredibly inefficient to:

1. serialize the result to JSON (string or object)
2. deserialize the result from JSON (string or object)
3. normalize result
4. serialize the result

Could you please investigate making this change one level up, before we first serialize the result? That would be in `edr_provider`

AmosAidoo:
> Thank you for taking a stab at this. I left some requests to try and move the changes into `edr_provider` as the current approach goes through multiple steps of serialization and deserialization.

Perfect, thank you for the review. I'll make the changes.

crates/edr_napi/src/provider.rs
AmosAidoo: I have moved the normalisation into `edr_provider`

crates/edr_napi/src/provider.rs
AmosAidoo: No need for these lines now after moving normalisation to edr_provider`.

crates/edr_napi/src/debug_trace.rs
```diff
@@ -36,3 +36,37 @@ pub struct DebugTraceLogItem {
...
+    pub failed: bool,
+    pub gas: BigInt,
+    pub pass: bool,
+    pub gas_used: BigInt,
+    pub return_value: String,
+    pub struct_logs: Vec<RpcDebugTraceLogItem>,
+}
+
+#[napi(object)]
+pub struct RpcDebugTraceLogItem {
```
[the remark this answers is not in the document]
AmosAidoo: Done

crates/edr_napi/src/debug_trace.rs
```diff
@@ -36,3 +36,37 @@ pub struct DebugTraceLogItem {
...
+
+#[napi(object)]
+pub struct RpcDebugTraceResult {
+    pub failed: bool,
+    pub gas: BigInt,
+    pub pass: bool,
+    pub gas_used: BigInt,
+    pub return_value: String,
+    pub struct_logs: Vec<RpcDebugTraceLogItem>,
+}
```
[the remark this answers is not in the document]
AmosAidoo: Done

crates/edr_napi/src/provider.rs
```diff
@@ -102,6 +105,9 @@ impl Provider {
     #[doc = "Handles a JSON-RPC request and returns a JSON-RPC response."]
     #[napi]
     pub async fn handle_request(&self, json_request: String) -> napi::Result<Response> {
+        let method_name = serde_json::from_str::<JsonRpcMethodExtractor>(&json_request)
+            .map_or_else(|_| "".to_string(), |m| m.method);
```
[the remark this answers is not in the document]
AmosAidoo: Removed this as well

AmosAidoo:
Seems to be looking good so far regarding the tests. Looks like the remaining failures are because of environmental variables not set as I am an external contributor.

crates/edr_napi/src/debug_trace.rs
```diff
@@ -36,3 +36,54 @@ pub struct DebugTraceLogItem {
...
     pub storage: Option<HashMap<String, String>>,
 }
+
+/// Result of `debug_traceTransaction` and `debug_traceCall` after
+/// normalisation. `pass` and `gas_used` exist before normalisation. They will
+/// be replaced by `failed` and `gas` respectively. They currently exist
+/// together because Hardhat still depends on them but `pass` and `gas_used`
+/// should likely to be removed after a while.
+#[napi(object)]
+pub struct RpcDebugTraceResult {
```
Wodann, line 46: As this type is no longer part of the API, do we really need to expose it?

I assume Hardhat has its own definition, given that it's a standard.

cc: @fvictorio

crates/edr_napi/src/debug_trace.rs
```diff
@@ -36,3 +36,54 @@ pub struct DebugTraceLogItem {
...
+    /// Return values of the function.
+    pub return_value: String,
+    /// Debug logs after normalisation
+    pub struct_logs: Vec<RpcDebugTraceLogItem>,
+}
+
+/// Debug logs after normalising the EIP-3155 debug logs.
+/// This is the format Hardhat expects
+#[napi(object)]
+pub struct RpcDebugTraceLogItem {
```
Wodann, line 68: As this type is no longer part of the API, do we really need to expose it?

I assume Hardhat has its own definition, given that it's a standard.

cc: @fvictorio

crates/edr_napi/test/provider.ts
```diff
@@ -415,3 +515,48 @@ function assertEqualMemory(stepMemory: Buffer | undefined, expected: Buffer) {
 
   assert.isTrue(stepMemory.equals(expected));
 }
+
+function assertJsonRpcFormatNormalised(trace: RpcDebugTraceResult) {
```
Wodann, line 519: I think these assertions can happen without knowing the type `RpcDebugTraceResult`, but am not 100%. If so, we can just remove the type from EDR's N-API, reducing our API surface.

crates/edr_napi/test/provider.ts
```diff
@@ -415,3 +515,48 @@ function assertEqualMemory(stepMemory: Buffer | undefined, expected: Buffer) {
...
+      });
+    }
+
+    if ("storage" in log) {
+      assert.isObject(log.storage);
+      Object.entries(log.storage!).forEach(([key, value]) => {
+        assert.isString(key);
+        assert.isString(value);
+        // assert.isFalse(key.startsWith("0x"));
+        // assert.isFalse(value.startsWith("0x"));
```
Wodann, line 554: Why are these two disabled?

crates/edr_provider/src/lib.rs
```diff
@@ -380,10 +382,12 @@ impl<LoggerErrorT: Debug + Send + Sync + 'static, TimerT: Clone + TimeSinceEpoch
             // debug_* methods
             MethodInvocation::DebugTraceTransaction(transaction_hash, config) => {
                 debug::handle_debug_trace_transaction(data, transaction_hash, config)
+                    .and_then(normalise_rpc_debug_trace)
```
Wodann, line 385: Please move this call inside the `handle_debug_trace_transaction` call, following the same pattern as all other functions. The `handle_*` functions determine the result type, not the provider.

crates/edr_provider/src/lib.rs
```diff
@@ -514,3 +518,70 @@ fn to_json_with_traces<T: serde::Serialize, LoggerErrorT: Debug>(
         traces: value.1,
     })
 }
+
+// Rust port of https://github.com/NomicFoundation/hardhat/blob/024d72b09c6edefb00c012e9514a0948c255d0ab/v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/utils/convert-to-edr.ts#L176
+/// This normalization is done because this is the format Hardhat expects
+fn normalise_rpc_debug_trace<LoggerErrorT: Debug>(
```
Wodann, line 524: Please move this to the same file as the `handle_` function I mentioned in my other comment.

crates/edr_provider/src/lib.rs
```diff
@@ -514,3 +518,70 @@ fn to_json_with_traces<T: serde::Serialize, LoggerErrorT: Debug>(
...
+            pc: log.pc,
+            op: log.op_name.clone(),
+            op_name: log.op_name,
+            gas: u64::from_str_radix(log.gas.trim_start_matches("0x"), 16).unwrap_or(0),
+            gas_cost: u64::from_str_radix(log.gas_cost.trim_start_matches("0x"), 16).unwrap_or(0),
+            stack: log.stack.map(|values| {
+                values
+                    .into_iter()
+                    // Removing this trim temporarily as the Hardhat test assumes 0x is there
+                    // .map(|value| value.trim_start_matches("0x").to_string())
```
Wodann, line 542: If Hardhat requires the `0x`, then why do you believe we should trim the `0x`?

crates/edr_provider/src/lib.rs
```diff
@@ -514,3 +518,70 @@ fn to_json_with_traces<T: serde::Serialize, LoggerErrorT: Debug>(
...
+                    .collect()
+            }),
+            depth: log.depth,
+            mem_size: log.mem_size,
+            error: log.error,
+            memory: log.memory,
+            storage: log.storage.map(|storage| {
+                storage
+                    .into_iter()
+                    // Removing this trim temporarily as the Hardhat test assumes 0x is there
```
Wodann, line 552: If Hardhat requires the `0x`, then why do you believe we should trim the `0x`?

crates/edr_provider/src/requests/debug.rs
```diff
@@ -120,3 +121,48 @@ impl From<DebugTraceConfig> for edr_evm::DebugTraceConfig {
         }
     }
 }
+
+/// This is the JSON-RPC Debug trace format
+#[derive(Debug, Clone, serde::Deserialize, serde::Serialize)]
```
Wodann, line 126: Do we need both `Deserialize` and `Serialize`? Based on the code, it seems that `serde::Serialize` is sufficient.

crates/edr_provider/src/requests/debug.rs
```diff
@@ -120,3 +121,48 @@ impl From<DebugTraceConfig> for edr_evm::DebugTraceConfig {
...
+#[serde(rename_all = "camelCase")]
+pub struct RpcDebugTraceResult {
+    pub failed: bool,
+    pub gas: u64,
```
Wodann, line 130: Did the old format require a `0x...` representation? What made you decide to go with normal `u64` serialization?

Do you have a reference to support this choice?

If we should support a hex representation of a quantity, you should use `alloy_serde::quantity`.

This applies to all `u64` usages here.

Wodann (changes requested):
This is going in the right direction, but quite a few things are still unclear to me about the chosen solution. I'd appreciate if you could clarify.

crates/edr_napi/test/provider.ts
AmosAidoo: I disabled these two because at this moment, the normalisation is still being done in Hardhat so it is still correct to have `0x` at the start of the strings.
Here is the reference in Hardhat: https://github.com/NomicFoundation/hardhat/blob/024d72b09c6edefb00c012e9514a0948c255d0ab/v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/utils/convert-to-edr.ts#L197-L203

The Hardhat tests also operate on this assumption so I believe these two asserts should be disabled until the normalisation is completely removed from Hardhat.

crates/edr_provider/src/lib.rs
AmosAidoo: Hardhat doesn't require the `0x` but assumes they are there. Hardhat later on trims the `0x` when normalising. Here is the relevant code:
https://github.com/NomicFoundation/hardhat/blob/024d72b09c6edefb00c012e9514a0948c255d0ab/v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/utils/convert-to-edr.ts#L192-L195

When running the `hardhart-tests` for example, I confirmed this expected format from the trace generated from Geth which the tests depend on: 
- https://github.com/NomicFoundation/edr/blob/main/hardhat-tests/test/fixture-debug-traces/traceCall/contractAGetMessage.ts
- https://github.com/NomicFoundation/edr/blob/main/hardhat-tests/test/fixture-debug-traces/traceCall/contractAGetMessageDebugConfig.ts

crates/edr_provider/src/lib.rs
AmosAidoo: Same as the reasons stated for the two related comments above. Hardhat doesn't require `0x` but assumes it is there and so it get removed during normalisation.

crates/edr_provider/src/requests/debug.rs
AmosAidoo: Yes you are right, thanks for pointing it out. I'll remove it.

crates/edr_provider/src/requests/debug.rs
AmosAidoo: Yes so I used `u64` because `gas_used` in `DebugTraceResult` is also `u64`:
https://github.com/NomicFoundation/edr/blob/main/crates/edr_evm/src/debug_trace.rs#L220

AmosAidoo: Noted the point on `alloy_serde::quantity` for hex quantities.

crates/edr_napi/src/debug_trace.rs
AmosAidoo: You are right, Hardhat has its own definition

AmosAidoo:
I have made some changes and replied to the questions

crates/edr_napi/test/provider.ts
AmosAidoo: Ah perfect. Thank you

Wodann:
I added some information to the [original issue](https://github.com/NomicFoundation/edr/issues/783) on how to create test vectors for your code. You can essentially use geth to create some expected output formats and do string comparison between your generated output and the expected output.

AmosAidoo:
Noted. Thank you

AmosAidoo:
> I added some information to the [original issue](https://github.com/NomicFoundation/edr/issues/783) on how to create test vectors for your code. You can essentially use geth to create some expected output formats and do string comparison between your generated output and the expected output.
> 
> > Most of the info about `debug_traceCall` is [here](https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-debug#debugtracecall), but note that they use a js lib, not the raw rpc, in the snippets.
> > To be able to port it with confidence, we should create a few test cases using geth, and making sure we match their format.

@Wodann is it sufficient to use geth in dev mode to generate the test cases?

Wodann:
> > I added some information to the [original issue](https://github.com/NomicFoundation/edr/issues/783) on how to create test vectors for your code. You can essentially use geth to create some expected output formats and do string comparison between your generated output and the expected output.
> > > Most of the info about `debug_traceCall` is [here](https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-debug#debugtracecall), but note that they use a js lib, not the raw rpc, in the snippets.
> > > To be able to port it with confidence, we should create a few test cases using geth, and making sure we match their format.
> 
> @Wodann is it sufficient to use geth in dev mode to generate the test cases?

I don't see why not; as long as the generated output conforms to the JSON-RPC spec.

PR #841, opened 2025-03-26
feat: error for unsupported cheatcodes

agostbiro:
Adds better error messages when invoking unsupported cheatcodes. Previously we'd just return `"unknown selector 0xafc98040"`, now we return `"cheatcode 'broadcast()' is not supported"` instead.

In order to know if an unknown selector refers to an unsupported cheatcode, we have to have access to the latest cheatcode abi from the Foundry repo. I solved this by including the `foundry-cheatcodes-spec` package from the Foundry repo as a git dependency. This'll not work if we want to publish the `edr_solidity_tests` package on crates.io, but I think it's fine until then. The alternative would be to copy over the cheatcode abi from the Foundry repo, but that's more difficult to keep up-to-date.

No changeset needed, as we don't use changesets for the `feat/solidity-tests` branch.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2663,7 +2663,8 @@ dependencies = [
  "edr_common",
  "edr_defaults",
  "eyre",
- "foundry-cheatcodes-spec",
+ "foundry-cheatcodes-spec 0.3.8",
+ "foundry-cheatcodes-spec 1.0.0",
  "foundry-compilers",
  "foundry-evm-core",
  "itertools 0.12.1",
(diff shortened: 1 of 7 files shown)


Wodann (approved):
LGTM!

One alternative that comes to mind (no need to address in this PR) is to have a build script that reads the upstream cheatcodes and generates an `rs` file with all of them listed. This would mean that we're only as up-to-date with Foundry's cheatcodes as our latest commit (as we can run a check in CI, similar to our JS bindings), but does mean we can publish the crate in the future.

PR #845, opened 2025-03-27
fix: napi converts () to undefined so use Either type for reason

### LINKED ISSUE #796: Null instead of undefined in `TestResult`
**Steps to reproduce**

1. Check out the `fix/null-instead-of-undefined` branch
2. Navigate to `js/integration-tests/solidity-tests` in the repo
3. Run `pnpm i`
4. Run `pnpm test`

**Expected result**

All tests pass.

**Actual result**

Many tests fail with the following message:

```
Error: Expected reason to be undefined for test that didn't fail, instead it is: 'null'
```

This is caused by returning `null` instead of `undefined` as the optional failure reason for succeeding tests.

AmosAidoo:
## Which issue does this PR close?

- Closes #796

## Changes

- Changed the type of `reason` to `Option<Either<String, ()>>`

## How was this tested?

- By existing tests. All tests pass

### DIFF
--- a/crates/edr_napi/index.d.ts
+++ b/crates/edr_napi/index.d.ts
@@ -1094,7 +1094,7 @@ export declare class TestResult {
   /** See [edr_solidity_tests::result::TestResult::status] */
   readonly status: TestStatus
   /** See [edr_solidity_tests::result::TestResult::reason] */
-  readonly reason?: string
+  readonly reason?: string | undefined
   /** See [edr_solidity_tests::result::TestResult::counterexample] */
   readonly counterexample?: BaseCounterExample | Array<BaseCounterExample>
   /** See [edr_solidity_tests::result::TestResult::decoded_logs] */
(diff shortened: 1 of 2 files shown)


AmosAidoo:
@agostbiro here is the PR

agostbiro:
Thanks @AmosAidoo ! It's a clever idea to use `Either<String, ()>` in order to force `string | undefined` on the JS side (note that the `Option` is not necessary anymore). 

After thinking about this more, I don't think we want to adopt this, because we have a bunch of other structs with optional members that we return to JS, and we'd have to consistently use `Either<T, ()>` in those places instead of `Option<T>` which is error-prone, as `Either<T, ()>` is not idiomatic. Wdyt @Wodann ?

(I rebased the base branch to get the CI running, but you'd have to rebase your branch too @AmosAidoo )

Wodann:
I don't see how this solved the problem, tbh.

The original API already should generate an `undefined`, as indicated by the `?`:

```
  readonly reason?: string
```

This change just makes it `String | undefined | undefined` - implicitly.

If there is an intermittent bug, this would obfuscate it rather than solve it.

AmosAidoo:
Makes sense.
With the limitations of `#[napi(object)]`, another solution I can think of on top of my head would be to write a custom conversion to `JsObject`

agostbiro:
@AmosAidoo thanks for looking into this! Tbh it looks like a napi-rs bug, so ideally we could create a repro for them and fix it there. I'd close this PR if you don't mind

AmosAidoo:
@agostbiro sure no problem

PR #846, opened 2025-03-28
ci: update op workflow to match 'feat/multichain'

Wodann:
We renamed Optimism to OP in `feat/multichain`. As our scheduled CI jobs are run from workflows in `main`, we also need to add this change to the `main` branch.


agostbiro (approved):

Wodann:
Failing tests are due to a known issue in provider responses, unrelated to this PR.

PR #847, opened 2025-03-28
chore: use hardhat v3 for soltest js crates

agostbiro:
- Use Hardhat v3 dependency for Solidity test crates, but keep Hardhat v2 dependency for Hardhat integration testing crates as these are HHv2 specific
- In the benchmark package, use both HH v2 and v3 as we can’t use v3 scenarios before merging the feature branches due to the provider in HH v3 using `@ignored/edr-optimism` (learned this the hard way 😅)
- Bump node version in CI to 22 as Hardhat v3 needs node 22
- Change type of Solidity test JS packages to “module” as Hardhat v3 uses modules
- Change to node test runner in Solidity test JS packages as it plays better with module type packages
- Added helper function to run Hardhat build programmatically, so that we don’t need to prefix commands with `npx hardhat build`
- Remove run against Anvil option as the scenarios are not compatible with Anvil and we're not using this option


agostbiro:
The test failures are due to Alchemy returning wrong data for certain receipts

fvictorio (approved):
I didn't review all the changes in depth, but the description of the PR plus what I skimmed look good to me.

The most controversial change here is to use both HH2 and HH3 in the same repo, but I think it's fine. The best next step there is to replace HH2 with HH3's generic chain type.

agostbiro:
20ms is ok for `StdMathTest` based on historic results, so I'm merging in spite of the benchmark failure: https://nomic-foundation-automation.github.io/edr-benchmark-results/soltests/

agostbiro:
Merged without squashing as the commit history will be useful for upgrading to HH v3 for the scenarios

PR #849, opened 2025-04-01
chore: add serialize for `SolidityTestRunnerConfigArgs` config

agostbiro:
I added this while debugging something and it can be useful in the future as well for debugging as the arguments are too long to print.


agostbiro:
20ms is ok for `StdMathTest` based on historic results:

https://nomic-foundation-automation.github.io/edr-benchmark-results/soltests/

`StdUtilsForkTest` does network requests so occasional failures are expected there, so merging in spite of benchmark failures.

Issue #837, opened 2025-03-19
Remove inlined bug fix for N-API


Wodann:
**Definition of Done**
Remove inlined implementation for `BigInt::get_u128` at locations marked with this issue URL once fix in https://github.com/napi-rs/napi-rs/pull/2524 has been incorporated

PR #855, opened 2025-04-02
build: upgrade to Rust v1.85.1

Wodann:
Upgrades to Rust v1.85, which has support for edition 2024.

Fixes https://github.com/NomicFoundation/edr/issues/837


agostbiro (approved):
Thanks, LGTM

PR #856, opened 2025-04-02
build: upgrade to Rust edition 2024

Wodann:
Upgrades to Rust edition 2024. This is required to upgrade to REVM v21


agostbiro (approved):
Thanks, LGTM

PR #859, opened 2025-04-03
build: upgrade to REVM v21

### LINKED ISSUE #858: Research differences between Base and OP Stack

Wodann:
Fixes #858

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ rpath = true
 
 [workspace.dependencies]
 alloy-eips = { version = "0.13.0", default-features = false }
-alloy-serde = { version = "0.11.1", default-features = false }
+alloy-serde = { version = "0.13.0", default-features = false }
 paste = { version = "1.0.14", default-features = false }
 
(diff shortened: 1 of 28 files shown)


Cargo.toml
```diff
@@ -25,7 +25,7 @@ rpath = true
 rpath = true
 
 [workspace.dependencies]
-alloy-eips = { version = "0.11.1", default-features = false }
+alloy-eips = { version = "0.13.0", default-features = false }
```
agostbiro, line 28: This is a bit risky as Foundry is still on `alloy` 0.12, but I assume it's necessary because of REVM 21. So having `alloy-eips = 0.12` wouldn't work, right?

agostbiro (approved):
Thanks, LGTM with one question

Cargo.toml
Wodann: No, it's unfortunately required 😔

agostbiro: ok thanks for checking!

PR #860, opened 2025-04-04
docs: update getting started and related code

frangio:
Fixes what seem to be outdated instructions and adds a few useful notes in the book, and addresses a test issue I ran into.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -5,6 +5,8 @@ root = true
 charset = utf-8
 end_of_line = lf
 indent_style = space
-indent_size = 2
 insert_final_newline = true
 trim_trailing_whitespace = true
+
+[*.{js,ts}]
+indent_size = 2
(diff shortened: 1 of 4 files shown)


.editorconfig
```diff
@@ -5,6 +5,8 @@ root = true
 charset = utf-8
 end_of_line = lf
 indent_style = space
-indent_size = 2
```
frangio, line 8: This was affecting Rust files.

crates/edr_rpc_client/src/client.rs
frangio, line 1: This is just moving code around. See diff with whitespace hidden.

crates/edr_rpc_client/src/client.rs
```diff
@@ -916,6 +894,28 @@ mod tests {
             }
         }
 
+        #[tokio::test]
+        async fn call_bad_api_key() {
```
frangio, line 898: Moved this test under the `test-remote` flag because it requires an internet connection.

Wodann (approved):
Thanks!

PR #853, opened 2025-04-02
refactor: remove pre-Byzantium receipt

Wodann:
> ⚠️ **BEWARE**
> It might be necessary to clear your local RPC cache, as it might contain old responses.

Removes support for relaying pre-Byzantium receipts in transactions, incl. a test for RLP en-/decoding.

Resolves #854

PR #863, opened 2025-04-07
test: remove failing tests for unsupported receipt types

Wodann:
This PR is intended to fix the CI, to unblock other people. It only backports the removal of failing receipt tests from https://github.com/NomicFoundation/edr/pull/853 (multi-chain branch), as backporting the full fix turned out to take a lot more time - due to code divergence. Given that we'll soon do a full merge of the multi-chain branch that removes support for those receipts, I'd rather avoid doing that work.

Please let me know if you disagree with that line of reasoning.


agostbiro (approved):
Thanks for fix! I'm ok removing that test, but it looks like there is one more failure: `client::tests::alchemy::get_storage_at_future_block`. I'm approving to expedite merging once that's fixed.

Wodann:
> Thanks for fix! I'm ok removing that test, but it looks like there is one more failure: `client::tests::alchemy::get_storage_at_future_block`. I'm approving to expedite merging once that's fixed.

I re-ran the test. The failure seems to have been intermittent.

PR #867, opened 2025-04-11
feat: add configuration for Base chain

### LINKED ISSUE #857: Implement accurate L2 Base support
**Definition of Done**
- [x] Research differences between Base and OP Stack
- [x] Implement a `BaseChainSpec` that implements all traits listed in [this doc](https://github.com/NomicFoundation/edr/blob/feat/multichain/book/src/04_multichain/01_chain_spec.md) in [this branch](https://github.com/NomicFoundation/edr/tree/feat/multichain)
  - Can take inspiration from [edr_op](https://github.com/NomicFoundation/edr/tree/feat/multichain/crates/edr_op)
- [x] Add tests, similar to `edr_op`
- [ ] Expose `BaseChainSpec` to Hardhat 3, in [Optimism release branch](https://github.com/NomicFoundation/edr/tree/release/optimism)
- [ ] Release new version of `@ignored/edr-optimism` that includes Base
- [ ] Update or add missing [documentation](https://github.com/NomicFoundation/edr/blob/feat/multichain/book/src/04_multichain/01_chain_spec.md) regarding multi-chain support, making it accessible for third-party contributors

frangio:
Fixes #857 

Base is an OP Stack chain, so it behaves the same as OP itself, with different parameters. This PR extends `edr_op` with the config necessary to connect to it.

The approach was to add a module `edr_op::chains` with different constants for each of the chains.

The following are pending:
- [x] Review the genesis state in `edr_napi::chains::op`, probably moving it to `edr_op`.
- [x] Consider running more integration tests on Base: `full_block`, `provider`, `rpc`.


crates/edr_op/src/chains.rs
```diff
@@ -0,0 +1,80 @@
...
+use op_revm::OpSpecId;
+
+/// OP Mainnet chain ID
+pub const OP_MAINNET_CHAIN_ID: u64 = 0xa;
+
+/// OP Mainnet chain config
+///
+/// https://github.com/ethereum-optimism/superchain-registry/blob/51804a33655ddb4feeb0ad88960d9a81acdf6e62/superchain/configs/mainnet/op.toml
+pub static OP_MAINNET_CONFIG: LazyLock<ChainConfig<OpSpecId>> = LazyLock::new(|| ChainConfig {
+    name: "op-mainnet".into(),
```
frangio, line 13: Is it okay to change the name of the existing network? (`mainnet` -> `op-mainnet`)

crates/edr_op/src/lib.rs
```diff
@@ -22,17 +22,20 @@ pub use self::spec::OpChainSpec;
...
 use edr_eth::U256;
 pub use op_revm::{OpHaltReason, OpSpecId};
 
-/// OP chain type
+/// OP Stack chain type
 pub const CHAIN_TYPE: &str = "op";
 
 /// OP Mainnet chain ID
-pub const MAINNET_CHAIN_ID: u64 = 0xa;
+pub const MAINNET_CHAIN_ID: u64 = chains::OP_MAINNET_CHAIN_ID;
```
frangio, line 35: Kept for backwards compatibility but I'm not sure it's needed.

Cargo.toml
```diff
@@ -1,18 +1,5 @@
...
-    "crates/edr_provider",
-    "crates/edr_rpc_client",
-    "crates/edr_rpc_eth",
-    "crates/edr_scenarios",
-    "crates/edr_solidity",
-    "crates/edr_test_utils",
-    "crates/edr_utils",
-    "crates/tools",
-]
+members = ["crates/*"]
```
Wodann, line 2: I recall @agostbiro purposefully made this per-crate as some of the folders inside `crates/*` (in the Solidity test runner branch) are not supposed to be included as crates.

crates/edr_op/src/chains.rs
Wodann: Yup. It's in fact more accurate! https://chainlist.org/chain/10

crates/edr_op/src/chains.rs
```diff
@@ -0,0 +1,80 @@
+use std::sync::LazyLock;
```
Wodann, line 1: For this, maybe split it up into:

hardfork
|
- op.rs
- base.rs

Given that these are essentially child-modules of `hardfork.rs`. Having that structure will also be easy if we adopt code generation in the future. One file can be generated per chain.

crates/edr_op/src/chains.rs
```diff
@@ -0,0 +1,80 @@
...
+
+use edr_evm::hardfork::{Activations, ChainConfig, ForkCondition};
+use op_revm::OpSpecId;
+
+/// OP Mainnet chain ID
+pub const OP_MAINNET_CHAIN_ID: u64 = 0xa;
+
+/// OP Mainnet chain config
+///
+/// https://github.com/ethereum-optimism/superchain-registry/blob/51804a33655ddb4feeb0ad88960d9a81acdf6e62/superchain/configs/mainnet/op.toml
```
Wodann, line 11:
```suggestion
/// <https://github.com/ethereum-optimism/superchain-registry/blob/51804a33655ddb4feeb0ad88960d9a81acdf6e62/superchain/configs/mainnet/op.toml>
```

crates/edr_op/src/chains.rs
```diff
@@ -0,0 +1,80 @@
...
+});
+
+/// Base Mainnet chain ID
+pub const BASE_MAINNET_CHAIN_ID: u64 = 8453;
+
+/// Base Mainnet chain config
+///
+/// https://github.com/ethereum-optimism/superchain-registry/blob/51804a33655ddb4feeb0ad88960d9a81acdf6e62/superchain/configs/mainnet/base.toml
+pub static BASE_MAINNET_CONFIG: LazyLock<ChainConfig<OpSpecId>> = LazyLock::new(|| ChainConfig {
+    name: "base-mainnet".into(),
```
Wodann, line 51: This seems to be called just "Base" (https://chainlist.org/chain/8453) but I'm happy to follow the pattern of using `*-mainnet` everywhere.

crates/edr_op/tests/integration/mod.rs
```diff
@@ -1,16 +1,30 @@
+#![cfg(feature = "test-remote")]
```
Wodann, line 1: Not all integration tests are guaranteed to be remote, so this should remain at a more granular level

Wodann (commented):
Overall it's LGTM! :D

Cargo.toml
frangio: If I understand the issue correctly I think we could use:

```toml
members = ["crates/*", "crates/foundry/*"]
exclude = ["crates/foundry"]
```

agostbiro: Unfortunately the issue with wildcards is that if you add a crate in a branch, and then switch branches, the directory may remain if it has ignored files and that breaks cargo.

crates/edr_op/src/chains.rs
frangio: I didn't give this much thought initially. "Base" or "base" may be better.

Where will this name show up?

Wodann: It's used in [this error message](https://github.com/NomicFoundation/edr/blob/feat/multichain/crates/edr_evm/src/blockchain/forked.rs#L62), collected at [this point](https://github.com/NomicFoundation/edr/blob/feat/multichain/crates/edr_evm/src/blockchain/forked.rs#L207).

frangio: If the name is meant for error messages I would go with something like this:
- OP Mainnet
- OP Sepolia
- Base
- Base Sepolia

Existing names use lowercase ("goerli", "kovan", etc.) though, so these could also be "op mainnet", "base sepolia", etc. Let me know if you have a preference.

Wodann: I like your proposal better than the lowercase names. Given that they are in error messages, readability is a priority.

EDR and Hardhat use lowercase for the chain types but those are a different set than the chain names (match Chain ID).

frangio:
We've decided to keep the genesis state the same across all OP Stack chains for now. Ideally we'd customize the predeploys depending on the chain the user is developing for, but at the moment this can't depend on the chain id because it might not yet be known at the time the predeploys are prepared. For future reference on this see the [genesis files in `superchain-registry`](https://github.com/ethereum-optimism/superchain-registry/tree/main/superchain/extra/genesis).

Since Base chain is to the best of our knowedge identical to OP as far as EDR should be concerned, we haven't added Base-specific integration tests, but we will add a [scheduled check](https://github.com/NomicFoundation/edr/blob/main/.github/workflows/test-recent-optimism-block.yml) against the latest block to be sure.

.github/workflows/test-recent-optimism-block.yml
```diff
@@ -7,8 +7,12 @@ on:
 
 jobs:
   test-recent-op-block:
-    name: Test recent OP block
+    name: Test recent OP block (${{ matrix.network }})
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        network: [opt-mainnet, base-mainnet]
```
Wodann, line 15: Nit:
```suggestion
        network: [op-mainnet, base-mainnet]
```

crates/edr_op/src/hardfork/mod.rs
```diff
@@ -0,0 +1,45 @@
+use std::sync::OnceLock;
```
Wodann, line 1: We don't use `mod.rs`, except for one exception in `tests/integration/mod.rs` (due to compiler limitations for that instance). Could you please make this `crates/edr_op/src/hardfork.rs` to follow that naming convention.

Wodann (changes requested):
One nitpick and one request for change. Otherwise LGTM! Thanks for putting this together!

.github/workflows/test-recent-optimism-block.yml
frangio: This has to be `opt-mainnet` in order to get the right Alchemy URL:
https://github.com/NomicFoundation/edr/blob/1504c5c72c2b0300481d8bc1d5d52b4d0b7cdbe7/.github/workflows/test-recent-optimism-block.yml#L32

See this line in `replay-block.rs` that I forgot to remove previously as it's no longer necessary: https://github.com/NomicFoundation/edr/pull/867/commits/1504c5c72c2b0300481d8bc1d5d52b4d0b7cdbe7

Wodann (approved):

PR #873, opened 2025-04-15
misc: improve error message for parsing BuildInfo

Wodann:
Improves the errors reported for JSON deserialization errors to help pinpoint the issue for callers of the API.

See an [example](https://github.com/NomicFoundation/hardhat/actions/runs/14472747805/job/40590793003?pr=6577) of the previously ambiguous error.


crates/edr_napi/test/issues.ts
```diff
@@ -149,4 +151,125 @@ describe("Provider", () => {
...
+          ignoreContracts: false,
+        }
+      );
+
+      assert.fail("Expected an error but none was thrown");
+    } catch (error) {
+      assert.equal(
+        (error as any).message,
+        "Failed to parse build info: EOF while parsing a value at line 1 column 0"
+      );
```
Wodann, line 272: @fvictorio I tried to get this to work with chai's `assert.isRejected` but it would still throw an error the moment that `context.createProvider` was called. The thrown error would also point to the line of `context.createProvider` as causing the thrown error.

That was surprising to me as I'd expect the `Promise` to not be linkable to its original callsite. I'd also expect the error to nicely be bubbled up to when I call `const provider = await providerPromise;`.

Is my expectation of Javascript promises incorrect?

agostbiro (approved):
Thanks, Rust part LGTM

crates/edr_solidity/src/artifacts.rs
```diff
@@ -14,10 +14,10 @@ use serde::{Deserialize, Serialize};
 #[derive(Debug, thiserror::Error)]
 pub enum BuildInfoConfigError {
     /// JSON deserialization error
-    #[error(transparent)]
+    #[error("Failed to parse build info: {0}")]
```
alcuadrado, line 17: Out of curiosity: Can't a linter rule force you to add these messages?

crates/edr_napi/test/issues.ts
alcuadrado: chai is notoriously bad, so it may be its fault. I you share the code, I can take a look.

crates/edr_solidity/src/artifacts.rs
agostbiro: I think it's difficult, because sometimes you really just want to propagate the error message of the wrapped error.

crates/edr_napi/test/issues.ts
frangio: It sounds and looks like the error is happening synchronously. So even though `createContext` returns a `Promise` the error is happening before the call to `spawn_blocking`.

PR #877, opened 2025-04-17
chore: backport fuzz & invariant test improvements from Foundry

agostbiro:
Backport fuzz & invariant test improvements from Foundry (commit: a2ecefce6).

Also adds relevant new tests from Foundry except for those that require `forge` binary.


crates/edr_solidity_tests/tests/it/invariant.rs
```diff
@@ -778,3 +790,177 @@ async fn test_invariant_scrape_values() {
...
+    //         "default/fuzz/invariant/common/InvariantRollFork.t.sol:
+    // InvariantRollForkBlockTest",         vec![(
+    //             "invariant_fork_handler_block()",
+    //             false,
+    //             Some("revert: too many blocks mined".into()),
+    //             None,
+    //             None,
+    //         )],
+    //     )]),
+    // );
```
Wodann, line 829: Was this directly copied when applying patches or is this leftover from fixing something locally?

js/integration-tests/solidity-tests/package.json
```diff
@@ -23,7 +23,7 @@
     "build:dev": "tsc --build --incremental .",
     "prespecificTest": "cd ../../.. && pnpm build:dev",
     "pretest": "cd ../../.. && pnpm build:dev",
-    "specificTest": "node --import tsx/esm --test-name-pattern=\"Latest fork cheatcode\" test/unit.ts",
+    "specificTest": "node --test-concurrency=1 --import tsx/esm --test-name-pattern=\"ImpureFuzz\" test/fuzz.ts",
```
Wodann, line 26: What's the reason for setting the test-concurrency to 1?

Wodann (approved):
Given that these are backported changes, I reviewed them by skimming through the changes. I only noticed two things that stood out to me, so I left questions there.

Otherwise, LGTM!

Wodann:
The failing tests in the "Test EDR" job would likely be fixed if you backported this change from the `hh2` branch: https://github.com/NomicFoundation/edr/commit/5406076ff3380837106ba1ea886d899e03dfaa92

crates/edr_solidity_tests/tests/it/invariant.rs
agostbiro: Good catch thanks! This was commented out while debugging something, but I added it back now in 27c8af376dd071e161166aad36686a818e652b40

agostbiro:
> The failing tests in the "Test EDR" job would likely be fixed if you backported this change from the `hh2` branch: [5406076](https://github.com/NomicFoundation/edr/commit/5406076ff3380837106ba1ea886d899e03dfaa92)

Thanks, since we're very close to merging the branches I won't do that now.

PR #878, opened 2025-04-22
chore: improve linking error message

fvictorio:
While this is technically user-facing and produces a change in an output, I don't think it needs a changeset because it's very small and it only happens in a corner case (that we are going to fix anyway). So it's fine for this to be released whenever there's a release.


agostbiro (approved):
Thanks, LGTM

PR #879, opened 2025-04-23
feat: add statement coverage instrumentation

Wodann:
This PR adds an API for instrumenting statement code coverage, based on [this research spike](https://github.com/NomicFoundation/hardhat-coverage-spike/blob/main/src/lib.rs).

Should be merged after #866!

### DIFF
--- a/crates/edr_instrument/src/coverage.rs
+++ b/crates/edr_instrument/src/coverage.rs
@@ -1,0 +1,209 @@
+use edr_eth::B256;
+pub use semver::Version;
+use serde::Serialize;
+use sha3::{Digest, Keccak256};
+use slang_solidity::{
+    cst::{Node, NonterminalKind},
+    parser::{ParseError, Parser as SolidityParser, ParserInitializationError},
+};
+
+#[derive(Serialize)]
+pub struct InstrumentationMetadata {
+    pub tag: B256,
+    pub kind: &'static str,
+    pub start_utf16: usize,
+    pub end_utf16: usize,
+}
+
+pub struct InstrumentationResult {
+    pub source: String,
+    pub metadata: Vec<InstrumentationMetadata>,
+}
+
+/// Computes a deterministic hash for a statement in a Solidity file.
+///
+/// For the time being, we're assuming that compilation configuration has no
+/// impact on the statement hash. This is a simplification, but is considered a
+/// reasonable trade-off for the current use case.
+fn compute_deterministic_hash_for_statement(
+    source_id: &str,
+    content_hash: &B256,
+    statement_counter: u64,
+) -> B256 {
+    let mut hasher = Keccak256::new();
+
+    hasher.update(source_id);
+    hasher.update(content_hash);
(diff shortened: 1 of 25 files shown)


crates/edr_instrument/src/coverage.rs
```diff
@@ -0,0 +1,212 @@
...
+///
+/// For the time being, we're assuming that compilation configuration has no
+/// impact on the statement hash. This is a simplification, but is considered a
+/// reasonable trade-off for the current use case.
+fn compute_deterministic_hash_for_statement(
+    source_id: &str,
+    content_hash: &B256,
+    statement_counter: u64,
+) -> B256 {
+    let mut hasher = Keccak256::new();
```
agostbiro, line 33: I'm curious about the choice of the hash function. Is the purpose just to generate unique ids deterministically? If yes, we could consider using Blake3 instead of Keccak, as it's an order of magnitude faster: https://github.com/BLAKE3-team/BLAKE3?tab=readme-ov-file#blake3

crates/edr_instrument/src/coverage.rs
```diff
@@ -0,0 +1,212 @@
...
+
+    instrumented_source.push_str("\nimport \"hardhat/coverage.sol\";");
+
+    Ok(InstrumentationResult {
+        source: instrumented_source,
+        metadata,
+    })
+
+    // serde_json::to_string_pretty(&metadata)
+    //.context("Failed to serialize metadata")?
```
agostbiro, line 108: Leftover?

crates/edr_instrument/src/lib.rs
```diff
@@ -0,0 +1,2 @@
+/// Types for instrumenting code for the purpose of code coverage.
+pub mod coverage;
```
agostbiro, line 2: Could potentially live in `edr_solidity`

agostbiro (approved):
Thanks LGTM with some comments, but none of them blockers.

crates/edr_instrument/src/coverage.rs
Wodann: Good catch. Addressed in https://github.com/NomicFoundation/edr/pull/879/commits/86e6e12ae68fc8fcedd45b0b6f7e7e41dba5eca7

crates/edr_instrument/src/lib.rs
Wodann: I moved this to a standalone crate as the idea is that this can be shipped in its own NPM package in the future. For the time being - as a shortcut - we're including it in one large NPM package with all other parts of EDR.

I considered including it in `edr_solidity` but it would require `slang` to be used as a dependency for that, even when we never would require it for compilation - in the future.

crates/edr_instrument/src/coverage.rs
Wodann: I started a thread to discuss this internally. If we agree to move forward with Blake3 (an idea I'm all for based on your description) I'll make that change in a follow-up PR such that the Hardhat team can start working with the API earlier.

PR #882, opened 2025-04-23
fix: properly defer results of a promise

Wodann:
Fixes the issue raised [here](https://github.com/NomicFoundation/edr/pull/873#discussion_r2046953753) by only returning errors when the promise is resolved.


agostbiro (approved):
Thanks, LGTM

Wodann:
> Hmm on a second look, there are some Hardhat 2 integration tests failing that we should fix before merging I think

I'm looking into this. It has to do with my changes, but I'm still uncertain why.

crates/edr_napi/src/context.rs
frangio, line 1: I wanted to see if I could come up with a nicer way to write this since the `match` statements introduce a lot of noise. We may not want to do this but I'll share what I thought of:

```rust
        let (deferred, promise) = env.create_deferred()?;
        let deferred = &std::cell::Cell::new(Some(deferred));

        let try_create_provider = || -> Result<(), napi::Error> {
            let provider_config = edr_napi_core::provider::Config::try_from(provider_config)?;

            let logger_config = logger_config.resolve(&env)?; // <-- keep using ? operator

            // ...

            let deferred = deferred.take().unwrap();

            runtime.clone().spawn_blocking(move || {
                let result = builder.build(runtime.clone()).map(|provider| {
                    Provider::new(
                        provider,
                        runtime,
                        contract_decoder,
                        #[cfg(feature = "scenarios")]
                        scenario_file,
                    )
                });

                deferred.resolve(|_env| result);
            });

            Ok(())
        };

        try_create_provider().unwrap_or_else(|error| {
            deferred.take().unwrap().reject(error);
        });

        Ok(promise)
```

Wodann: We could introduce a macro (for which `?` is syntactic sugar) or extract all of the conversions into a local function - resulting in a single match statement. That way we're remaining fully statically typed, as opposed to `Cell<Option>`'s  runtime checks.

Wodann:
> Hmm on a second look, there are some Hardhat 2 integration tests failing that we should fix before merging I think

The issue was resolved in https://github.com/NomicFoundation/edr/pull/882/commits/24b0d94391658917a2710e89d820f7a63d1d6bf7

crates/edr_napi/src/logger.rs
```diff
@@ -101,8 +101,22 @@ impl LoggerConfig {
         print_line_callback.unref(env)?;
 
         let print_line_fn = Arc::new(move |message, replace| {
-            let status =
-                print_line_callback.call((message, replace), ThreadsafeFunctionCallMode::Blocking);
+            let (sender, receiver) = channel();
```
frangio, line 104: Should we consider using a `tokio::oneshot::channel` here? Not sure how significant a difference it may be.

frangio (approved):

crates/edr_napi/src/context.rs
Wodann: Addressed by https://github.com/NomicFoundation/edr/pull/882/commits/fc8e3f1d2d5c8b085c821d39117af6aec32ab233

crates/edr_napi/src/logger.rs
Wodann: Tokio's version would require `async` and this is executed in a sync environment.

Wodann:
I force pushed to create a clean commit message

PR #884, opened 2025-04-24
fix: support different solc versions for libs and contracts

agostbiro:
Updates the Solidity test logic that finds which library to link so that it:
- Finds all the matching artifacts. Most of the time there will be only one.
- If there's only one matching artifact, link that.
- If there's more than one, use the one that has the same version as the contract being linked. If there isn't, use the latest one.


js/integration-tests/solidity-tests/test-contracts/UnsupportedCheatcode.t.sol
agostbiro, line 1: Unrelated change, I just noticed the source file was in the wrong place

Wodann (approved):

agostbiro:
Test failures are expected due to Alchemy issues and the benchmark regression is due to `StdUtilsForkTest` and `StdMathTest` having recorded outliers: https://nomic-foundation-automation.github.io/edr-benchmark-results/soltests/. The

PR #885, opened 2025-04-24
chore: mask workflow variable

frangio:
[ref](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#masking-a-value-in-a-log)


frangio:
>We would want something similar for the GH action workflow for L1 mainnet.

The masking happens automatically there, but because we manipulate the URL here it doesn't get the same treatment.

PR #891, opened 2025-04-26
fix: ensure coverage collected callback completes

Wodann:
This is a similar fix as https://github.com/NomicFoundation/edr/commit/24b0d94391658917a2710e89d820f7a63d1d6bf7 in a previous PR where we were unexpectedly performing async callbacks.


agostbiro (approved):
Thanks, LGTM

PR #893, opened 2025-04-28
doc: add code coverage documentation to API

Wodann:
This PR adds detailed documentation for code coverage and instrumentation.

### DIFF
--- a/crates/edr_napi/src/instrument.rs
+++ b/crates/edr_napi/src/instrument.rs
@@ -4,8 +4,10 @@ use napi_derive::napi;
 
 #[napi(object)]
 pub struct InstrumentationResult {
+    /// The generated source code with coverage instrumentation.
     #[napi(readonly)]
     pub source: String,
+    /// The metadata for each instrumented code segment.
     #[napi(readonly)]
     pub metadata: Vec<InstrumentationMetadata>,
 }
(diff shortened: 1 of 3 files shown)


crates/edr_napi/src/instrument.rs
```diff
@@ -4,8 +4,10 @@ use napi_derive::napi;
 
 #[napi(object)]
 pub struct InstrumentationResult {
+    /// The generated source code with coverage instrumentation.
     #[napi(readonly)]
     pub source: String,
+    /// The metadata for each piece of instrumented code.
```
frangio, line 10: Nit: This threw me off a little because I saw `source` as a single "piece of instrumented code". This sounds a little clearer to me:

```suggestion
    /// The metadata for each instrumented code segment.
```

crates/edr_napi/src/instrument.rs
```diff
@@ -31,12 +33,21 @@ impl TryFrom<edr_instrument::coverage::InstrumentationResult> for Instrumentatio
 
 #[napi(object)]
 pub struct InstrumentationMetadata {
+    /// The tag that identifies the instrumented code. The tag is
```
frangio, line 36: Some potential for confusion here that there would be one tag per (source code, source id, version) instead of many.

```suggestion
    /// The tag that identifies the instrumented code. Tags are
```

frangio (approved):

PR #895, opened 2025-04-29
chore: fix ignored profile warning

frangio:
This fixes the following warning:

```
warning: profiles for the non root package will be ignored, specify profiles at the workspace root:
package:   edr/crates/edr_napi/Cargo.toml
workspace: edr/Cargo.toml
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,6 +26,7 @@ edition = "2024"
 rpath = true
 
 [profile.release]
+lto = true
 rpath = true
 
 [workspace.dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,6 @@ edition = "2024"
 rpath = true
 
 [profile.release]
-lto = true
 rpath = true
 
 [workspace.dependencies]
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -26,6 +26,7 @@ edition = "2024"
 rpath = true
 
 [profile.release]
+lto = true
```
Wodann, line 29: This enables LTO for all binaries in the workspace. I'd recommend doing the [following](https://doc.rust-lang.org/cargo/reference/profiles.html#custom-profiles):

1. Create a separate profile:
```toml
[profile.release-lto]
inherits = "release"
lto = true
````
2. use `release-lto` profile in `edr_napi/package.json`: `cargo build --profile release-lto` instead of `--release`

I looked at overriding lto for a single package (from the workspace Cargo.toml) but [it's not supported](https://doc.rust-lang.org/cargo/reference/profiles.html#overrides):

> Overrides cannot specify the `panic`, `lto`, or `rpath` settings.

Wodann (changes requested):

agostbiro:
Hey, thanks for the PR!

In `feat/solidity-tests` we have 

https://github.com/NomicFoundation/edr/blob/a7378a2faf03e7266e01381f74db7d68f80a0b83/Cargo.toml#L41-L47

to match Foundry's build profile and I've already removed `lto` from `crates/edr_napi/Cargo.toml`, so in the `main` branch I'd just leave as is. And then merging `feat/solidity-tests` will take care of this 🙏

frangio:
Closing since will be fixed in after merging `feat/solidity-tests`.

Issue #887, opened 2025-04-25
Update EIP-1559 params for OP Stack networks post-Holocene


frangio:
After adding the basic config for Base chain to EDR following the parameters documented in the Superchain registry, we found issues replaying recent Base blocks. We realized then that the [EIP-1559 base fee parameters](https://github.com/ethereum-optimism/superchain-registry/blob/38c054f1d4179252c91e856fe7115dcbe28962f0/superchain/configs/mainnet/base.toml#L25-L27) were not up to date in these docs (https://github.com/ethereum-optimism/superchain-registry/issues/993).

We later found that the implementation of base fee params used in `edr_op` no longer matches the latest OP release. While EDR allows them to [change with hard forks](https://github.com/NomicFoundation/edr/blob/4fd0dcfca839298000a3f5c2c392f039fd083793/crates/edr_eth/src/eips/eip1559.rs#L23-L25) (as they had changed in OP’s Canyon hard fork), the more recent Holocene hard fork (Jan 2025) [has made them configurable](https://specs.optimism.io/protocol/holocene/system-config.html) at runtime via an L1 `SystemConfig` contract that is owned and modifiable by the chain operator.

For example, Base changed their elasticity multiplier from 6 to 2 in this transaction: https://etherscan.io/tx/0x765a2eb3c7eecea5722b120037123eaec9e6ef4b6a53ba2bcfb88ef08fae074b

Since this potentially requires a more significant rethink of how we handle base fee params we’ve now put this aside for reprioritization.

We may also want to discuss and consider whether it is important to accurately match these parameters to the real chain, since they just affect how the base fee reacts to the usage/congestion of the chain and that’s never going to match the real chain anyway (?).

We have to decide what values we put in for OP Stack chains, whether we should attempt to match the historical parameters of the chain, and how forking is affected.

We probably want to implement the [new Holocene behavior](https://specs.optimism.io/protocol/holocene/exec-engine.html) where the `extraData` in a block encodes the current EIP-1559 base fee parameters, since this is observable and could be used.

Wodann:
I recommend splitting this into a separate function, so we can address it before the base fee parametrisation issue:

> We probably want to implement the [new Holocene behavior](https://specs.optimism.io/protocol/holocene/exec-engine.html) where the extraData in a block encodes the current EIP-1559 base fee parameters, since this is observable and could be used.

PR #896, opened 2025-05-01
fix: elasticity multiplier for Base network

frangio:
Fixes https://github.com/NomicFoundation/edr/issues/887

Base fee parameters were a constant, this PR turns them into a function of chain id. This also prepares things for the ability to provide overrides for the parameters via config like with hardfork activations.

- In `calculate_next_base_fee_per_gas` and `PartialHeader::new` they are now passed as an argument rather than taken from `ChainSpecT`.
- Added `Blockchain::base_fee_params`.
- In `LocalBlockchain::new` we get the constant params for the hardfork from `ChainSpecT` and store them in the struct.
- In `ForkedBlockchain::new` we get the hardfork-variable params for the remote chain id and store them in the struct.

### DIFF
--- a/crates/edr_eth/src/eips/eip1559.rs
+++ b/crates/edr_eth/src/eips/eip1559.rs
@@ -2,7 +2,7 @@ pub use alloy_eips::eip1559::BaseFeeParams as ConstantBaseFeeParams;
 
 /// A mapping of hardfork to [`ConstantBaseFeeParams`]. This is used to specify
 /// dynamic EIP-1559 parameters for chains like OP.
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub struct ForkBaseFeeParams<HardforkT: 'static> {
     activations: &'static [(HardforkT, ConstantBaseFeeParams)],
 }
@@ -16,6 +16,7 @@ impl<HardforkT> ForkBaseFeeParams<HardforkT> {
 
 /// Type that allows specifying constant or dynamic EIP-1559 parameters based on
 /// the active hardfork.
+#[derive(Clone, Copy)]
 pub enum BaseFeeParams<HardforkT: 'static> {
     /// Constant [`ConstantBaseFeeParams`]; used for chains that don't have
     /// dynamic EIP-1559 parameters
--- a/crates/edr_eth/src/eips/eip1559.rs
+++ b/crates/edr_eth/src/eips/eip1559.rs
@@ -16,7 +16,7 @@ impl<HardforkT> ForkBaseFeeParams<HardforkT> {
 
 /// Type that allows specifying constant or dynamic EIP-1559 parameters based on
 /// the active hardfork.
-#[derive(Clone, Copy, Debug)]
+#[derive(Clone, Debug)]
 pub enum BaseFeeParams<HardforkT: 'static> {
     /// Constant [`ConstantBaseFeeParams`]; used for chains that don't have
     /// dynamic EIP-1559 parameters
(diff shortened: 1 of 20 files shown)


crates/edr_eth/src/eips/eip1559.rs
```diff
@@ -16,6 +16,7 @@ impl<HardforkT> ForkBaseFeeParams<HardforkT> {
 
 /// Type that allows specifying constant or dynamic EIP-1559 parameters based on
 /// the active hardfork.
+#[derive(Clone, Copy, Debug)]
```
Wodann, line 19: What made you decide to add `Copy` here?

crates/edr_eth/src/spec.rs
```diff
@@ -26,10 +26,10 @@ pub trait ChainSpec {
 }
 
 /// Constants for constructing Ethereum headers.
-pub trait EthHeaderConstants: ChainSpec<Hardfork: 'static + PartialOrd> {
-    /// Parameters for the EIP-1559 base fee calculation.
-    const BASE_FEE_PARAMS: BaseFeeParams<Self::Hardfork>;
```
Wodann, line 31: I think it would be clearer to the user that the `fn chain_base_fee_params` can be wrong if we return an `Option`. That way we are signalling: "If the chain_id is known, we will return its BaseFeeParams".

That would require the a default `BASE_FEE_PARAMS` to remain. Maybe just rename it to `DEFAULT_BASE_FEE_PARAMS`.

At a call site, the user can decide to do:

```rs
// reverting example
ChainSpecT::chain_base_fee_params(chain_id).or(ChainSpecT::DEFAULT_BASE_FEE_PARAM)

// propagation error
ChainSpecT::chain_base_fee_params(chain_id).ok_or_else(|| anyhow!("Chain ID is not supported"))?;
```

crates/edr_evm/src/blockchain/forked.rs
```diff
@@ -576,6 +581,13 @@ where
             })?)
         }
     }
+
+    fn base_fee_params(&self) -> ConstantBaseFeeParams {
+        *self
+            .base_fee_params
+            .at_hardfork(self.hardfork())
+            .expect("Chain spec must have base fee params for post-London hardforks")
```
Wodann, line 589: It seems safer to have a default to revert to if the hardfork is not present. Since we won't have control over the defined base fee parameters, we would now be panicking.

crates/edr_eth/src/eips/eip1559.rs
frangio: It's used here:

https://github.com/NomicFoundation/edr/blob/ce413233603d3a35e07ccb50cdd1c652a43176ec/crates/edr_op/src/spec.rs#L167-L168

Generally I add `Copy` along with `Clone` if it's valid. Any reason not to?

Wodann: Rust as a language has a very valid reason to have a distinction between `Clone` and `Copy` (see [trait docs](https://doc.rust-lang.org/std/marker/trait.Copy.html)):
> By default, variable bindings have ‘move semantics.’
>
> However, if a type implements Copy, it instead has ‘copy semantics’
> 
> Copies happen implicitly, for example as part of an assignment y = x.
> 
> Cloning is an explicit action, x.clone().

Normally, variables use move semantics when being assigned, to make it clear to the user that a value has gone out of scope. If they want to make a copy of the underlying memory, they have to be explicit about this by using `.clone()` as this operation may be expensive.

The `Copy` trait is only implemented for types that are cheap to copy. As a general rule, we do not use `Copy` for values that are larger than 64-bits. Instead, the caller has to be explicit about calling `Clone`, signalling that they are don't want to move the value.

In your example usage you can use `cloned()` if you remove the `Copy` derivation.

frangio: Right, I explicitly chose to use `Copy`/`.copied()` here as I didn't think this was an expensive copy, but I can see that it's a pretty large type at 384 bits.

frangio:
Closing as the fix will have to take a different shape: see https://github.com/NomicFoundation/edr/issues/887.

PR #899, opened 2025-05-06
feat: solidity test filtering

frangio:
Adds a `test_pattern` option in the Solidity tests configuration to filter tests by a regex. Eventually this would be reflected in a `hardhat test solidity --grep` option, similar to the existing `hardhat test node --grep`, or a `--match-test` option.


crates/edr_napi/src/solidity_tests/config.rs
```diff
@@ -413,6 +417,20 @@ impl TryFrom<FuzzConfigArgs> for FuzzConfig {
     }
 }
 
+impl SolidityTestRunnerConfigArgs {
+    pub fn try_get_test_filter(&self) -> napi::Result<TestFilterConfig> {
```
Wodann, line 421:
```suggestion
    pub fn try_to_get_test_filter(&self) -> napi::Result<TestFilterConfig> {
```

Nit: I have a preference for readable function names, albeit slightly longer.

Feel free to decide either way.

crates/edr_napi/test/solidity-tests.ts
```diff
@@ -76,6 +76,31 @@ describe("Solidity Tests", () => {
...
+    ];
+    // All artifacts are test suites.
+    const testSuites = artifacts.map((artifact) => artifact.id);
+
+    const results = await runAllSolidityTests(artifacts, testSuites, {
+      projectRoot: __dirname,
+      testPattern: 'Multiply',
+    });
+
+    assert.equal(results.length, artifacts.length);
```
Wodann, line 92: From an integration test perspective, this result would be the same as no filter being applied. I'd recommend adding at least one test that doesn't match all artifacts.

Wodann (commented):
One nit and one request for change. Otherwise this looks good to me.

I'll leave the final approval to @agostbiro in case he wants to merge this after he finishes his merge to `main`.

agostbiro (changes requested):
Thanks looks good overall with some minor comments. 

The following CI test failures are safe to ignore:

```
    remote::receipt_rlp_encoding_pre_eip658
    remote::remote_block_receipt_root_pre_eip658
```

crates/edr_napi/test/solidity-tests.ts
frangio: Note that every artifact will produce a suite result, even if it contains zero test results, so this assertion will always be true for every filter config.

crates/edr_napi/src/solidity_tests.rs
```diff
@@ -79,7 +81,7 @@ pub fn run_solidity_tests(
         };
 
         // Returns immediately after test suite execution is started
-        runner.test(Arc::new(EverythingFilter), tx_results);
+        runner.test(test_filter.clone(), tx_results);
```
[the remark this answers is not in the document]
frangio: Not necessary, removed.

agostbiro (approved):
Thanks LGTM. Please make sure that linters pass and after that feel free to merge.

frangio:
Will look into moving the test in a follow up PR.

PR #902, opened 2025-05-07
refactor: allow timestamp-based hardfork activations in N-API

Wodann:
Apart from exposing EDR's internal API for timestamp-based hardfork activations, this PR also:
- updates scenario files to conform to the new deserialization format
- added a CLI command to convert old to new scenario formats.
  - By default, this contains only a placeholder. For an example of conversion code, see https://github.com/NomicFoundation/edr/commit/948cbcd651953e553de20df16e042218e05eb55c
- updates the JS benchmark to use the new scenario format
- removes hardfork activations for deprecated testnets


crates/tools/js/benchmark/src/index.ts
```diff
@@ -419,10 +417,17 @@ function preprocessConfig(config: any) {
...
-    for (const { hardfork, blockNumber } of hardforks) {
-      hardforkHistory.set(normalizeHardfork(hardfork), blockNumber);
+    const chainConfig = config.providerConfig.chains[key];
+    for (const { condition, hardfork } of chainConfig.hardforkActivations) {
+      if (!_.isUndefined(condition.timestamp)) {
+        throw new Error("Hardfork activations by timestamp are not supported");
+      } else if (!_.isUndefined(condition.block)) {
+        hardforkHistory.set(normalizeHardfork(hardfork), condition.block);
+      } else {
+        throw new Error("Unsupported hardfork condition");
```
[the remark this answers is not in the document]
Wodann: This cannot be added until HH2 supports timestamp-based configuration in their config.

crates/tools/js/benchmark/src/index.ts
```diff
@@ -391,10 +393,6 @@ function preprocessConfig(config: any) {
     config.providerConfig.hardfork
   );
 
-  // "accounts" in EDR are "genesisAccounts" in Hardhat
```
Wodann, line 394: We no longer have `genesisAccounts` in EDR. Instead we have a `genesisState` but Hardhat doesn't expose this in its config.

crates/edr_evm/src/hardfork/l1.rs
frangio, line 1: Not sure if we need to add in EDR or if it should be in this PR but there is the new Hoodi testnet:

https://github.com/NomicFoundation/hardhat/blob/36df3e4820b52d60be51b2704d0f7d5d50e3b5ff/v-next/hardhat/src/internal/builtin-plugins/network-manager/chain-descriptors.ts#L87-L93

crates/edr_scenarios/src/old.rs
```diff
@@ -0,0 +1,23 @@
...
+            chain_type: value.chain_type,
+            logger_enabled: value.logger_enabled,
+            provider_config: value.provider_config,
+        }
+    }
+}
+
+/// Placeholder for the actual old scenario provider config.
+/// This should be replaced with the actual old scenario provider config
+pub type ScenarioProviderConfig = super::ScenarioProviderConfig;
```
frangio, line 23: Why is this necessary?

frangio (approved):
One thing I didn't understand but otherwise LGTM.

crates/tools/js/benchmark/src/index.ts
frangio: I assume this was left over from a previous change?

Wodann: Correct. The benchmark code wasn't updated in the past, after that field was removed.

crates/edr_scenarios/src/old.rs
frangio: Used above in line 8.

crates/edr_evm/src/hardfork/l1.rs
Wodann: Thanks for the tip! I've added a task on my todo list to do this in a follow-up PR.

PR #904, opened 2025-05-07
feat: require coverage library path to be specified

Wodann:
Require specification of the coverage library path when adding coverage instrumentation.

### DIFF
--- a/crates/edr_instrument/src/coverage.rs
+++ b/crates/edr_instrument/src/coverage.rs
@@ -7,7 +7,7 @@ use slang_solidity::{
     parser::{ParseError, Parser as SolidityParser, ParserInitializationError},
 };
 
-#[derive(Serialize)]
+#[derive(Debug, Serialize)]
 pub struct InstrumentationMetadata {
     pub tag: B256,
     pub kind: &'static str,
@@ -15,6 +15,7 @@ pub struct InstrumentationMetadata {
     pub end_utf16: usize,
 }
 
+#[derive(Debug)]
 pub struct InstrumentationResult {
     pub source: String,
     pub metadata: Vec<InstrumentationMetadata>,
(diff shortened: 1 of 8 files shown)


crates/edr_instrument/src/coverage.rs
```diff
@@ -97,7 +98,7 @@ pub fn instrument_code(
         }
     }
 
-    instrumented_source.push_str("\nimport \"hardhat/coverage.sol\";");
+    instrumented_source.push_str(&format!("\nimport \"{coverage_library_path}\";"));
```
frangio, line 101: Note that this could result in invalid syntax, for example if the library path contains a quote symbol.

I think we should try to avoid that at this point, rather than failing to compile the file at a later point, by either validating the library path or escaping invalid characters (quotes and backslashes?).

frangio (commented):
One point to consider, not sure if we want to do something about it.

crates/edr_napi/src/instrument.rs
```diff
@@ -91,8 +92,13 @@ pub fn add_statement_coverage_instrumentation(
...
 
-    let instrumented = coverage::instrument_code(&source_code, &source_id, solidity_version)
-        .map_err(|error| napi::Error::new(napi::Status::GenericFailure, error.to_string()))?;
+    let instrumented = coverage::instrument_code(
+        &source_code,
+        &source_id,
+        solidity_version,
+        &coverage_library_path,
+    )
+    .map_err(|error| napi::Error::new(napi::Status::GenericFailure, error.to_string()))?;
```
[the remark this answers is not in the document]
Wodann: Thanks for catching that! Addressed by https://github.com/NomicFoundation/edr/pull/904/commits/32358f49bdfa6ea5fed547605037c2bfa1d2c47b

crates/edr_instrument/src/coverage.rs
Wodann: Good call. Addressed in https://github.com/NomicFoundation/edr/pull/904/commits/da1e5c40fc64da7bf9ddd126450525ea5950ee0e

Wodann:
After talking to Omar, he told me that Slang has not been doing any crate releases for a long time. He recommended me to upgrade to a git dependency with the commit that I used. I incorporated the change in this PR to simplify the review process.

crates/edr_instrument/src/coverage.rs
frangio: Very cool solution!

frangio (approved):
LGTM

Wodann:
Squashed to clean up the merge commit

PR #928, opened 2025-06-05
refactor: make HaltReason generic in Foundry crates

Wodann:
This PR makes the `HaltReason` generic in Foundry crates.


frangio:
Are the test failures related to this change?

PR #943, opened 2025-06-16
chore: downgrade to rust edition 2021

agostbiro:
Downgrade to Rust edition 2021 for compatibility with `feat/solidity-tests. We [originally](https://github.com/NomicFoundation/edr/pull/856) bumped the Rust edition to 2024 in main for compatibility with REVM 21, but that doesn't seem necessary anymore.

This PR contains two types of changes:
- Changes the edition to 2021 in the root Cargo.toml
- Runs rustfmt that reorders the imports as they're sorted differently between the editions

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ resolver = "2"
 
 [workspace.package]
 version = "0.3.5"
-edition = "2024"
+edition = "2021"
 
 [profile.dev]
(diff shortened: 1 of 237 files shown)


Wodann (approved):
I checked the alloy workspace's Cargo.toml and they depend on the 2021 edition. This should be safe to merge

PR #944, opened 2025-06-17
chore: merge main into feat/solidity-tests

agostbiro:
This PR merges the latest and greatest from `main` into `feat/solidity-tests`. 

Mechanical conflicts are resolved in the merge commit 007ed18d3a76d71733eb028bef2432ac1314812b.

Subsequent commits in the branch fix small issues to make the CI pass. Once approved, I'll squash these into the merge commit, then merge the PR manually.

Issue #685, opened 2024-09-26
TODOs before merging `feat/solidity-tests` to `main`


agostbiro:
- [x] Remove `.github/workflows/edr-benchmark-feat-solidity-tests.yml`
- [x] Fix the TODOs in `.github/workflows/edr-benchmark.yml`
- [x] Fix the TODO in `hardhat-tests/test/internal/hardhat-network/stack-traces/test.ts`
- [x] Change pnpm override from `"hardhat>@ignored/edr"` to `"hardhat>@nomicfoundation/edr"` in package.json

PR #945, opened 2025-06-17
ci: adapt soltest branch benchmarks for main

agostbiro:
Removes the scenario benchmark for `feat/solidity-tests` and adapts the Solidity test benchmark to use `main` as the baseline instead of `feat/solidity-tests`.

Completes the todos in https://github.com/NomicFoundation/edr/issues/685

Stacked on top of https://github.com/NomicFoundation/edr/pull/944


Wodann (approved):
LGTM!

PR #950, opened 2025-06-18
chore: fix workflow caching of rpc requests

frangio:
We've identified potential issues with the use of GitHub Actions cache for RPC requests.

The cache at a key is immutable (except for evictions), once a cache key is written to it won't be updated. This is an issue for the RPC requests cache because the set of RPC requests can change as tests are added or depend on the OS (some tests are OS-specific). This PR fixes that by hashing the cache contents and using that in the key. Since the contents aren't known until the end of the run we always restore from the most recently created cache via `restore-keys`.

~Another issue is that caches are by default not shared across OSs. So the Windows run is not sharing a cache with the Ubuntu and macOS runs. This PR enables cross-OS caching.~ &rarr; Disabling this since some tests are OS-specific.

### DIFF
--- a/.github/workflows/edr-ci.yml
+++ b/.github/workflows/edr-ci.yml
@@ -58,8 +58,7 @@ jobs:
         with:
           path: |
             **/edr-cache
-          key: edr-rs-rpc-cache-v1
-          enableCrossOsArchive: true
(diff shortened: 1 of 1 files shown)


frangio:
This is not doing what we want. Seeing `Cache saved with key: edr-rs-rpc-cache-v1-` in the logs.

PR #951, opened 2025-06-19
ci: disable test that times out

agostbiro:
Disable "issue 543" `edr_napi` test on Ubuntu and MacOS that regularly times out in CI on these platforms. Locally the test passes for me on Ubuntu.


agostbiro:
Thanks for proposing a fix @Wodann I added it in a TODO comment [f0b698b](https://github.com/NomicFoundation/edr/pull/951/commits/f0b698b2836551f7177218b8312c282db9af724c)

PR #953, opened 2025-06-19
chore: replace Buffer with Uint8Array in Solidity tests interface

### LINKED ISSUE #946: Use `Uint8Array` instead of `Buffer` in Solidity tests interface
**Motivation**

Hardhat asked us to move away from Node specific types in the interface.

**Definition of Done**

- Remove `impl TryCast<Address> for Buffer` in [crates/edr_napi/src/cast.rs](https://github.com/NomicFoundation/edr/pull/944/files/007ed18d3a76d71733eb028bef2432ac1314812b..4018ae77f352935acec6f8a0dcccc4511d3f868c#diff-789df0bf7a4b9a3eee3025b30d4e26cdd4085a50ffd6a7687f42d56165a5bb06)
- Use `Uint8Array` instead of `Buffer` in Solidity test runner interface in `edr_napi` for consistency  with the EDR provider
- If we're manually constructing a `Buffer` in Rust in Solidity test code and return that to JS, replace that with an `ArrayBuffer`. For an example, see https://github.com/NomicFoundation/edr/blob/main/crates/edr_napi/src/call_override.rs#L59

**Notes**

This is breaking change so we need to bump the minor version of `@nomicfoundation/edr`.

More context: https://github.com/NomicFoundation/edr/pull/944#discussion_r2152228659

Related: https://github.com/NomicFoundation/edr/issues/746

popescuoctavian:
This PR replaces all occurrences of `Buffer` with `Uint8Array` in Solidity tests interface.

I have also bumped up the minor version since this is a breaking change.

closes #946

### DIFF
--- a/.changeset/rotten-vans-divide.md
+++ b/.changeset/rotten-vans-divide.md
@@ -1,0 +1,5 @@
+---
+"@nomicfoundation/edr": minor
+---
+
+Replaced `Buffer` with `Uint8Array` in Solidity tests interface
--- a/crates/edr_napi/index.d.ts
+++ b/crates/edr_napi/index.d.ts
@@ -524,12 +524,12 @@ export interface SolidityTestRunnerConfigArgs {
    * The value of `msg.sender` in tests as hex string.
    * Defaults to `0x1804c8AB1F12E6bbf3894d4083f33e07309d1f38`.
    */
-  sender?: Buffer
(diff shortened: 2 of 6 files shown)


Wodann (changes requested):
One suggestion. You can look at other call sites for conversion from `[u8]` to `Uint8Array`

PR #954, opened 2025-06-19
ci: improve soltest reliability

agostbiro:
This PR does two things to improve the reliability of the Rust integration test suite for Solidity tests:

1. Reduce the max decimals in the assertion test fuzz tests, as with the previous setting [this line](https://github.com/NomicFoundation/edr/blob/f4f1408ec106bae9142c80fbd9d1392d0c704886/crates/edr_solidity_tests/tests/testdata/default/cheats/Assert.t.sol#L57) could overflow. 
2. Create a new temporary directory for failure persistence for each test instead of sharing it. This improves the reliability of invariant tests.

I've also removed the `serial` attributes from cheatcode test as they showed no difference in my testing and this speeds up test execution.


crates/edr_solidity_tests/tests/testdata/default/cheats/Assert.t.sol
```diff
@@ -6,7 +6,8 @@ import "cheats/Vm.sol";
 
 contract AssertionsTest is DSTest {
     string constant errorMessage = "User provided message";
-    uint256 constant maxDecimals = 77;
+    // For i256
+    uint256 constant maxDecimals = 76;
```
frangio, line 10: Instead of this I'd consider updating our definition of `_formatWithDecimals` to [Foundry's current](https://github.com/foundry-rs/foundry/blob/master/testdata/default/cheats/Assert.t.sol#L48-L53).

agostbiro: Thanks, fixed in https://github.com/NomicFoundation/edr/commit/66dde67f649e8150acfad8abd95c99015586abdb

PR #956, opened 2025-06-20
test: fix trying to get a compiler before downloaded

agostbiro:
Fixes the occasional 

```
HardhatError: HH11: An internal invariant was violated: Trying to get a compiler before it was downloaded
```

error when running stack trace tests.


frangio (approved):
Makes sense based on: https://github.com/NomicFoundation/hardhat/blob/3a11c02612edc7c3844f3331e94ee58b78e60337/packages/hardhat-core/src/internal/solidity/compiler/downloader.ts#L70

PR #959, opened 2025-06-20
chore: backport latest Foundry expectRevert

frangio:
This was backported from the latest Foundry commit with matching revm and alloy versions (foundry-rs/foundry@066e0ce0).

The starting point was to copy over `ExpectRevert.t.sol` as is. All `expectRevert` cheatcode impls and `expect_revert` helper were copied over (overwriting the existing ones), then adjusted to compile (generics, `apply_full`, `IsPure`). `depth` values were kept `u64` to minimize impact in the rest of the code (they're `usize` upstream now). `handle_expect_revert` was moved into a new module like upstream and updated to the latest version. Finally, `inspector.rs` aspects related to expected reverts were selectively copied over. This left some errors in the rest of the tests that I tried to address with updates from upstream or minimal custom adjustments where necessary, the largest one being a custom cheatcode `_expectInternalRevert` (with underscore prefix as it's for internal use) to replace the inline config that EDR doesn't support.

### DIFF
--- a/crates/edr_solidity_tests/tests/testdata/default/cheats/ExpectRevert.t.sol
+++ b/crates/edr_solidity_tests/tests/testdata/default/cheats/ExpectRevert.t.sol
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
-pragma solidity 0.8.18;
+pragma solidity ^0.8.18;
 
 import "ds-test/test.sol";
 import "cheats/Vm.sol";
@@ -28,6 +28,10 @@ contract Reverter {
     function callThenRevert(Dummy dummy, string memory message) public pure {
         dummy.callMe();
         revert(message);
+    }
+
+    function callThenNoRevert(Dummy dummy) public pure {
+        dummy.callMe();
     }
 
     function revertWithoutReason() public pure {
@@ -80,20 +84,10 @@ contract ExpectRevertTest is DSTest {
         reverter.revertWithMessage("revert");
     }
 
-    function testFailExpectRevertWrongString() public {
+    function testShouldFailIfExpectRevertWrongString() public {
         Reverter reverter = new Reverter();
-        vm.expectRevert("my not so cool error");
+        vm.expectRevert("my not so cool error", 0);
         reverter.revertWithMessage("my cool error");
-    }
-
-    function testFailRevertNotOnImmediateNextCall() public {
-        Reverter reverter = new Reverter();
-        // expectRevert should only work for the next call. However,
-        // we do not immediately revert, so,
-        // we fail.
-        vm.expectRevert("revert");
-        reverter.doNotRevert();
-        reverter.revertWithMessage("revert");
     }
 
     function testExpectRevertConstructor() public {
@@ -133,18 +127,6 @@ contract ExpectRevertTest is DSTest {
         dummy.largeReturnType();
     }
 
-    function testFailExpectRevertErrorDoesNotMatch() public {
-        Reverter reverter = new Reverter();
-        vm.expectRevert("should revert with this message");
-        reverter.revertWithMessage("but reverts with this message");
-    }
-
-    function testFailExpectRevertDidNotRevert() public {
-        Reverter reverter = new Reverter();
-        vm.expectRevert("does not revert, but we think it should");
-        reverter.doNotRevert();
-    }
-
     function testExpectRevertNoReason() public {
         Reverter reverter = new Reverter();
         vm.expectRevert(bytes(""));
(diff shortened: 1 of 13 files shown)


crates/edr_solidity_tests/tests/testdata/default/cheats/ExpectRevert.t.sol
```diff
@@ -80,22 +84,13 @@ contract ExpectRevertTest is DSTest {
         reverter.revertWithMessage("revert");
     }
 
-    function testFailExpectRevertWrongString() public {
+    function testFailIfExpectRevertWrongString() public {
+        vm.skip(true); // TODO
```
frangio, line 88: This test is failing with:

```
thread 'tokio-runtime-worker' panicked at /home/frangio/nomic/edr/crates/edr_solidity_tests/src/multi_runner.rs:428:26:
failed to join task: JoinError::Panic(Id(57), "internal error: entered unreachable code: slice OOB: 64..96", ...)

thread 'cheats::test_cheats_local_default' panicked at crates/edr_solidity_tests/tests/it/helpers/config.rs:138:30:
called `Result::unwrap()` on an `Err` value: Test testFailIfExpectRevertWrongString() did not pass as expected.
Reason: None
Logs:
```

<details><summary>Traces</summary>

```
  [45524] → new <unknown>@0x5F65cD7D792E9746EF82929D60de9a1C526f93A5
    └─ ← [Return] 227 bytes of code

  [43321] → new <unknown>@0x57fDDB5Bf9b1660898bEf442fC60ee4924CF6226
    └─ ← [Return] 216 bytes of code

  [44521] → new <unknown>@0x4d343D3b8C3c5765746BcfF748AB3E32C6cc37BC
    └─ ← [Return] 222 bytes of code

  [99572] → new <unknown>@0x21bCA69b10e0a09bf9557E9eDfdaC03AAD44Bd59
    └─ ← [Return] 497 bytes of code

  [80354] → new <unknown>@0x373EbB870b150685555e46A9097965609b787f2d
    └─ ← [Return] 401 bytes of code

  [162233] → new <unknown>@0x208a10454580563ED24A5dD56b03F51928A95771
    └─ ← [Return] 810 bytes of code

  [46924] → new <unknown>@0xFB6784F8386FDc81FE4eFeFC42e48cb1F33018d9
    └─ ← [Return] 234 bytes of code

  [48324] → new <unknown>@0x05F6Dc308173EFcBfB1853f130E466C4A5c60698
    └─ ← [Return] 241 bytes of code

  [43321] → new <unknown>@0x8d35bb3497605b2b3d32dcDC311133B120B0A7D8
    └─ ← [Return] 216 bytes of code

  [43321] → new <unknown>@0x74c524f11A3943a30ed2459156848932A6B99f64
    └─ ← [Return] 216 bytes of code

  [73748] → new <unknown>@0x4E4D0B8A32abfe85Dfa3b54DeFfA75Df2EE4F531
    └─ ← [Return] 368 bytes of code

  [2150989] → new <unknown>@0x7FA9385bE102ac3EAc297483Dd6233D62b3e1496
    └─ ← [Return] 10633 bytes of code

  [310950] 0x7FA9385bE102ac3EAc297483Dd6233D62b3e1496::testFailIfExpectRevertWrongString()
    ├─ [273912] → new <unknown>@0x5615dEB798BB3E4dFa0139dFa1b3D433Cc23b72f
    │   └─ ← [Return] 1368 bytes of code
    ├─ [0] 0x7109709ECfa91a80626fF3989D68f67F5b1DD12D::expectRevert()
    │   └─ ← [Return]
    ├─ [1000] 0x5615dEB798BB3E4dFa0139dFa1b3D433Cc23b72f::revertWithMessage("my cool error") [staticcall]
    │   └─ ← [Revert] revert: my cool error
    └─ ← [Stop]
```

</details>

agostbiro: This is concerning. I did some debugging and the test result is reverted as success, which probably means the revert message is not checked?

agostbiro (changes requested):
Thanks, looks good overall, except for one test failure that needs to be fixed. I verified that the newly added test suites are executed as part of the cheatcode tests.

crates/foundry/cheatcodes/src/test/expect.rs
```diff
@@ -92,34 +101,97 @@ pub enum ExpectedRevertKind {
...
+    pub fn eq(&self, create_scheme: Self) -> bool {
+        matches!(
+            (self, create_scheme),
+            (Self::Create, Self::Create) | (Self::Create2, Self::Create2 { .. })
+        )
+    }
 }
 
 impl_is_pure_true!(expectCall_0Call);
+
```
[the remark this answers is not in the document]
frangio: It was a Claude artifact that I noticed and later forgot to remove.

crates/edr_solidity_tests/tests/testdata/default/cheats/ExpectRevert.t.sol
frangio: After a second look, this error (the test succeeding) was correct, because the test says it expects `count = 0` errors with the specified message, and that is indeed the case. You will see that the `, 0` was added in this PR, I'm not sure at what point this was added, whether it was me or Claude.

One of the subtle things that I had done manually was to port all failure tests from Foundry's `ExpectRevertFailure.t.sol` to `ExpectRevert.t.sol` here. After seeing this error, I've decided to redo that part in a less error-prone way, by copying the test files (include `*Failure.t.sol`) as is, without moving any tests around. The only two changes that are required are to adjust imports (e.g., `'./Vm.sol'` to `cheats/Vm.sol`) and to rename `testShouldFail*` functions in `*Failure.t.sol` to `testFail*`. This is now what's included in this PR.

frangio: More thoughts: I'm not sure this test should pass though... It seems like a bug that `vm.expectRevert` is swallowing a revert that is not the one that's expected. I'll investigate.

crates/edr_solidity_tests/tests/testdata/default/cheats/ExpectRevert.t.sol
```diff
@@ -80,20 +84,24 @@ contract ExpectRevertTest is DSTest {
         reverter.revertWithMessage("revert");
     }
 
-    function testFailExpectRevertWrongString() public {
+    function testExpectRevertWithEncodedErrorPrefix() public {
+        vm.skip(true);  // TODO: This change has not been backported yet.
```
frangio, line 88: This should be resolved in a future PR.

agostbiro: Wdyt about removing the entire test to make it easy to see that the feature is not supported in a code diff?

frangio: Yeah, I think that will be more obvious when diffing, just as if we had backported before the test was added.

crates/edr_solidity_tests/tests/testdata/default/cheats/ExpectRevert.t.sol
frangio: Reported in https://github.com/foundry-rs/foundry/issues/10858.

PR #966, opened 2025-06-25
misc: add comments to send transaction implementation

agostbiro:
Adds some comments to the automining implementation to document edge cases and explain the code. Please review that the comments are correct.


popescuoctavian (approved):
Explanation seems clear to me. Thanks for looking into it!

PR #969, opened 2025-06-25
feat: add local predeploys for soltests

agostbiro:
Adds the option to set local predeploy contracts for Solidity tests. This unblocks setting OP predeploys for Solidity tests which is a must have. It also unblocks setting the beacon root and history storage account for L1 for Solidity tests. The latter is a nice to have.

The config interface follows the provider to make it easy to integrate in Hardhat.

The predeploys are added in local mode only (as opposed to fork mode). The rationale for this is that the forked network should have the predeploys and this restriction greatly simplifies the implementation. The downside is that testing a new predeploy that is not live on the forked network yet will not work.


crates/edr_napi/package.json
```diff
@@ -55,7 +55,7 @@
     "artifacts": "napi artifacts",
     "build": "pnpm run build:publish",
     "build:debug": "napi build --platform  --no-const-enum",
-    "build:dev": "napi build --platform  --release --no-const-enum",
+    "build:dev": "pnpm build:op",
```
agostbiro, line 58: I changed this, because now not just the `edr_napi` tests depend on the `op` feature, but `js/integration-tests/solidity-tests` depend on it as well.

crates/edr_napi/src/account.rs
```diff
@@ -1,28 +1,134 @@
...
+    cast::TryCast,
+    serde::{
+        serialize_bigint_as_struct, serialize_optional_bigint_as_struct,
+        serialize_optional_uint8array_as_hex, serialize_uint8array_as_hex,
+    },
+};
+
 /// Specification of overrides for an account and its storage.
 #[napi(object)]
+#[derive(Clone, serde::Serialize)]
```
agostbiro, line 15: Needs `Serialize` as the soltest config is `Serialize`

crates/edr_napi/src/account.rs
```diff
@@ -1,28 +1,134 @@
...
+        let nonce = account_override
+            .nonce
+            .ok_or_else(predeploy_error!("nonce"))?;
+        let code = account_override.code.ok_or_else(predeploy_error!("code"))?;
+
+        if code.is_empty() {
+            return Err(napi::Error::from_reason(
+                "Predeploy with address '{address}' must have non-empty code",
+            ));
+        }
```
agostbiro, line 106: I made these errors, because all the current predeploys (L1 & OP) have all these fields and I assume that all future predeploys would have these fields set as well.

Wodann: This is in fact not guaranteed. At least for JSON-RPC providers, Hardhat allows users to specify genesis state using `AccountOverride`. As such, all values are in fact overrides and not guaranteed to be specified

crates/foundry/evm/core/src/backend/mod.rs
```diff
@@ -579,17 +585,21 @@ impl<
     > Backend<BlockT, TxT, EvmBuilderT, HaltReasonT, HardforkT, TransactionErrorT, ChainContextT>
 {
     /// Creates a new Backend with a spawned multi fork thread.
-    pub fn spawn(fork: Option<CreateFork<BlockT, TxT, HardforkT>>) -> Self {
-        Self::new(MultiFork::spawn(), fork)
+    pub fn spawn(
+        fork: Option<CreateFork<BlockT, TxT, HardforkT>>,
+        local_predeploys: impl IntoIterator<Item = Predeploy>,
```
Wodann, line 590: Aren't these mutually exclusive?

In that case, it would have stronger typing if defined as an `enum`

crates/foundry/evm/core/src/backend/mod.rs
```diff
@@ -579,17 +585,21 @@ impl<
...
     /// Creates a new instance of `Backend`
     ///
-    /// if `fork` is `Some` this will launch with a `fork` database, otherwise
-    /// with an in-memory database
+    /// If `fork` is `Some` this will launch with a `fork` database, otherwise
+    /// with an in-memory database.
     pub fn new(
         forks: MultiFork<BlockT, TxT, HardforkT>,
         fork: Option<CreateFork<BlockT, TxT, HardforkT>>,
+        local_predeploys: impl IntoIterator<Item = Predeploy>,
```
Wodann, line 602: Same question regarding mutual exclusivity

js/integration-tests/solidity-tests/test/testContext.ts
```diff
@@ -59,20 +65,35 @@ export class TestContext {
...
 
     return context;
   }
 
-  defaultConfig(): SolidityTestRunnerConfigArgs {
+  defaultConfig(
+    chainType: string = L1_CHAIN_TYPE
+  ): SolidityTestRunnerConfigArgs {
+    let localPredeploys = undefined;
+    if (chainType === L1_CHAIN_TYPE) {
```
Wodann, line 80: I'm not sure whether this should be accounted for in these tests, but just to make sure we're not forgetting anything: the generic chain type also uses `l1GenesisState`

crates/edr_napi/src/account.rs
Wodann: See https://github.com/NomicFoundation/edr/pull/971

crates/edr_napi/src/account.rs
Wodann: What about using default values like we do for the JSON-RPC provider?

See https://github.com/NomicFoundation/edr/blob/main/crates/edr_provider/src/data.rs#L2880

Wodann (commented):
The biggest question I have about the implementation is:

Why pigeon hole to "predeploys" when this could just as easily be called "genesis state", allowing users to start from any genesis state?

Pre-deploys are one form of genesis state, but different chains will have different types of genesis state. E.g. [pre-installs](https://specs.optimism.io/protocol/preinstalls.html); or sometimes accounts start with balance as a reward for work done in test nets; or users want to have some test accounts with balance.

agostbiro:
> Why pigeon hole to "predeploys" when this could just as easily be called "genesis state", allowing users to start from any genesis state?

The purpose of this PR is to make the minimal change to support OP predeploys (which we have promised to ship this week) while making it easy to integrate it Hardhat. Allowing the user to start from any genesis state for Solidity tests is a different beast that needs a broader product discussion.

> Pre-deploys are one form of genesis state, but different chains will have different types of genesis state. E.g. [pre-installs](https://specs.optimism.io/protocol/preinstalls.html); or sometimes accounts start with balance as a reward for work done in test nets; or users want to have some test accounts with balance.

Again, the purpose of this PR is to support OP predeploys, but it incidentally supports preinstalls as well which work the same as predeploys. Setting custom balances for accounts is achieved with cheatcodes in Solidity tests.

crates/edr_napi/src/account.rs
agostbiro: What problem would that solve?

Wodann: > My understanding is that Hardhat will not let users set these values, but populate them with the return data of l1GenesisState or opGenesisState

At least for the JSON-RPC provider, a user can override the balance of accounts in the genesis state - for owned accounts.

crates/edr_napi/package.json
Wodann: This requires an update of `index.d.ts` for the CI to succeed.

crates/edr_napi/src/account.rs
agostbiro: To pedal back a bit: are we in agreement that not setting an `AccountOverride` field for a predeploy is a mistake from the JS caller? This is for me the key question here.

> You are deviating from the meaning of the AccountOverride type by requiring that all values are set, which I don't understand the need for

I'm reusing the `AccountOverride` type to supply predeploys for Solidity tests. The reason for this is that predeploys are returned as `AccountOverride` from the `l1GenesisState` and `opGenesisState` functions and I want to make it possible to reuse these functions for Solidity tests configuration.

The code that we're discussing here doesn't require that all the fields of every `AccountOverride` are set. It only requires that if you supply an `AccountOverride` as a predeploy for Solidity tests, then it needs to have all of its fields set.

Now I don't think this approach is great, because it'd be better to express the need for all fields to be set in the interface, but this would require changes to the return types of the `l1GenesisState` and `opGenesisState` functions. I initially wanted to avoid changing this in this PR, because it's technically a breaking change, but I'm happy to do it if you agree with it. More on this here: https://github.com/NomicFoundation/edr/pull/969#discussion_r2167239265

> This ambiguity can be avoided by doing unwrap_or with default values, if you don't want to use the AccountOverride type inside the other Solidity test crates.

If we agree that not setting an `AccountOverride` field for a predeploy is a mistake, then I think hiding that mistake by having a default fallback value is not a good solution.

> At least for the JSON-RPC provider, a user can override the balance of accounts in the genesis state - for owned accounts.

This is not the case for Solidity tests, and making that possible is a product change out of scope for this PR.

agostbiro: Ok, done in [5320ccd](https://github.com/NomicFoundation/edr/pull/969/commits/5320ccdddd60fac14ab64098b74f1add5bf2495b)

agostbiro: Fixed the error message in https://github.com/NomicFoundation/edr/pull/969/commits/1e8719cd7ac85455a2d1e65b8763c413713de725

crates/edr_napi/package.json
agostbiro: I updated the `index.d.ts` here: https://github.com/NomicFoundation/edr/pull/969/commits/d8fb8af48db6fc1ffbe9b94f8ad47965a1f3785d

I also revised the `edr_napi` build commands and the CI check a bit to make it more consistent:

Both `build:dev` and `build:debug` now build with feature OP for consistency in development. I removed `build:op` as I didn't see a need for it anymore.

Since `build:dev` now builds with feature OP, but in the CI we want to check that the typings file matches the released version, I added a new command `build:typingFile` that matches `build:publish` except it's unoptimized.

agostbiro: Commit for changing the `edr_napi` build commands: https://github.com/NomicFoundation/edr/pull/969/commits/96d6eb6e4e9005b0ddb632dec4454d719618ceb5

crates/edr_napi/src/account.rs
Wodann: Thank you for hearing my feedback and making the change!

PR #971, opened 2025-06-25
refactor: remove unused code

Wodann:
This removes - seemingly - dead code that was made obsolete by the introduction of the multi-chain Solidity test runner API, incl. https://github.com/NomicFoundation/edr/blob/main/crates/edr_napi_core/src/solidity/config.rs#L88


agostbiro (commented):
I don't agree with this change. I have found it useful in the past to save the Solidity test runner config that was received from JS to disk to debug issues potentially caused by wrong conversion between NAPI and Rust types.

Original PR that added `Serialize`: https://github.com/NomicFoundation/edr/pull/849

Wodann:
> I have found it useful in the past to save the Solidity test runner config that was received from JS to disk to debug issues potentially caused by wrong conversion between NAPI and Rust types.

Just for my understanding, why wouldn't you just store the values in JS in that case? Isn't it error prone to mimic JS formatting on the Rust side?

agostbiro:
> Just for my understanding, why wouldn't you just store the values in JS in that case? Isn't it error prone to mimic JS formatting on the Rust side?

It's not always easy to log the values in JS. I'm pretty sure I added this when testing in a third-party repo

PR #974, opened 2025-06-27
fix: add custom precompiles support for eth_sendTransaction

popescuoctavian:
This PR adds support for custom precompiles for `eth_sendTransaction`. Previously, custom precompiles were only supported for `eth_call` with the addition of RIP-7212.


.changeset/perfect-meals-tie.md
```diff
@@ -0,0 +1,5 @@
+---
+"@nomicfoundation/edr": patch
+---
+
+Added custom precompiles support for `eth_sendTransaction`
```
agostbiro, line 5: Since we're treating this a bugfix, I'd probably phrase the changelog entry something like this: "Fixed custom precompiles not being applied in `eth_sendTransaction`. This enables RIP-7212 support in transactions.".

crates/edr_evm/src/test_utils.rs
```diff
@@ -244,7 +244,7 @@ pub async fn run_full_block<
     assert_eq!(replay_header.base_fee_per_gas, builder.header().base_fee);
 
     for transaction in replay_block.transactions() {
-        builder.add_transaction(transaction.clone())?;
+        builder.add_transaction(transaction.clone(), &HashMap::new())?;
```
agostbiro, line 247: Nit: this would be easier to read if the empty hash map for the custom precompiles was assigned to a variable outside of the loop.

crates/edr_napi/test/artifacts/CustomPrecompile.json
```diff
@@ -0,0 +1,18 @@
...
+  "linkReferences": {},
+  "deployedLinkReferences": {}
+}
```
agostbiro, line 18: Nit: could you add an empty line as the last?

crates/edr_napi/test/provider.ts
```diff
@@ -475,6 +478,179 @@ describe("Provider", () => {
       assert.lengthOf(rawTraces, 1);
     });
   });
+
+  it("custom precompile enabled", async function () {
+    const contractArtifact = loadContract("./artifacts/CustomPrecompile.json");
```
agostbiro, line 483: Could you add the location of the Solidity source file in a comment?

crates/edr_napi/test/provider.ts
```diff
@@ -475,6 +478,179 @@ describe("Provider", () => {
...
+      })
+    );
+
+    const precompileReceipt = JSON.parse(
+      precompileTransactionReceiptResponse.data
+    ).result;
+    assert.strictEqual(precompileReceipt.status, "0x1");
+  });
+
+  it("custom precompile disabled", async function () {
```
agostbiro, line 569: The "custom precompile disabled" and "custom precompile enabled" seem to share most of their code. Could you factor that out into a helper function that loads the contract, creates the provider, deploys the contract and makes the precompile transaction?

crates/edr_provider/tests/integration/rip7212.rs
agostbiro, line 1: This file was restored from history, right?

agostbiro (changes requested):
Thanks, looks good overall! Just some minor issues to address

agostbiro (approved):

crates/edr_evm/src/block/builder.rs
```diff
@@ -110,6 +110,7 @@ where
     fn add_transaction(
         &mut self,
         transaction: ChainSpecT::SignedTransaction,
+        custom_precompiles: &HashMap<Address, PrecompileFn>,
```
Wodann, line 113: Currently it's possible for different transactions to be mined with different precompiles. Why did you not add them at construction time of the `BlockBuilder`?

agostbiro: That's a good suggestion for improvement. No need to phrase it like this though.

PR #976, opened 2025-06-27
fix: add custom precompiles support in Hardhat v2 for eth_sendTransaction

### LINKED ISSUE #886: Precompile 0x0100 (RIP-7212) returns empty returndata in eth_sendTransaction, but works with eth_call
Likely combine this with: #558

### Discussed in https://github.com/NomicFoundation/hardhat/discussions/6629

<div type='discussions-op-text'>

<sup>Originally posted by **craterface77** April 24, 2025</sup>
Hello!
I'm using Hardhat v2.22.18 with the option enableRip7212: true in hardhat.config.ts.

In my tests, I encountered unexpected behavior when calling precompile 0x0100 (P-256 / RIP-7212):

Behavior
• When directly calling eth_call on a contract, where staticcall(0x0100, ...) is used, the correct returndata = 0x...01 (32 bytes) is returned.
• But when calling the same method through a real transaction (eth_sendTransaction) — returndata is empty (0x), although success == true.

Context:
I call 0x0100 from the WebAuthnValidator smart contract inside EntryPoint.handleOps(...).
Here is an example of a Solidity call:
```
(bool success, bytes memory returndata) = address(0x100).staticcall(
abi.encode(h, r, s, qx, qy)
);
```

The arguments h, r, s, qx, qy are the same in both cases.

What I tried
• I made sure that everything works with eth_call.
• I made sure that all values ​​are valid (external signature validation is confirmed).
• I put console.logBytes(...) before the call - everything matches.
• I see the same thing on the OP Sepolia and Citrea networks: eth_call works, but tx returns an empty returndata.

Expected behavior

I expect 0x0100 to return not only on eth_call, but also in the actual transaction (eth_sendTransaction) if enableRip7212: true is enabled.

Possible explanation

It seems that the precompile is only enabled on the eth_call execution path, and not on the block simulation (tx-runner / mining path). If so, is there a fix planned?

```
// Setup in hardhat.config.ts
hard hat: { 
hardfork: "cancun", 
enableRip7212: true
}

// In test
await webAuthnValidator.validateUserOp(...) // eth_call → success
await entryPoint.handleOps(...) // eth_sendTransaction → returns 0x from 0x0100
```

Thank you!</div>

popescuoctavian:
This PR adds support for custom precompiles for eth_sendTransaction. Previously, custom precompiles were only supported for eth_call with the addition of RIP-7212.

This is the HH2 version of #974.

closes #886


frangio:
Should any cloning be necessary? I'd expect the hash map to be shared. Additionally, does cloning affect stateful precompiles? (In particular, does it prevent their state from persisting?)

PR #977, opened 2025-06-28
fix: OP block replay

Wodann:
This PR contains assorted fixes for the CI job that tests the OP block builder. The two fixes are related to:
- https://specs.optimism.io/protocol/isthmus/exec-engine.html#l2tol1messagepasser-storage-root-in-header
- https://specs.optimism.io/protocol/holocene/exec-engine.html#dynamic-eip-1559-parameters

I combined the fixes as they touch the same code.

We did not fix the first issue for forked blockchains yet, as we don't accurately calculate account storage roots. The changeset does not mention the second fix as it only affects our CI job at the moment.

For users to be able to fix the remaining issues on their end, we need to expose configuration parameters in the JSON-RPC or `ProviderConfig` to override the base fee parameters and `withdrawals_root` manually.


crates/edr_evm/src/miner.rs
```diff
@@ -394,8 +400,14 @@ where
         }
     }
 
-    let mut block_builder =
-        ChainSpecT::BlockBuilder::new_block_builder(blockchain, state, cfg.clone(), options)?;
+    // TODO: Add support for withdrawals
```
agostbiro, line 403: Could you link an issue please?

crates/edr_evm/src/test_utils.rs
```diff
@@ -334,12 +325,12 @@ pub async fn run_full_block<
...
-                    actual.inner.data.data,
-                    "{:?}",
-                    replay_block.transactions()[expected.transaction_index as usize]
-                );
+                // debug_assert_eq!(
+                //     expected.inner.data.data,
+                //     actual.inner.data.data,
+                //     "{:?}",
+                //     replay_block.transactions()[expected.transaction_index as
+                // usize] );
```
agostbiro, line 333: Was this commented on purpose?

crates/edr_eth/src/block.rs
```diff
@@ -253,30 +257,47 @@ impl PartialHeader {
...
+                if hardfork.into() >= l1::SpecId::SHANGHAI {
+                    let withdrawals_root = withdrawals.map_or(KECCAK_NULL_RLP, |withdrawals| {
+                        trie::ordered_trie_root(withdrawals.iter().map(alloy_rlp::encode))
+                    });
+
+                    Some(withdrawals_root)
+                } else {
+                    None
+                }
+            }),
```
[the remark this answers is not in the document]
frangio: I believe the line `if hardfork.into() >= l1::SpecId::SHANGHAI {` is taking care of the Canyon case.

See: https://github.com/bluealloy/revm/blob/8718223e941cf1dc5f02c9b993f3b490ff817542/crates/optimism/src/spec.rs#L27

But I guess matching the spec exactly (`keccak256(rlp(empty_string_code))`) will depend on `withdrawals` being empty.

crates/edr_evm/src/block/builder.rs
```diff
@@ -94,7 +95,8 @@ where
         >,
         state: Box<dyn SyncState<Self::StateError>>,
         cfg: CfgEnv<ChainSpecT::Hardfork>,
-        options: BlockOptions,
+        withdrawals: Option<Vec<Withdrawal>>,
```
frangio, line 98: IIUC, this will only ever be used for L1. Does it make sense to add it to the `BlockBuilder` trait that will need to be used for all chain types? I'm wondering if this should be specific to the `ChainSpecT` somehow...

crates/edr_evm/src/miner.rs
Wodann: Thanks for catching that. I added an issue and referenced it: https://github.com/NomicFoundation/edr/pull/977/commits/39c8d0968e298cbbd115865da5249384292321cf

crates/edr_evm/src/test_utils.rs
```diff
@@ -348,12 +339,12 @@ pub async fn run_full_block<
...
-            actual.as_execution_receipt(),
-            "{:?}",
-            replay_block.transactions()[expected.transaction_index() as usize]
-        );
+        // debug_assert_eq!(
+        //     expected.as_execution_receipt(),
+        //     actual.as_execution_receipt(),
+        //     "{:?}",
+        //     replay_block.transactions()[expected.transaction_index() as
+        // usize] );
```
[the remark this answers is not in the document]
Wodann: It was commented for debugging purposes. Thanks for catching it.

Addressed by https://github.com/NomicFoundation/edr/pull/977/commits/44cddf20db35ab376cd5561d91d1eb59634c96d2

crates/edr_eth/src/block.rs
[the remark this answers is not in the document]
Wodann: I discussed options with @frangio. There is still room for improvement to fully reject the presence of `Withdrawals` for chain types that don't need them. However, that will require larger refactors in the provider, to prevent passing those arguments to the block builder.

For now, I only made changes to make it clearer to validate that the implementation is to spec; i.e. that withdrawals is empty and thus we calculate `keccak256(rlp(empty_string_code))`.

Addressed by https://github.com/NomicFoundation/edr/pull/977/commits/4aa7cca914f42c1ed2916ccb1616d7ced7741eba

crates/edr_evm/src/block/builder.rs
Wodann: I tried to adapt this but as expressed in this [comment](https://github.com/NomicFoundation/edr/pull/977#discussion_r2183195009), it requires larger refactors in the provider to be able to determine which inputs should be sent to which block builder.

crates/edr_op/src/block/builder.rs
```diff
@@ -75,16 +83,30 @@ where
...
+                let elasticity: [u8; 4] = u32::try_from(base_fee_params.elasticity_multiplier)
+                    .expect("Base fee elasticity can only be up to u32::MAX")
+                    .to_be_bytes();
+
+                let bytes: Box<[u8]> = Box::new([
+                    DYNAMIC_BASE_FEE_PARAM_VERSION,
+                    denominator[0],
+                    denominator[1],
+                    denominator[2],
+                    denominator[3],
```
frangio, line 98: Personally prefer `extend_from_slice` since it doesn't depend on getting the number of elements right here.

frangio (commented):
Looks good. Unsure if the CI errors are caused by these changes.

crates/edr_op/src/block/builder.rs
```diff
@@ -35,12 +41,72 @@ where
...
+        mut overrides: edr_eth::block::HeaderOverrides,
+    ) -> Result<Self, BlockBuilderCreationError<Self::BlockchainError, OpSpecId, Self::StateError>>
+    {
+        // TODO: https://github.com/NomicFoundation/edr/issues/990
+        // Replace this once we can detect chain-specific block inputs in the provider
+        // and avoid passing them as input.
+        if cfg.spec >= OpSpecId::CANYON {
+            // `EthBlockBuilder` expects `inputs.withdrawals.is_some()` despite OP not
+            // supporting withdrawals.
+            inputs.withdrawals = Some(Vec::new());
```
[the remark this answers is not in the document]
Wodann: The reason I don't do this is because we cannot distinguish between OP and L1 from the caller. As such, doing that would break OP.

Until we introduce custom handlers for OP vs L1, we cannot be that accurate in the typing nor runtime values.

crates/edr_evm/src/block/builder.rs
```diff
@@ -50,6 +57,33 @@ impl<BlockchainErrorT, HardforkT: Debug, StateErrorT>
     }
 }
 
+/// Chain-agnostic inputs for building a block.
```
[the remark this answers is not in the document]
Wodann: I'm using "chain-agnostic" here to mean that it doesn't care about which chain type is being used. Opposite to "chain-specific", in which case we'd have separate types for OP (without withdrawals) and L1 (with withdrawals).

PR #983, opened 2025-07-01
refactor: use errors instead of panics for invariant violations

agostbiro:


agostbiro:
I realized this should have a changeset so I added one in https://github.com/NomicFoundation/edr/pull/983/commits/52c9826e15dc469607f24f0207e55c2266380087

crates/edr_solidity/src/error_inferrer.rs
```diff
@@ -635,7 +652,11 @@ fn check_last_submessage<HaltReasonT: HaltReasonTrait>(
     let call_stack_frame_source_reference = call_stack_frame
```
popescuoctavian, line 652: Nit: comment from above doesn't make sense anymore, it can probably be deleted (or was it added to indicate that we should not return an error message?)

agostbiro: Thanks, I'm not sure about the context (the file was moved so Git history is not too helpful), but I think it should be deleted which I did in https://github.com/NomicFoundation/edr/pull/983/commits/bf8f8877f2946b45f5e9c0b70426aaecb9469d71

Wodann:
Was this implemented as a temporary solution?

I mainly ask because of the usage of `anyhow` errors rather than `thiserror` errors. `anyhow` errors are designed for binary applications rather than libraries.

agostbiro:
> Was this implemented as a temporary solution?
> 
> I mainly ask because of the usage of `anyhow` errors rather than `thiserror` errors. `anyhow` errors are designed for binary applications rather than libraries.

There are some modules in `edr_solidity` that use `anyhow::Error` (I don't know why). Even though it'd be desirable, I didn't refactor these modules to use `thiserror`, as that was not the goal of this PR, and there would be too many changes to bundle into a single PR imo. 

I added `anyhow::Error` to one module: https://github.com/NomicFoundation/edr/pull/983/files#diff-ff0f7720d45a4cb47e8b33abc06ff55e0f11f995269e48e3c67eef0511998639

The rationale here is that these functions are called from a function that returns `anyhow::Error`, so as things are, there'd be no benefit to adding more granular error information.

PR #985, opened 2025-07-01
docs: add setup notes to improve VS Code experience

popescuoctavian:
Adds a note to the book to make sure the dev container has enough memory. Increasing from the default 8GB on macOS to 16GB solved most of the Rust language server crashes for me. Also adds a setting that speeds up rust-analyzer checks significantly by ignoring expanding on napi proc-macro.


agostbiro (approved):
Thanks, lgtm

PR #986, opened 2025-07-02
chore: disable Codecov statuses

frangio:
#982 did not disable the right thing: [status checks](https://docs.codecov.com/docs/commit-status) are not the same as [GitHub checks](https://docs.codecov.com/docs/github-checks).

Note Codecov should continue to post a comment in PRs.


agostbiro (approved):
Thanks!

Issue #894, opened 2025-04-29
Issue with the receive function


hyeonLewis:
I'm using hardhat with edr@0.10.0 and found the following issues related to the `receive` function. 
 
When the `receive() payable` has a modifier that uses a private function and is expected to be reverted, the EDR can't recognize the fallback. The existing function with payable (`callMe` in the example below) works as expected. Please check the below PoC.

<details><summary>PoC</summary>

```
// SPDX-License-Identifier: MIT
pragma solidity 0.8.25;

contract Test {
    bool public isReverted;

    modifier revertIf() {
        require(!isReverted, "I'm reverted"); // No issue when directly check the condition
        _;
    }

    function setIsReverted(bool _isReverted) external {
        isReverted = _isReverted;
    }

    function callMe() external payable revertIf {}

    receive() external payable revertIf {}
}
```

```
--- a/contracts/Mock/Test.sol
+++ b/contracts/Mock/Test.sol
@@ -5,7 +5,7 @@ contract Test {
     bool public isReverted;
 
     modifier revertIf() {
-        require(!isReverted, "I'm reverted");
+        _checkReverted(); // Issue occurred when using private function
         _;
     }
 
@@ -16,4 +16,8 @@ contract Test {
     function callMe() external payable revertIf {}
 
     receive() external payable revertIf {}
+
+    function _checkReverted() private view {
+        require(!isReverted, "I'm reverted");
+    }
 }
```

- Testing script:
```
it.only("#fallback: PoC", async function () {
  const { deployer, test } = await loadFixture(TestFixture);

  await test.setIsReverted(true);

  await expect(deployer.sendTransaction({ to: test.target, value: 100 })).to.be.revertedWith("I'm reverted");
});
```

- Testing result:
```
❯ export RUST_BACKTRACE=full
❯ hh test

thread '<unnamed>' panicked at crates/edr_solidity/src/error_inferrer.rs:1164:17:
This shouldn't happen: trying to get fallback source reference from a contract without fallback
stack backtrace:
   0:        0x13ecbce34 - _napi_register_module_v1
   1:        0x13ecd931c - _napi_register_module_v1
   2:        0x13ecba184 - _napi_register_module_v1
   3:        0x13ecbcce8 - _napi_register_module_v1
   4:        0x13ecbda54 - _napi_register_module_v1
   5:        0x13ecbd898 - _napi_register_module_v1
   6:        0x13ecbe208 - _napi_register_module_v1
[...]

PR #989, opened 2025-07-02
edr-0.11.3

agostbiro:
# @nomicfoundation/edr release

### Patch Changes

- 1ac9b2e4: Fixed custom precompiles not being applied in `eth_sendTransaction`. This enables RIP-7212 support in transactions.
- 7e12f64: Fixed panic on stack trace generation for receive function with modifier that calls another method. https://github.com/NomicFoundation/edr/issues/894
- 5861ad6: Turned panics during stack trace generation due to invalid assumptions into errors.

### DIFF
--- a/crates/edr_napi/npm/linux-x64-musl/package.json
+++ b/crates/edr_napi/npm/linux-x64-musl/package.json
@@ -4,7 +4,7 @@
     "url": "https://github.com/NomicFoundation/edr.git",
     "type": "git"
   },
-  "version": "0.2.0-alpha.1",
+  "version": "0.11.3",
   "main": "edr.linux-x64-musl.node",
   "files": [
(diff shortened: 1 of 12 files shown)


crates/edr_napi/npm/linux-x64-musl/package.json
```diff
@@ -4,7 +4,7 @@
     "url": "https://github.com/NomicFoundation/edr.git",
     "type": "git"
   },
-  "version": "0.2.0-alpha.1",
+  "version": "0.11.3",
```
frangio, line 7: Do you know why these weren't up to date?

agostbiro: I think the automated release process only bumps the package version in the root package.json in `crates/edr_napi`.

PR #991, opened 2025-07-03
chore: use mocking to speed up CI test

popescuoctavian:
Adds a fix that allows us to re-enable the test disabled in #951, for issue #543. The PR adds a `MockProvider` that always returns the given response, removing the need to perform a large amount of JSON-RPC requests and allowing us to test with custom JSONs.

The original response was very large, ~3GB. To reduce the time needed to run the test, we're generating a large JSON (~950 mil characters) ad-hoc, using a sample from the original response that gets duplicated. This still tests that the original fix works, since V8 limit is at ~500 mil characters.


crates/edr_napi/Cargo.toml
```diff
@@ -64,6 +89,7 @@ napi-build = "2.0.1"
 op = ["dep:edr_op"]
 scenarios = ["dep:edr_scenarios", "dep:rand"]
 tracing = ["edr_evm/tracing", "edr_napi_core/tracing", "edr_provider/tracing"]
+test-mock = []
```
popescuoctavian, line 92: Formatter changes apart from this one

crates/edr_napi/src/provider.rs
```diff
@@ -113,3 +115,17 @@ impl Provider {
...
+    #[doc = "Replaces the existing provider with a mock one, which always returns the given response."]
+    #[doc = "For testing purposes."]
+    #[napi]
+    pub fn use_mock_provider(&mut self, mocked_response: serde_json::Value) -> napi::Result<()> {
+        let mock_provider = mock::MockProvider::new(mocked_response);
+        self.provider = Arc::new(mock_provider);
+
+        Ok(())
+    }
+}
```
Wodann, line 131: Did you consider adding a standalone function `pub fn create_mock_provider -> Provider`? If so, did that cause any issues?

crates/edr_napi/test/files/issue-543.json
```diff
@@ -0,0 +1,582 @@
+{
```
Wodann, line 1: Given that we use the naming `data` for other folders like this (e.g. `./data` and `crates/edr_napi/data`), I would use the same name here instead of `files`.

crates/edr_napi/src/provider.rs
popescuoctavian: That was my first idea, to add it in [context.rs](https://github.com/NomicFoundation/edr/blob/d903f352ca6dc2c104742622eac49c26280ba32c/crates/edr_napi/src/context.rs#L140), but I figured we may want to reuse other parts of the provider in the future (config, logger, etc.) and adding it there would mean creating a separate function with all of those arguments that copies most of the code in `create_provider`.

Another option that I considered was to add it a new parameter to `create_provider` under the feature flag, but I'm not sure that's good practice for test code.

I decided this was the most elegant/less complex solution, but I'm open to changing it.

Wodann: Assuming that we just want to keep the design of this `MockProvider` as a provider that returns the configured response, there would be no need for any of the other configuration options that a provider normally receives. Those are all specific to the implementation of a normal provider.

It's safe to ignore them.

By doing so, you greatly simplify setup of the `MockProvider` too. You just need to pass in one value.

popescuoctavian: That makes sense, I figured we may want to have it set up as a "regular" provider. At that point the test was still in the initial file so it was already set up (moved it after I ran into CI failures due to it being behind a feature flag).

I'll do the changes quickly, thanks!

crates/edr_napi/test/mock.ts
```diff
@@ -0,0 +1,47 @@
...
+    const parsedJson = JSON.parse(fileContent);
+    const structLog = parsedJson.structLogs[0];
+
+    // This creates a JSON of length ~950 000 000 characters.
+    // JSON.stringify() crashes at ~500 000 000 characters.
+    for (let i = 1; i < 20000; i++) {
+      parsedJson.structLogs.push(structLog);
+    }
+
+    this.timeout(500_000);
```
popescuoctavian, line 19: In my tests it was always under 100.000ms, but gave it more to account for possible CI slowness

Wodann: As a nice to have, could you add this as a comment above the timeout so someone doesn't inadvertently remove it from the code because they don't understand the context?

crates/edr_napi/src/context.rs
```diff
@@ -315,6 +317,28 @@ impl EdrContext {
     }
 }
 
+#[cfg_attr(feature = "test-mock", napi)]
+#[cfg(feature = "test-mock")]
+impl EdrContext {
```
Wodann, line 322: Up to you, but for informational purpose:

In Rust you can have multiple `impl` blocks, spread across multiple files. As such, you could move this code to `mock.rs` to avoid the extra `#[cfg_attr]` etc.

Wodann (approved):
Awesome work! Looks good to me!

Such two informational comments / nice to haves. Feel free to incorporate the ideas if they before merging - if you agree.

No need to ask for re-review as long as CI passes.

crates/edr_napi/src/context.rs
popescuoctavian: That is way cleaner! Moved it, thanks for the suggestion!

PR #993, opened 2025-07-04
fix: use previous block's dynamic base fee parameters

Wodann:
The implementation in https://github.com/NomicFoundation/edr/pull/977 was part of the [spec](https://specs.optimism.io/protocol/holocene/exec-engine.html#dynamic-eip-1559-parameters). In particular, we were not reading the dynamic base fee parameters from the previous block header's `extra_data`:

> Prior to the Holocene upgrade, the EIP-1559 denominator and elasticity parameters used to compute the block base fee were [constants](https://specs.optimism.io/protocol/exec-engine.html#1559-parameters).
> 
> With the Holocene upgrade, these parameters are instead determined as follows:
> 
>     if Holocene is not active in parent_header.timestamp, the [prior EIP-1559 constants](https://specs.optimism.io/protocol/exec-engine.html#1559-parameters) are used. Note that parent_header.extraData is empty prior to Holocene, except possibly for the genesis block.
>     if Holocene is active at parent_header.timestamp, then the parameters from parent_header.extraData are used.

This PR fixes that oversight and tests it.

### DIFF
--- a/crates/edr_op/src/block/builder.rs
+++ b/crates/edr_op/src/block/builder.rs
@@ -1,5 +1,6 @@
 use edr_eth::{
-    block::PartialHeader, spec::EthHeaderConstants, trie::KECCAK_NULL_RLP, Address, Bytes, HashMap,
+    block::PartialHeader, eips::eip1559::ConstantBaseFeeParams, spec::EthHeaderConstants,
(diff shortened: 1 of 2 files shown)


frangio (approved):
LGTM

PR #995, opened 2025-07-07
chore: move custom precompiles to BlockBuilder

popescuoctavian:
This PR improves #974 by moving custom precompiles to the `BlockBuilder`, which removes the possibility of using different precompiles for different transactions.

### DIFF
--- a/crates/edr_evm/src/block/builder.rs
+++ b/crates/edr_evm/src/block/builder.rs
@@ -130,7 +130,7 @@ where
         cfg: CfgEnv<ChainSpecT::Hardfork>,
         inputs: BlockInputs,
         overrides: HeaderOverrides,
-        custom_precompiles: HashMap<Address, PrecompileFn>,
+        custom_precompiles: &'builder HashMap<Address, PrecompileFn>,
     ) -> Result<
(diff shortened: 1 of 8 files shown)


crates/edr_evm/src/block/builder.rs
```diff
@@ -130,6 +130,7 @@ where
         cfg: CfgEnv<ChainSpecT::Hardfork>,
         inputs: BlockInputs,
         overrides: HeaderOverrides,
+        custom_precompiles: HashMap<Address, PrecompileFn>,
```
Wodann, line 133: I recall that there was a discussion about using a reference for `HashMap` in a previous PR. Does that same line of reasoning apply here?

Wodann (approved):
Overall, LGTM.

I did have one question about an implementation detail

crates/edr_evm/src/block/builder.rs
popescuoctavian: Changed to using references. I also changed lifetime names to `'builder` for accuracy and consistency with implementations. Thanks!

Wodann (approved):

PR #996, opened 2025-07-07
chore: move custom precompiles to BlockBuilder

popescuoctavian:
This PR improves #976 by moving custom precompiles to the BlockBuilder, which removes the possibility of using different precompiles for different transactions.

The hh2 branch version of #995


popescuoctavian:
CI is still failing in `hh2` since it doesn't have Frangio's fix for alchemy tests.

PR #997, opened 2025-07-07
feat: add code coverage to Solidity test runner

Wodann:
Adds a code coverage reporter to the Solidity test runner that operates on instrumented Solidity code.


js/integration-tests/solidity-tests/test/testContext.ts
```diff
@@ -84,6 +84,7 @@ export class TestContext {
     }
 
     return {
+      observability: {},
```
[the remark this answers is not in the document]
Wodann: Addressed by https://github.com/NomicFoundation/edr/pull/997/commits/3ce013956e0bd55110a70aa7d5dd8653ed7f7c7c

crates/edr_napi/index.d.ts
```diff
@@ -637,6 +637,8 @@ export interface SolidityTestRunnerConfigArgs {
    * None.
    */
   includeTraces?: IncludeTraces
+  /** The configuration for the Solidity test runner's observability */
+  observability: ObservabilityConfig
```
Wodann, line 641: @frangio based on your TS experience, would you prefer this to be `observability?: ObservabilityConfig`?

frangio: Yes, if all of the subobject fields are optional I'd prefer to omit the field rather than to have to write `observability: {}`. In this case there's also a backwards compatibility argument, though I don't think we consider that an important concern atm.

Cargo.toml
```diff
@@ -183,7 +183,7 @@ chrono = { version = "0.4", default-features = false, features = [
     "std",
 ] }
 color-eyre = "0.6"
-derive_more = "0.99"
+derive_more = "2"
```
Wodann, line 186: I needed to bump the dependency to be able to use `derive_more::Debug`.

crates/foundry/evm/core/src/abi/console.rs
```diff
@@ -9,70 +9,70 @@ sol! {
 #[sol(abi)]
 #[derive(Display)]
 interface Console {
-    #[display(fmt = "{val}")]
```
Wodann, line 12: The format changed after upgrading `derive_more`

crates/edr_solidity_tests/src/multi_runner.rs
```diff
@@ -272,6 +277,7 @@ impl<
                     stack
                         .cheatcodes(Arc::new(cheats_config))
                         .trace(self.include_traces != IncludeTraces::None)
+                        .code_coverage(self.code_coverage.clone())
```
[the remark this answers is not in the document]
Wodann: Good catch. Addressed by https://github.com/NomicFoundation/edr/pull/997/commits/5db34dee3b2af717b85780f3e269276eab861b1b

frangio (approved):
LGTM!

PR #1002, opened 2025-07-10
fix: add catch unwind to napi interface

agostbiro:
## Problem

- While EDR should only panic if there is a logic error in the code, and ideally we would detect panics during development with the test suite, it is possible that panics happen in the wild.
- Rust functions that panic crash the Node.js main thread, and the panic messages are not reported to Sentry.

## Solution

- Annotate every function/method exposed from `edr_napi` to JS via the `catch_unwind` NAPI-RS macro attribute. This will turn panics into errors that are reported via the normal Node.js Sentry integration.

## Considerations

- Annotating all exposed functions and methods in `edr_napi` with `catch_unwind` showed [no performance regression.](https://github.com/NomicFoundation/edr/actions/runs/16192366605?pr=1002)
- This will not capture panics from background threads. To fix this, we could use Sentry from Rust, but EDR is a library and Sentry should be integrated in executables.

### DIFF
--- a/book/src/02_development/06_style_guide.md
+++ b/book/src/02_development/06_style_guide.md
@@ -62,8 +62,6 @@ impl Foo {
 
 Annotate every function/method exposed from `edr_napi` to JS with the `catch_unwind` NAPI-RS macro attribute.
 
-Note that this will not capture panics from background threads.
-
 ### Example
 
 ```rust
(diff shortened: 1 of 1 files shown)


agostbiro:
> Might be worth adding something like the following to the book (maybe style guide): "Apply #[napi(catch_unwind)] to all exported methods/functions that execute Rust code to allow errors to be propagated. "

Thanks, added in https://github.com/NomicFoundation/edr/pull/1002/commits/8e29a10e910aac106352a504403e0db298f1b3a5. Please re-review!

PR #1003, opened 2025-07-10
chore: update node to v22 in devcontainer

popescuoctavian:
Updates devcontainer node version to 22. Some parts of the code (HH3 project in `js/integration_tests/solidity_tests`) require node v22 to run.


frangio:
Based on [Hardhat's policy](https://hardhat.org/hardhat-runner/docs/reference/stability-guarantees#node.js-versions-support) I think we would want to maintain support for Node v20 for now, since it [remains in Maintenance status](https://nodejs.org/en/about/previous-releases).

PR #1005, opened 2025-07-15
feat: implement function-level gas tracking for Solidity tests

popescuoctavian:
This PR adds the ability to extract gas usage of Solidity test functions. Since input is chain-agnostic, we are implementing this directly in Typescript and exporting it as a separate package entry point.

### DIFF
--- a/js/helpers/package.json
+++ b/js/helpers/package.json
@@ -5,11 +5,6 @@
     "@nomicfoundation/edr": "workspace:*",
     "@nomicfoundation/hardhat-errors": "3.0.0-next.0",
     "@nomicfoundation/hardhat-utils": "3.0.0-next.0",
-    "fs-extra": "^7.0.1",
-    "hardhat": "3.0.0-next.4",
-    "semver": "^6.3.0"
-  },
-  "devDependencies": {
     "@tsconfig/node20": "^20.1.6",
     "@types/fs-extra": "^5.1.0",
     "@types/node": "^20.0.0",


crates/edr_napi/tsconfig.json
```diff
@@ -10,15 +10,17 @@
     "noEmitOnError": true,
     "skipDefaultLibCheck": true,
     "skipLibCheck": true,
-    "outDir": "./build-test",
+    "outDir": "./dist",
     "rootDirs": [
-      "./test"
+      "./test",
+      "./src/ts"
```
popescuoctavian, line 16: Tests depend on the root `index`, so I feel like their should still be in the root directory. Another solution is to have a separate `tsconfig.test.json` to avoid building test every time. @frangio what do you think is the best way to handle this?

crates/edr_napi/package.json
```diff
@@ -57,14 +61,16 @@
...
+    "build:publish": "napi build --platform --profile napi-publish --no-const-enum",
     "build:scenarios": "napi build --platform --release --no-const-enum --features scenarios",
     "build:tracing": "napi build --platform --release --no-const-enum --features tracing",
     "build:typingFile": "napi build --platform --no-const-enum",
     "clean": "rm -rf @nomicfoundation/edr.node",
     "eslint": "eslint 'test/**/*.ts'",
     "lint": "pnpm run prettier && pnpm run eslint",
     "lint:fix": "pnpm run prettier --write",
+    "postbuild": "pnpm tsc",
+    "postbuild:dev": "pnpm tsc",
```
popescuoctavian, line 73: `--target <target_arch>` gets appended to the build command, so I think this is the best solution here because we can't build the TS code before the napi build.

frangio:
There are some outdated aspects of our tsconfig.json such as the target. I'd recommend using TypeScript's recommended base config for our minimum supported Node.js version, eg. [@tsconfig/node22](https://www.npmjs.com/package/@tsconfig/node22), adding the fields we need (`declaration`, `declarationMap`, etc.).

crates/edr_napi/package.json
frangio, line 73: `pnpm` is not necessary here.

frangio, line 73: `pnpm` is not necessary here.

frangio: Since we have so many `build` task variants that only differ on flags, maybe we should create a script `build.sh` to share across all tasks with the ability to customize the flags and also run `tsc` after `napi` but still passing the `--target` flags to `napi`, so like:
```
#!/usr/bin/env bash
set -euo pipefail
napi build --platform --no-const-enum "$@"
tsc
```

Then:
```json
"build:publish": "bash script.sh --profile napi-publish",
```
etc.

crates/edr_napi/package.json
```diff
@@ -27,6 +27,10 @@
   "engines": {
     "node": ">= 18"
   },
+  "exports": {
+    ".": "./index.js",
+    "./solidity-tests": "./dist/src/ts/solidity_tests.js"
+  },
   "files": [
```
frangio, line 34: We need to add the new files in this array. Probably `"dist/src/"`.

frangio, line 34: I wonder why `Cargo.toml` and `build.rs` are here? I don't think they're necessary.

popescuoctavian:
> There are some outdated aspects of our tsconfig.json such as the target. I'd recommend using TypeScript's recommended base config for our minimum supported Node.js version, eg. [@tsconfig/node22](https://www.npmjs.com/package/@tsconfig/node22), adding the fields we need (`declaration`, `declarationMap`, etc.).

Working on this. Should we also update the node version here in `package.json`?
> "engines": {
    "node": ">= 18"
  },

frangio:
>Should we also update the node version here in `package.json`?

Yes, it should reflect our minimum required Node.js version.

crates/edr_napi/src/ts/solidity_tests.ts
```diff
@@ -0,0 +1,54 @@
...
+  return "runs" in object && "meanGas" in object && "medianGas" in object;
+}
+
+export function extractGasUsage(
+  testResults: Array<{
+    name: string;
+    kind: StandardTestKind | FuzzTestKind | InvariantTestKind;
+  }>,
+  filter?: GasUsageFilter,
+  ordering?: SortOrder
```
frangio, line 27: What's the reasoning to provide this functionality rather than let the caller do filtering and sorting?

I can see some argument for filtering because it makes more efficient use of memory. Ordering feels unnecessary since the caller of the function can sort the array in exactly the same way without much difficulty. They may not even want a sorted array.

I'd remove `ordering`. For filtering I'd consider taking a function that returns a boolean.

crates/edr_napi/src/ts/solidity_tests.ts
```diff
@@ -0,0 +1,54 @@
...
+): Array<{ name: string; gas: bigint }> {
+  const gasUsage: Array<{ name: string; gas: bigint }> = [];
+
+  for (const result of testResults) {
+    // Default to zero gas for invariant tests
+    const gas = isStandardTestKind(result.kind)
+      ? result.kind.consumedGas
+      : isFuzzTestKind(result.kind)
```
frangio, line 35: I'd suggest this, which is simpler and more type-safe (the helpers defined above are not typechecked):

```suggestion
    const gas = "consumedGas" in result.kind
      ? result.kind.consumedGas
      : "medianGas" in result.kind
```

crates/edr_napi/dist/src/ts/solidity_tests.d.ts
```diff
@@ -0,0 +1,17 @@
...
+    maxThreshold?: bigint;
+}
+export declare function extractGasUsage(testResults: Array<{
+    name: string;
+    kind: StandardTestKind | FuzzTestKind | InvariantTestKind;
+}>, filter?: GasUsageFilter, ordering?: SortOrder): Array<{
+    name: string;
+    gas: bigint;
+}>;
+//# sourceMappingURL=solidity_tests.d.ts.map
```
Wodann, line 17: Out of curiosity, what does this line do?

frangio: This tells tooling (runtime, LSP, etc.) that this file has been compiled/transpiled from another source file, and specifies how to convert positions in this file into positions in the source file. This sort of thing is used by stack traces or by the editor's "go to definition".

crates/edr_napi/src/ts/solidity_tests.ts
popescuoctavian: @frangio I'm happy to make the changes, if they still make more sense considering the Foundry feature-parity requirements.

frangio: I don't have a strong opinion, maybe we should merge this PR and discuss if we want to do API changes tomorrow?

frangio (approved):
Review changes look good to me. Merge may depend on resolution of the [above](https://github.com/NomicFoundation/edr/pull/1005#discussion_r2237999940).

crates/edr_napi/src/ts/solidity_tests.ts
popescuoctavian: Sounds good, merging this and will raise it today. Thanks!

PR #1006, opened 2025-07-16
research: move L1 types to separate crate

Wodann:
Historically, when making changes to the multi-chain architecture, it has frequently happened that adding a new chain-specific behaviour would compile in the "core crates": `edr_eth`, `edr_evm`, `edr_provider`, and `edr_napi_core`; but would fail to be integrable in `edr_op` and `edr_generic`. As a result, we'd need to go back to the drawing board to create a design that works across crates.

This problem stems from the `L1ChainSpec` implementation being part of the "core crates" which makes all types available within the same compilation unit AND avoids the [foreign trait cannot be implemented for foreign type problem](https://stackoverflow.com/questions/35160995/implement-foreign-trait-for-foreign-type). The normal solution for the latter problem: wrapping the foreign type in a new local type; doesn't work in our case because the REVM implementation depends on the original types.

In this PR, I was researching whether it's possible to split `L1ChainSpec` and its types in a standalone crate: `edr_chain_l1`, similar to the `edr_op` and `edr_generic`.

The PR proves that it's possible, but I wasn't happy with one big limitation:

`edr_chain_l1` depends on the "core crates", while tests of the "core crates" depend on `L1ChainType` (and its types). This creates a recursive dependency. This can partially be worked around by using `edr_chain_l1` as a dev-dependency and moving tests to the `tests/` directory.

I say partially because using types from `edr_chain_l1` that implement or use types from the "core crates" will result in obscure errors due to duplicate type definitions when used inside `#[cfg(test)]` scopes of the "core crates"'s source file.

I recommend that future research explores an alternative to mitigate or avoid this issue. In order of confidence:
1. Introduce separate crates for the `trait` definitions such as `RuntimeSpec` and `ProviderSpec`, such that we can break the recursive dependency.

    For example, this way `edr_provider` can depend on `edr_chain_l1` while `edr_chain_l1` depends on `edr_provider_spce`
2. Introduce a `TestChainSpec` and/or `MockChainSpec` to be used instead of `L1ChainSpec` in (some of) the tests.


Wodann:
A different strategy, used in #1067, was successful. Closing this PR.

PR #1010, opened 2025-07-18
fix: report code coverage

Wodann:
The bug consists of two parts:
- The TS test was [not resetting](https://github.com/NomicFoundation/edr/pull/1010/files#diff-cf0f4e94cf07b1b4788a10b23ed45b9855d18995b3c73768531ee1b9b629bc20R54) the `CoverageReporter` before each mocha test
- The `on_coverage_collected` callback was not being called upon completion of a Solidity test executor

This also changes the API of code coverage collection to allow asynchronous callbacks.

### DIFF
--- a/crates/edr_napi/src/config.rs
+++ b/crates/edr_napi/src/config.rs
@@ -41,9 +41,8 @@ pub struct CodeCoverageConfig {
     /// The callback receives an array of unique coverage hit markers (i.e. no
     /// repetition) per transaction.
     ///
-    /// # Safety
-    ///
-    /// Errors should not be thrown inside the callback.
+    /// Exceptions thrown in the callback will be propagated to the original
+    /// caller.
     #[napi(ts_type = "(coverageHits: Uint8Array[]) => Promise<void>")]
     pub on_collected_coverage_callback: JsFunction,
 }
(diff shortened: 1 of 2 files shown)


crates/edr_napi/src/config.rs
```diff
@@ -362,16 +367,23 @@ impl ObservabilityConfig {
...
-                                    sender.send(()).map_err(|_error| {
-                                        napi::Error::new(
-                                            napi::Status::GenericFailure,
-                                            "Failed to send result from on_collected_coverage_callback",
-                                        )
-                                    })
+                                .call_with_return_value(hits, ThreadsafeFunctionCallMode::Blocking, move |result: Promise<()>| {
+                                    // We spawn a background task to handle the async callback
+                                    runtime.spawn(async move {
+                                        let () = result.await?;
```
frangio, line 378: If the promise rejects, `sender` is dropped and `receiver.recv()` below will error and panic, which I don't think we want. We should probably ignore the promise result:
```suggestion
                                        let _ = result.await;
```

Wodann: Good catch!

frangio (approved):
New error propagation looks good.

PR #1012, opened 2025-07-22
chore: increase benchmark samples

frangio:
Since we're seeing benchmark CI jobs fail often, and this is attributed to high variability, this PR increases the number of samples for some test suites.

### DIFF
--- a/crates/tools/js/benchmark/src/solidity-tests.ts
+++ b/crates/tools/js/benchmark/src/solidity-tests.ts
@@ -52,7 +52,7 @@ export const FORGE_STD_SAMPLES = {
   [TOTAL_NAME]: DEFAULT_SAMPLES,
   StdCheatsTest: DEFAULT_SAMPLES,
   StdCheatsForkTest: 45,
-  StdMathTest: 45,
+  StdMathTest: 65,
   StdStorageTest: DEFAULT_SAMPLES,
   StdUtilsForkTest: DEFAULT_SAMPLES,
 };


frangio:
> Is the intention to make `StdUtilsForkTest` more reliable?

Yes, but also a few of the others I've seen fail:

```
| `StdCheatsTest` | `274` ms | `247` ms | `1.11` |
| `StdMathTest` | `16` ms | `12` ms | `1.33` |
```

> the failures for `StdUtilsForkTest` are typically caused by temporarily elevated response times from Alchemy, so running it more won't help.

This is a valid point. I wonder if there's some easy way to cache Alchemy requests from GitHub Actions until the mid-term fix?

agostbiro:
> Yes, but also a few of the others I've seen fail:
> | `StdCheatsTest` | `274` ms | `247` ms | `1.11` |
> | `StdMathTest` | `16` ms | `12` ms | `1.33` |

Ok let's bump the iterations for these, but I wouldn't change the iterations for `StdUtilsForkTest` as that might lead to more test failures due to rate limiting. As an alternative, we could also consider using a more robust statistics than comparing medians (e.g. throw away outliers and compute the mean).

> I wonder if there's some easy way to cache Alchemy requests from GitHub Actions until the mid-term fix?

It looks like there is a proxy server option for self-hosted runners (which we are using for the benchmark): https://docs.github.com/en/actions/how-tos/managing-self-hosted-runners/using-a-proxy-server-with-self-hosted-runners#configuring-a-proxy-server-using-environment-variables

But maybe the simplest solution would be to use a local Ethereum node in fork mode instead of Alchemy.

frangio:
Ok increased samples for `StdMathTest`.

PR #1013, opened 2025-07-22
feat: test execution time in nanoseconds

agostbiro:
Report Solidity test execution times in nanoseconds instead of milliseconds. This is needed for accurate comparisions with forge which reports nanoseconds when running `forge test --json`.

### DIFF
--- a/.changeset/tender-pillows-grab.md
+++ b/.changeset/tender-pillows-grab.md
@@ -1,0 +1,5 @@
+---
+"@nomicfoundation/edr": minor
+---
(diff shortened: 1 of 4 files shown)


frangio (approved):
I assume that backwards compatibility of `durationMs` is not a concer.

agostbiro:
> I assume that backwards compatibility of durationMs is not a concern.

Yeah I made it a minor bump, but only Hardhat is using it and will make the necessary there.

PR #1014, opened 2025-07-22
test: add test for matching Hardhat source names

agostbiro:
Adds a test for artifact id matching in `get*Code` cheatcodes to make sure that it matches Hardhat source names that are now prefixed with `project` or `npm` depending on whether the source file is a dependency. 

Also refactors `get_artifact_code` to allow testing the artifact id matching in isolation.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2994,6 +2994,7 @@ dependencies = [
  "dunce",
  "edr_common",
  "edr_defaults",
+ "edr_solidity",
  "eyre",
  "foundry-cheatcodes-spec 0.3.8",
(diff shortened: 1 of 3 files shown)


agostbiro:
> Also refactors `get_artifact_code` to allow testing the artifact id matching in isolation.

I realize now that refactoring `get_artifact_code` might be a bad idea while the backporting efforts are ongoing. Lmk pls @frangio

frangio:
> I realize now that refactoring `get_artifact_code` might be a bad idea while the backporting efforts are ongoing.

I haven't started yet so don't worry about it.

frangio (approved):

PR #1015, opened 2025-07-22
feat: add `allowInternalExpectRevert` config option

agostbiro:
Add `allowInternalExpectRevert` global config option which is needed by the [Uniswap v4 ](https://github.com/Uniswap/v4-core/blob/59d3ecf53afa9264a16bba0e38f4c5d2231f80bc/foundry.toml#L10) and [prb-math](https://github.com/PaulRBerg/prb-math/blob/aad73cfc6cdc2c9b660199b5b1e9db391ea48640/foundry.toml#L4) repos.

Only JS tests, no Rust tests as this will be tested once on the Rust side once we have inline config option support, and the JS tests give enough confidence until then.

### DIFF
--- a/.changeset/stupid-dryers-know.md
+++ b/.changeset/stupid-dryers-know.md
@@ -1,0 +1,5 @@
+---
+"@nomicfoundation/edr": patch
+---
+
+Add `allowInternalExpectRevert` config option
--- a/.changeset/stupid-dryers-know.md
+++ b/.changeset/stupid-dryers-know.md
@@ -2,4 +2,4 @@
 "@nomicfoundation/edr": patch
 ---
 
-Add `allowInternalExpectRevert` config option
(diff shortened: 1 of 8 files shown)


.changeset/stupid-dryers-know.md
```diff
@@ -0,0 +1,5 @@
+---
+"@nomicfoundation/edr": patch
+---
+
+Add `allowInternalExpectRevert` config option
```
frangio, line 5:
```suggestion
Add `allowInternalExpectRevert` config option to customize the behavior of the `expectRevert` cheatcode
```

crates/edr_napi/index.d.ts
frangio, line 1: It looks like `index.d.ts` and `index.js` were generated with the `op` feature flag, but so far we commit them into the repo without that flag.

frangio (approved):
`edr_napi` build outputs may need to be regenerated but otherwise looks good.

PR #1016, opened 2025-07-23
edr-0.12.0-next.0

frangio:
This PR releases `@nomicfoundation/edr` version `0.12.0-next.0`. This replaces `@ignored/edr-optimism`, as such the `op` feature will now be enabled on `main` (and possibly disabled in a later release).

This is done using Changesets' [prerelease mode](https://github.com/changesets/changesets/blob/main/docs/prereleases.md). Further prereleases can be done with the automated "Version packages" PR (#880). Once we want to move out of prerelease, we should run `changesets pre exit`.

I've marked all internal packages with `"private": true` so that Changesets doesn't think it has to version them (I saw some issues with that in testing).


nebasuke (approved):
Looks good, thanks Frangio!

PR #1018, opened 2025-07-24
fix: add dynamic base fee parameters to OP genesis block

Wodann:
https://github.com/NomicFoundation/hardhat/issues/7039 reported a regression when locally mining OP blocks. This happened because we were not setting the encoded dynamic base parameters as part of a genesis block's `extraData`.

I added a test to validate the panic.

To fix this, I had to make genesis block construction chain-specific.

In making that change, there were a lot of boilerplate changes that needed to happen to switch from the `ChainSpecT: RuntimeSpec` trait bound to `ChainSpecT: ProviderSpec<TimerT>, TimerT: Clone + TimeSinceEpoch` as well, which is why this PR ended up quite large.


crates/edr_evm/src/blockchain/local.rs
```diff
@@ -215,6 +108,33 @@ where
...
+    #[allow(clippy::too_many_arguments)]
+    pub fn new(
+        genesis_diff: StateDiff,
+        chain_id: u64,
+        hardfork: <ChainSpecT as ChainHardfork>::Hardfork,
+        options: GenesisBlockOptions,
+    ) -> Result<Self, <ChainSpecT as GenesisBlockFactory>::CreationError> {
+        let genesis_block = ChainSpecT::genesis_block(genesis_diff.clone(), hardfork, options)?;
+
+        Ok(unsafe {
```
frangio, line 132: `with_genesis_block_unchecked` requires that the block number be zero, but as far as I can tell nothing in this function guarantees that, so it should not be okay to call it. We should either assert that it is zero here, or propagate the requirement up by making something else unsafe (e.g. `GenesisBlockFactory` an `unsafe trait`). From previous conversations I assume this doesn't lead to memory unsafety, it's just about functional correctness, so I'd avoid propagating unsafe upwards and I'd also consider removing `unsafe` from `with_genesis_block_unchecked`.

frangio: Or we could use the safe `with_genesis_block`.

Wodann: I'll use the safe version instead

Wodann: Addressed by https://github.com/NomicFoundation/edr/pull/1018/commits/15cb535131ea94d1f9863ebcfa67fd8da0b145c3

frangio:
There was a doc error I tried to fix. Please review https://github.com/NomicFoundation/edr/pull/1018/commits/c66941a346c0f44222d722fa32adc0266d680be4.

frangio (approved):

Wodann:
> There was a doc error I tried to fix. Please review [c66941a](https://github.com/NomicFoundation/edr/commit/c66941a346c0f44222d722fa32adc0266d680be4).

Thanks!

PR #1020, opened 2025-07-24
chore: configure changesets/action pr title

frangio:
Ref: https://github.com/changesets/action#usage


frangio:
> 1. My interpretation of these changes is that it will fetch the version number from the `package.json` as it currently exists on `main`.

Yes, this was an error, fixed in 4dfbd845f01f.

> 2. Does the release workflow correctly bump pre-release versions?

We will know for sure when the PR updates, but it should, based on how `changesets version` behaves locally.

Wodann (approved):

PR #1022, opened 2025-07-25
fix: failing cheatcode tests

agostbiro:


crates/edr_solidity_tests/tests/testdata/default/cheats/Roll.t.sol
```diff
@@ -12,7 +12,7 @@ contract RollTest is DSTest {
         assertEq(block.number, 10, "roll failed");
     }
 
-    function testRollFuzzed(uint128 jump) public {
+    function testRollFuzzed(uint32 jump) public {
```
agostbiro, line 15: Block number was previously `uint256`, now it's `uint64`, hence the change

crates/edr_solidity_tests/tests/testdata/default/cheats/Warp.t.sol
```diff
@@ -12,7 +12,7 @@ contract WarpTest is DSTest {
         assertEq(block.timestamp, 10, "warp failed");
     }
 
-    function testWarpFuzzed(uint128 jump) public {
+    function testWarpFuzzed(uint32 jump) public {
```
agostbiro, line 15: Block number was previously `uint256`, now it's `uint64`, hence the change

frangio:
This unfortunately seems to have significant overlap with #1007 which was just merged, and which also backported a more recent version of the `gas_metering` part of the cheatcode inspector. Can you look at the conflicts before I review it?

crates/edr_solidity_tests/src/runner.rs
```diff
@@ -627,7 +627,7 @@ impl<
             counterexample: None,
             decoded_logs: decode_console_logs(&logs),
             logs,
-            kind: TestKind::Standard(gas.overflowing_sub(stipend).0),
+            kind: TestKind::Standard(gas.wrapping_sub(stipend)),
```
agostbiro, line 630: matches https://github.com/foundry-rs/foundry/blob/710a1584aae8e0f8ca8d5ba552632dc72381091e/crates/forge/src/result.rs#L526

frangio (approved):

PR #1023, opened 2025-07-25
fix: use rayon for test suite parallelism

agostbiro:
Using `tokio` for parallelism on the test suite level, but using `rayon` for parallelism on the test level caused two problems:

1. Panics from `rayon` threads weren't propagated which caused https://github.com/NomicFoundation/edr/pull/1022.
2. Lower performance (7.5% higher total test execution time in the `prb-math` repo with `tokio` over `rayon`)

Using `tokio` for parallelism on the test suite level was originally [introduced](https://github.com/NomicFoundation/edr/pull/474/commits/1cc21e8ecbfbf95cbb16eed444ec8029bb5c3e95) as a workaround for messages not being sent over a `tokio` channel when written to from a `rayon` thread. Since we're no longer using a `tokio` channel in the multi runner, but we're calling callbacks there directly, this is no longer a concern. I added a test called `StreamingResults` to make sure that test suite results are sent to JS as soon as the test suite is finished.


.changeset/afraid-roses-begin.md
```diff
@@ -0,0 +1,5 @@
+---
+"@nomicfoundation/edr": patch
+---
+
+Imrpoved parallelism in test suite execution for better error reporting and improved performance
```
frangio, line 5:
```suggestion
Improved parallelism in test suite execution for better error reporting and improved performance
```

crates/edr_solidity_tests/src/multi_runner.rs
```diff
@@ -414,38 +414,17 @@ impl<
...
-                                    result,
-                                },
-                            );
-                        })
-                        .await
-                        .expect("failed to join task");
-                    },
-                )
-                .await;
+        contracts.into_par_iter().for_each(|(id, contract)| {
```
frangio, line 419: My understanding is that this blocks until all of the array elements have been processed, whereas `handle.spawn` in the previous version would return control immediately while the task runs asynchronously.

Am I understanding that correctly? Is there a reason why that's not an issue?

This is my only question about the PR.

PR #1031, opened 2025-07-29
feat: include created contract address in trace

frangio:
For contract creation, we want to return the address of the created contract in the trace.


agostbiro (commented):
Hey thanks for this PR! I don't have the full context so feel free to disregard: would it make sense to remove the contract property and instead just have a mandatory address property and an optional label? This way the display is left up to the consumer.

PR #1033, opened 2025-07-29
refactor: remove deprecated JSON-RPC API methods

Wodann:
This PR removes deprecated JSON-RPC methods from EDR as they're no longer needed for HH3, according to [this design doc](https://www.notion.so/nomicfoundation/Change-JSON-RPC-methods-1e3578cdeaf58008b084c5370ce6aceb).

To continue supporting HH2, changes needed to be made to HH2, which are available in a companion commit: https://github.com/NomicFoundation/hardhat/pull/6120/commits/61490313b745a022e39ec55c85cba00ee3c85d22. It might be easier to review this commit than the patch file included in this PR.

I updated the `pnpm` patch to ensure that the updated HH2 code is used for the `./hardhat-tests`.

Since `hardhat_interval` is no longer available, I had to move the `Provider logs` tests using that method to EDR. It's now using a provider using a newly added `MockTime`, which is exclusively available in the `edr_napi` tests.

Future work:
- [ ] Release new version of EDR with these changes
- [ ] Upgrade EDR version in https://github.com/NomicFoundation/hardhat/pull/6120
- [ ] Do a pre-release of HH2 from the PR branch: https://github.com/NomicFoundation/hardhat/pull/6120
- [ ] Upgrade HH2 dependency in EDR to pre-release version


crates/edr_napi/src/mock/time.rs
```diff
@@ -0,0 +1,101 @@
...
+
+        self.inner.add_seconds(seconds);
+        Ok(())
+    }
+}
+
+#[doc = "Creates a provider with a mock timer."]
+#[doc = "For testing purposes."]
+#[napi]
+pub fn create_provider_with_mock_timer(
```
frangio, line 45: Is this function supposed to be shipped with the library?

frangio, line 45: Would it be possible to define a helper to be used by this and the un-mocked `create_provider`? If not, is there a possibility that this function would get out of sync with that one?

Wodann: The whole `mod mock` (one level up) is only included if the `test-mock` feature flag is enabled. As a result, this module will never be compiled.

crates/edr_napi/src/provider.rs
```diff
@@ -45,6 +45,32 @@ impl Provider {
 
 #[napi]
 impl Provider {
+    #[doc = "Adds a compilation result to the instance."]
+    #[doc = ""]
+    #[doc = "For internal use only. Support for this method may be removed in the future."]
```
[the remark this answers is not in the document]
Wodann: I thought about doing that but decided to leave it as this since we're still doing `0.x.y` releases. Backwards incompatible changes only require a minor version bump, so any time we change this function we just do a minor version bump like for other API changes.

That in combination with the comment seemed sufficient to notify the user that this function API is "fragile".

Would you still prefer an underscore name?

crates/edr_napi/src/provider.rs
```diff
@@ -45,6 +45,32 @@ impl Provider {
...
+    #[napi(catch_unwind)]
+    pub async fn add_compilation_result(
+        &self,
+        solc_version: String,
+        compiler_input: serde_json::Value,
+        compiler_output: serde_json::Value,
```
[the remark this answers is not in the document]
Wodann: Performance should stay equivalent to prior to this PR, but we're not focusing on improving performance in this PR yet. For now, the idea is to move to an internal API before the beta (to avoid breaking changes in the JSON-RPC afterwards) and then iterate on this function when we have more time.

crates/edr_napi/src/provider.rs
```diff
@@ -45,6 +45,32 @@ impl Provider {
...
+    ) -> napi::Result<bool> {
+        let compiler_input = serde_json::from_value(compiler_input)
+            .map_err(|error| napi::Error::from_reason(error.to_string()))?;
+
+        let compiler_output = serde_json::from_value(compiler_output)
+            .map_err(|error| napi::Error::from_reason(error.to_string()))?;
```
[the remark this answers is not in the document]
Wodann: Yes, I can do that.

.github/workflows/edr-ci.yml
```diff
@@ -157,6 +159,44 @@ jobs:
...
           key: edr-ts-rpc-cache-v1-${{ matrix.os }}-${{ hashFiles('**/edr-cache/**/*') }}
 
+  test-edr-ts-integration:
+    name: Run TS integration tests (${{ matrix.os }})
+    runs-on: ${{ matrix.os }}
+    needs: check-edr
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
```
Wodann, line 169: I modified these lines to ensure:
- The matrix is set, as required in the last step
- I added `runs-on` to ensure these tests are run on all operating systems
- I added `fail-fast: fast` to ensure all OSes complete
- I modified `name` to include `matrix.os`
- I added `needs: check-edr` to prevent this from running if that job fails

.github/workflows/edr-ci.yml
```diff
@@ -157,6 +159,44 @@ jobs:
             **/edr-cache
           key: edr-ts-rpc-cache-v1-${{ matrix.os }}-${{ hashFiles('**/edr-cache/**/*') }}
 
+  test-edr-ts-integration:
```
Wodann, line 162: I renamed this job and moved it closer to the other test jobs to make it easier to cross-reference job description patterns

crates/edr_napi/src/mock/time.rs
Wodann: After discussing, I decided to revert https://github.com/NomicFoundation/edr/pull/1033/commits/795457ddddc4de21583a15c7f4bb8129d3b005d3

In a follow-up PR I'm making some changes that will allow us to more cleanly share code.

frangio (approved):
LGTM

Wodann:
PR didn't make changes to Solidity test runner. https://github.com/NomicFoundation/edr/actions/runs/16678709057/job/47211957562?pr=1033 seems to be a false positive

PR #1037, opened 2025-07-31
fix: mitigate EVM sandbox escape through cheatcodes

popescuoctavian:
This PR modifies permission interface to separate file and directory permissions to prevent users from unawarely expose themselves by sharing access to important project directories. We are changing the config interface and adding more detailed user-facing documentation to mark the danger posed by giving write access to directories, according to the [design spec](https://www.notion.so/nomicfoundation/Mitigate-EVM-sandbox-escape-through-cheatcodes-238578cdeaf5804c8e69d29a54ee39b5?source=copy_link), requiring extra steps to give full directory access.


popescuoctavian:
Opened #1038 as a follow up to this PR, since unexpected behavior of `is_path_allowed` (i.e. comparing un-normalized paths to normalized paths) was seen during testing.

crates/edr_solidity_tests/tests/testdata/default/cheats/Fs.t.sol
```diff
@@ -107,40 +118,6 @@ contract FsTest is DSTest {
         vm.removeFile("/etc/hosts");
     }
 
-    function testWriteLineFoundrytoml() public {
```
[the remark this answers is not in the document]
popescuoctavian: Currently there's no way to exclude files, but it should be easy enough to implement in the future if needed (similar to how file permissions are performed now).

It may be enough to just insert `Foundry.toml` in the permissions list with `FsAccessPermission::None`, which would mark the file as "no access" while still being able to access the other files. But we'd need to assess if there are ways to go around this with any of the cheatcodes (i.e. having write access to the directory and using a cheatcode to indirectly overwrite the excluded file).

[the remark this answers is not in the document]
popescuoctavian: I've added a simple test case for this. We're currently not exposing `FsAccessPermission::None` through `napi`, so this cannot be achieved by users with the current interface.

crates/tools/js/benchmark/src/solidity-tests.ts
```diff
@@ -346,7 +346,7 @@ async function createSolidityTestsInput(repoPath: string) {
   solidityTestsConfig.rpcCachePath = RPC_CACHE_PATH;
   const rootPermission = {
     path: repoPath,
-    access: FsAccessPermission.ReadWrite,
+    access: FsAccessPermission.DangerouslyReadWriteDirectory,
```
agostbiro, line 349: This is probably not necessary. I'll fix it in the new benchmark PR.

crates/foundry/cheatcodes/src/fs_permissions.rs
```diff
@@ -191,70 +237,119 @@ impl FsAccessPermission {
...
+            (FsAccessPermission::DangerouslyWriteDirectory, FsAccessKind::Write) => true,
             _ => false,
         }
     }
 }
 
-impl FromStr for FsAccessPermission {
-    type Err = String;
+#[cfg(test)]
+mod tests {
```
agostbiro, line 253: I'd move these tests to `crates/foundry/cheatcodes/src/config.rs` and change the tests to use `CheatsConfig::ensure_path_allowed`. The reason for this is that the paths in these tests are relative, but `CheatsConfig::new` joins the paths in `FsPermissions` with the root, and the tested paths are normalized in `CheatsConfig::ensure_path_allowed`.

agostbiro (changes requested):
Hey thanks LGTM overall, I made some suggestions to improve the tests

crates/foundry/cheatcodes/src/fs_permissions.rs
popescuoctavian: Good idea, moved them.

crates/edr_solidity_tests/tests/testdata/default/cheats/Fs.t.sol
```diff
@@ -107,40 +118,6 @@ contract FsTest is DSTest {
...
-        vm.writeLine("foundry.toml", "\nffi = true\n");
-
-        vm._expectCheatcodeRevert();
-        vm.writeLine("./foundry.toml", "\nffi = true\n");
-
-        vm._expectCheatcodeRevert();
-        vm.writeLine("./Foundry.toml", "\nffi = true\n");
-    }
-
-    function testWriteFoundrytoml() public {
```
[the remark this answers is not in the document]
popescuoctavian: Copied some tests from `edr_solidity_tests/tests/testdata/default/cheats/Fs.t.sol` and added tests for `cli.js`.

crates/foundry/cheatcodes/src/fs_permissions.rs
```diff
@@ -191,70 +237,14 @@ impl FsAccessPermission {
     pub fn is_granted(&self, kind: FsAccessKind) -> bool {
         #[allow(clippy::match_same_arms)]
         match (self, kind) {
-            (FsAccessPermission::ReadWrite, _) => true,
+            (FsAccessPermission::ReadWriteFile, FsAccessKind::Read | FsAccessKind::Write) => true,
             (FsAccessPermission::None, _) => false,
```
agostbiro, line 241: Could you make the change `FsAccessKind::Read | FsAccessKind::Write` here as well?

crates/foundry/cheatcodes/src/fs_permissions.rs
```diff
@@ -191,70 +237,14 @@ impl FsAccessPermission {
...
-            (FsAccessPermission::Read, FsAccessKind::Read) => true,
-            (FsAccessPermission::Write, FsAccessKind::Write) => true,
+            (FsAccessPermission::ReadFile, FsAccessKind::Read) => true,
+            (FsAccessPermission::WriteFile, FsAccessKind::Write) => true,
+            (FsAccessPermission::DangerouslyReadWriteDirectory, _) => true,
```
agostbiro, line 244: Could you make the change `FsAccessKind::Read | FsAccessKind::Write` here as well?

js/integration-tests/solidity-tests/test/fs_permissions.ts
```diff
@@ -0,0 +1,72 @@
...
+
+describe("Fs permissions tests", () => {
+  let testContext: TestContext;
+
+  before(async () => {
+    testContext = await TestContext.setup();
+  });
+
+  it("file + directory permissions", async function () {
+    const results = await runAllSolidityTests(
```
agostbiro, line 15: This works well, just for you to know there is a slightly simpler way of running the tests with `testContext.runTestsWithStats`: https://github.com/NomicFoundation/edr/blob/a5cc346fb982ca4c4ef74eeb32ae12a344a95790/js/integration-tests/solidity-tests/test/unit.ts#L368-L378

agostbiro (approved):
Thanks LGTM! Just one small request to replace every `_` with `FsAccessKind::Read | FsAccessKind::Write`

crates/foundry/cheatcodes/src/fs_permissions.rs
popescuoctavian: For `::None` I assumed we want to keep `_` since it makes sense to match anything

crates/foundry/cheatcodes/src/fs_permissions.rs
popescuoctavian: Oops. missed this one, adding now!

js/integration-tests/solidity-tests/test/fs_permissions.ts
popescuoctavian: Hmm, I thought it would be harder to specify the test runner config with `runTestsWithStats`, but just saw that there's a parameter for that 😄 Thanks!

crates/foundry/cheatcodes/src/fs_permissions.rs
popescuoctavian: That makes sense, thanks for the explanation! Added it in both places.

PR #1040, opened 2025-07-31
refactor: make SubscriptionEvent chain-agnostic

Wodann:
This PR makes subscription events chain-agnostic in `edr_napi_core`. This is the first step to simplifying the N-API code for constructing a provider.

Currently, the following happens:
1. Find a factory
2. Use the factory to construct a chain-specific builder object (this was required due to the chain-specific `SubscriptionEvent`)
3. Use the builder to construct a provider

After this, it should be possible to combine steps 2 & 3


crates/edr_napi_core/src/subscription.rs
```diff
@@ -1,23 +1,102 @@
...
+///
+/// An alternative would be to use `serde_json::Value` as an intermediate
+/// representation, but that would require an additional conversion step:
+///
+/// 1. Convert the value to `serde_json::Value`.
+/// 2. Send the `serde_json::Value` to the `ThreadsafeFunction`.
+/// 3. Convert the `serde_json::Value` to a JavaScript value using
+///    `napi::Env::to_js_value`.
+/// ```
+pub type ToJsValueFn = dyn FnOnce(&napi::Env) -> napi::Result<JsUnknown>;
```
frangio, line 51: I found it hard to understand what this type was. How about a name like `JsValueFromEnvFn`?

crates/edr_napi_core/src/subscription.rs
```diff
@@ -1,23 +1,102 @@
...
+        let edr_provider::SubscriptionEvent { filter_id, result } = event;
+
+        Self {
+            filter_id,
+            result: SubscriptionEventData::new::<_, RpcBlockT, _>(result),
+        }
+    }
+}
+
+/// Type alias for a function that converts a value to a JavaScript value.
```
frangio, line 37:
```suggestion
/// Type alias for a closure that produces a JavaScript value given a `napi::Env`.
```

crates/edr_napi_core/src/subscription.rs
```diff
@@ -1,23 +1,102 @@
...
-pub struct Callback<ChainSpecT: ProviderSpec<CurrentTime> + 'static> {
-    inner: ThreadsafeFunction<SubscriptionEvent<ChainSpecT>, ErrorStrategy::Fatal>,
+/// A chain-agnostic version of [`edr_provider::SubscriptionEvent`].
+pub struct SubscriptionEvent {
+    pub filter_id: U256,
+    pub result: SubscriptionEventData,
+}
+
+impl SubscriptionEvent {
+    pub fn new<BlockT, RpcBlockT, SignedTransactionT>(
```
frangio, line 19: Can this be a `From` impl?

crates/edr_napi_core/src/subscription.rs
```diff
@@ -1,23 +1,102 @@
...
+/// A chain-agnostic version of [`edr_provider::SubscriptionEventData`].
+pub enum SubscriptionEventData {
+    Logs(Vec<LogOutput>),
+    /// A function that converts a [`BlockAndTotalDifficulty`] to a JS value.
+    NewHeads(Box<ToJsValueFn>),
+    NewPendingTransactions(B256),
+}
+
+impl SubscriptionEventData {
+    pub fn new<BlockT, RpcBlockT, SignedTransactionT>(
```
frangio, line 62: Same question: Can this a `From` impl?

crates/edr_napi_core/src/subscription.rs
Wodann: It doesn't get it from the `napi::Env` though. It just uses it. How about `JsValueConstructor`?

frangio: That sounds good.

Wodann: Addressed in https://github.com/NomicFoundation/edr/pull/1040/commits/375b694bb6a684d25adfc612eeb5c3c8b1eb1b04

crates/edr_napi_core/src/subscription.rs
Wodann: Addressed in https://github.com/NomicFoundation/edr/pull/1040/commits/375b694bb6a684d25adfc612eeb5c3c8b1eb1b04

Wodann:
This PR didn't affect Solidity tests, so the CI failure is likely a false positive.

PR #1042, opened 2025-08-01
test: reduce memory consumption of issue 543 test by 40%

Wodann:
When my devcontainer was short on memory, I ran into a problem with the "issue 543" TS test "freezing" because the system would start writing RAM to disk and take forever.

By reducing the memory usage by 40%, we're still causing a crash in `JSON.stringify` but minimising the time & memory resources used.

I also lowered the timeout, as the test now runs quicker.


popescuoctavian (approved):
LGTM, thanks for taking care of it!

PR #1044, opened 2025-08-01
ci: reduce benchmark time by 80%

Wodann:
Following https://github.com/NomicFoundation/edr/pull/1012, Solidity test benchmark run times [increased](https://github.com/NomicFoundation/edr/actions/runs/16680238634/job/47217929068?pr=1042) to over an hour. The culprits are:

- `[TOTAL_NAME]` increased from 5 -> 25 (this task takes about 70 seconds)
- `StdStorageTest` increased from 5 -> 25 (this task takes about 69 seconds)

To the best of my knowledge, these two tests never caused false positives, so I propose returning them to their original sample count.


frangio (approved):
I've removed the default number entirely since the number needs to be tuned according to each test's runtime.

PR #1048, opened 2025-08-07
chore: warnings as errors in only in check job

frangio:
If the code compiles with warnings, the global `-Dwarnings` flag will cause all test jobs to fail, where I think it should be sufficient to cause the check job to fail.

### DIFF
--- a/.github/workflows/edr-ci.yml
+++ b/.github/workflows/edr-ci.yml
@@ -9,9 +9,6 @@ on:
       - "**"
   workflow_dispatch:
 
-env:
-  RUSTFLAGS: -Dwarnings
-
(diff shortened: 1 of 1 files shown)


.github/workflows/edr-ci.yml
```diff
@@ -28,9 +27,9 @@ jobs:
       - uses: taiki-e/install-action@cargo-hack
 
       - name: Cargo check no default features for all crates
-        run: cargo check --all --no-default-features
+        run: cargo check --workspace --all-targets --no-default-features
```
frangio, line 30: `--all` is deprecated and replaced by `--workspace`.

PR #1050, opened 2025-08-07
chore: disable test that times out

frangio:


frangio:
The test got stuck on windows 😅

PR #1051, opened 2025-08-08
chore: change error description when encountering an invalid Error(string)

### LINKED ISSUE #972: Change to description on revert with invalid error messge
In Hardhat Ignition we have an integration test suite that has a failing test (likely for many months). It is failing, we think, because an expected error message has changed.

Given the following contract:

```solidity
contract RevertExample {
    function revertWithReasonMessage() public pure {
        revert("reason");
    }

    function revertWithInvalidErrorMessage() public pure {
        // This starts with the right signature and then
        // invalid data wrt the ABI encoding
        bytes memory x = hex"08c379a0123456";
        assembly {
            revert(add(x, 32), 7)
        }
    }
}
```

Then invocation of the function `revertWithInvalidErrorMessage`, previously reverted with:

```
VM Exception while processing transaction: reverted with reason string ''
```

Instead we are getting:

```
Expected return data to be a Error(string) and contain a valid string
```

There is a reproduction repo here: https://github.com/kanej/revert-with-invalid-error-message

popescuoctavian:
This PR changes the error description when encountering an invalid error message, prepending the expected message `VM Exception while processing transaction:`.

It also changes the message for invalid panic decoding.

closes #972


crates/edr_napi/src/trace/return_data.rs
```diff
@@ -57,10 +57,17 @@ impl ReturnData {
             return Ok(String::new());
         }
 
+        if !self.is_error_return_data() {
+            return Err(napi::Error::new(
+                napi::Status::InvalidArg,
+                "VM Exception while processing transaction: Expected return data to be a Error(string)",
```
popescuoctavian, line 63: Note: During my testing I could not hit this case with the current implementation in HH - when using an invalid error, the message was `VM Exception while processing transaction: reverted with an unrecognized custom error`.

frangio: Was this tested with HH3 or HH2?

popescuoctavian:
@frangio I was planning to add a test after this releases and is integrated into HH, to test E2E.

frangio (approved):
Thanks!

PR #1074, opened 2025-09-10
build: only run setup after devcontainer creation

Wodann:
Prevents the setup script from running every time you start a devcontainer. It should suffice to only run it after devcontainer creation.


Wodann:
> For me sometimes rust-analyzer starts misbehaving and I reload the VSCode window to restart it

In case you weren't aware of the command: there is a command to restart the Rust Analyzer server. That might also solve the problem

PR #1090, opened 2025-09-18
fix json-rpc `eth_feeHistory` method

anaPerezGhiglia:
## Context
[Ethereum JSON-RPC specifies](https://ethereum.github.io/execution-apis/api-documentation/) that all `eth_feeHistory` method parameters are required

Before these changes, the third parameter `rewardPercentiles` was optional in EDR. This was probably because [Alchemy documentation](https://www.alchemy.com/docs/node/ethereum/ethereum-api-endpoints/eth-fee-history) states that the parameter is optional - even if now it's rejecting the requests that do specify it

## Changes
- Change rpc `eth_feeHistory` method definition to receive a `Vec<f64>` instead of an `Option<Vec<f64>>`
- Bonus track: [fixed ts type check failure](https://github.com/NomicFoundation/edr/commit/e3255dbe47df1f2464dac9c1b5584a6258ec5f8e)

### DIFF
--- a/.changeset/hungry-beans-agree.md
+++ b/.changeset/hungry-beans-agree.md
@@ -1,0 +1,6 @@
+---
+"hardhat-edr-tests": minor
+"@nomicfoundation/edr": minor
+---
+
+make all parameters of `eth_feeHistory` rpc call & provider method call required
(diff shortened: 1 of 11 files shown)


agostbiro (changes requested):
Thanks LGTM overall, just a small change needed in the caching logic.

agostbiro (approved):
Thanks for the changes! LGTM

PR #1109, opened 2025-10-03
chore: use os pinned versions in CI workflows 

anaPerezGhiglia:
Using `latest` tags is a bad practice that can lead to unexpected changes and opens the door to security risks.

This PR updates CI jobs to use pinned versions of ubuntu, macos and windows. The version mapping was done considering [this](https://github.com/actions/runner-images/blob/41a7cee3c1afb1b3aad170ffae8d85bf262db862/README.md) is the current github runner-images documentation

I also updated jobs that were running in `ubuntu-22.04` to `ubuntu-24.04`

### DIFF
--- a/.github/workflows/check-changeset-added.yml
+++ b/.github/workflows/check-changeset-added.yml
@@ -17,7 +17,7 @@ on:
 jobs:
   check-if-changeset:
     name: Check that PR has a changeset
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     # don't run this check in the changesets PR
     if: github.head_ref != 'changeset-release/main'
     steps:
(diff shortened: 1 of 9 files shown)


popescuoctavian (approved):
Agreed on not using `latest` versions. Thanks!

As a follow up, we should probably follow the instructions from [here](https://docs.github.com/en/subscriptions-and-notifications/get-started/configuring-notifications#configuring-your-watch-settings-for-an-individual-repository) to keep up with new releases. Not sure if we can be notified about new releases as a team, or we have to subscribe individually. Otherwise I have a feeling that we'll never remember to manually update them unless something breaks.

anaPerezGhiglia:
The actions that are failing are not related to these changes. Those tests were failing before this PR, so I'm considering this PR good to merge

PR #1110, opened 2025-10-07
fix: use hardfork-specific L1BlockInfo for dry_run

Wodann:
While adding support for the Isthmus OP stack hardfork in Hardhat 3, the Hardhat team observed [an EDR panic in their tests](https://github.com/NomicFoundation/hardhat/actions/runs/18316863161/job/52159414667?pr=7522).

This PR resolves panics that occurred due to the prior usage of a default `L1BlockInfo` for "dry runs" (e.g. `eth_call`). Now, an EVM instance is always constructed using block-specific `L1BlockInfo`.

### DIFF
--- a/.changeset/social-moments-move.md
+++ b/.changeset/social-moments-move.md
@@ -1,0 +1,5 @@
+---
+"@nomicfoundation/edr": patch
+---
+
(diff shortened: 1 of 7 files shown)


popescuoctavian (approved):
LGTM, thanks!

Issue #1112, opened 2025-10-08
Revisit OP RPC getTransactionReceipt tests after Alchemy investigation


popescuoctavian:
Some `getTransactionReceipt` tests are failing because of changes in the RPC response from Alchemy:
```
transaction_and_receipt_pre_bedrock
deposit_transaction_and_receipt_regolith
deposit_transaction_and_receipt_canyon
deposit_transaction_and_receipt_ecotone
```
Revisit the tests after the investigation is completed and revert/modify as needed.

popescuoctavian:
Alchemy investigation points to differences between Geth and Reth, with Geth not reporting l1 fees while Reth does. This means that the changes are expected. Closing the issue.

PR #1113, opened 2025-10-08
fix: update OP Alchemy RPC test expectations

popescuoctavian:
This PR modifies the test expectations for the tests that started failing because of changes in the RPC response from Alchemy. This is done to fix the CI failures until Alchemy investigates the issue.

Created #1112 as a follow up item.

### DIFF
--- a/crates/edr_op/tests/integration/rpc.rs
+++ b/crates/edr_op/tests/integration/rpc.rs
@@ -14,7 +14,6 @@ use edr_primitives::{b256, HashMap, B256};
 use edr_rpc_eth::client::EthRpcClient;
 use edr_test_utils::env::get_alchemy_url;
 use edr_transaction::TransactionType as _;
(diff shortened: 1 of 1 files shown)


crates/edr_op/tests/integration/rpc.rs
```diff
@@ -85,37 +84,7 @@ async fn block_with_deposit_transaction() -> anyhow::Result<()> {
     Ok(())
 }
 
-#[tokio::test(flavor = "multi_thread")]
```
Wodann, line 88: Question: Does this transaction no longer exist in Alchemy at all?

popescuoctavian: Yes, RPC call reports "not found".

Wodann (approved):
Approved, pending CI success.

PR #1115, opened 2025-10-13
feat: collect gas reports for Solidity tests

popescuoctavian:
This PR adds support for the collection of gas reports in Solidity tests.

### Changes
- A new parameter `generateGasReports` is added to `SolidityTestRunnerConfigArgs` which enables the caller of `runSolidityTests` to request the generation and collection of gas reports for the test run.
- `runSolidityTests` now returns a Promise which resolves to a `SolidityTestResult`. The gas report is part of that object.
- Gas report code was moved to a new crate `edr_gas_report`, which is shared by both the provider and soltests implementations.
- `gas_reports` module/file were renamed to `gas_report` for consistency.

### DIFF
--- a/crates/edr_solidity_tests/src/multi_runner.rs
+++ b/crates/edr_solidity_tests/src/multi_runner.rs
@@ -284,9 +284,11 @@ impl<
         contract: &TestContract,
         fork: Option<CreateFork<BlockT, TransactionT, HardforkT>>,
         filter: &dyn TestFilter,
-        mut gas_report: Option<&mut crate::gas_report::GasReport>,
         handle: &tokio::runtime::Handle,
-    ) -> SuiteResult<HaltReasonT> {
+    ) -> (
+        SuiteResult<HaltReasonT>,
+        Option<crate::gas_report::GasReport>,
+    ) {
         let identifier = artifact_id.identifier();
         let mut span_name = identifier.as_str();
 
@@ -354,6 +356,10 @@ impl<
             );
         let mut r = runner.run_tests(filter, &self.test_options, handle);
 
+        let mut gas_report = self
+            .generate_gas_report
+            .then(crate::gas_report::GasReport::default);
+
         if self.include_traces != IncludeTraces::None {
             let mut decoder = CallTraceDecoderBuilder::new().build();
             let mut trace_identifier = TraceIdentifiers::new().with_local(&self.known_contracts);
@@ -411,7 +417,7 @@ impl<
         }
         debug!(duration=?r.duration, "executed all tests in contract");
 
-        r
+        (r, gas_report)
     }
 
     /// Executes _all_ tests that match the given `filter`.
@@ -491,18 +497,8 @@ impl<
             .into_par_iter()
             .map(|(id, contract)| {
                 let _guard = tokio_handle.enter();
-                let mut gas_report = self
-                    .generate_gas_report
-                    .then(crate::gas_report::GasReport::default);
-
-                let result = self.run_tests(
-                    &id,
-                    &contract,
-                    fork.clone(),
-                    filter.as_ref(),
-                    gas_report.as_mut(),
-                    &tokio_handle,
-                );
+                let (result, gas_report) =
+                    self.run_tests(&id, &contract, fork.clone(), filter.as_ref(), &tokio_handle);
 
                 on_test_suite_completed_fn(SuiteResultAndArtifactId {
                     artifact_id: id,


crates/edr_gas_report/src/gas_report.rs
```diff
@@ -1,7 +1,13 @@
+use std::fmt::Display;
```
Wodann: Polish: Since the crate only consists of one file, feel free to make this a single-file crate with just a `lib.rs`; i.e. move `gas_report.rs` -> `lib.rs`.

crates/edr_solidity_tests/src/gas_report.rs
```diff
@@ -115,17 +115,36 @@ impl GasReport {
...
             let should_include = !name.test_function_kind().is_known();
-            if should_include {
+            if is_setup {
+                contract_info.status = if trace.is_error() {
+                    GasReportExecutionStatus::Halt
+                } else if trace.is_revert() {
+                    GasReportExecutionStatus::Revert
+                } else {
+                    GasReportExecutionStatus::Success
+                };
```
Wodann: Polish: Can the code duplication between these lines and lines 138-146 be easily avoided?

crates/edr_solidity_tests/src/multi_runner.rs
```diff
@@ -189,6 +196,13 @@ impl<
 
         let test_options: TestOptions = TestOptions { fuzz, invariant };
 
+        if generate_gas_report {
```
Wodann: Suggestion: It would be good to document these side-effects in the solidity test runner config's API docs

crates/edr_solidity_tests/src/multi_runner.rs
```diff
@@ -263,6 +278,7 @@ impl<
         contract: &TestContract,
         fork: Option<CreateFork<BlockT, TransactionT, HardforkT>>,
         filter: &dyn TestFilter,
+        gas_report: &mut Option<crate::gas_report::GasReport>,
```
Wodann: Suggestion(blocking): If the `Option` shouldn't be mutable, this should be an `Option<&mut ...>`.

Receiving a `&mut Option` means that the `Option` itself can be changed; e.g. from `Some` to `None`.
Receiving an `Option<&mut>` means that you optionally receive a mutable reference to the gas report. If you receive some argument, that argument can be modified.

From a usage perspective of the API, this is the difference:

```
fn foo(&mut Option<u32>);

let mut a = Some(0);
foo(&mut a);

foo(&mut None);
```
vs
```
fn foo(Option<&mut u32>);

let mut a = 3;
foo(Some(&mut a));
foo(None);
```

crates/edr_solidity_tests/src/multi_runner.rs
```diff
@@ -334,9 +350,13 @@ impl<
...
         if self.include_traces != IncludeTraces::None {
             let mut decoder = CallTraceDecoderBuilder::new().build();
+            let mut trace_identifier = TraceIdentifiers::new().with_local(&self.known_contracts);
 
             for result in r.test_results.values_mut() {
-                if result.status.is_success() && self.include_traces != IncludeTraces::All {
+                if !self.generate_gas_report
+                    && result.status.is_success()
+                    && self.include_traces != IncludeTraces::All
+                {
```
Wodann: If `self.generate_gas_report` always sets `self.include_traces = IncludeTraces::All`, is it necessary to check `!self.generate_gas_report` here?

crates/edr_solidity_tests/tests/it/fuzz.rs
```diff
@@ -175,3 +177,55 @@ async fn test_persist_fuzz_failure() {
...
+        .functions
+        .contains_key("incrementBy(uint256)"));
+
+    let increment_by_reports = sample_contract_report
+        .functions
+        .get("incrementBy(uint256)")
+        .unwrap();
+
+    assert!(!increment_by_reports.is_empty());
+    assert!(increment_by_reports.iter().any(|r| r.gas > 0));
```
Wodann: Question: Should only one of the reports have a gas usage > 0 or `.all(..)`?

crates/edr_solidity_tests/tests/it/helpers/config.rs
```diff
@@ -130,7 +131,12 @@ impl<
     }
 
     /// Executes the test runner
-    pub async fn test(self) -> BTreeMap<String, SuiteResult<HaltReasonT>> {
+    pub async fn test(
+        self,
+    ) -> (
+        SolidityTestResult,
+        BTreeMap<String, SuiteResult<HaltReasonT>>,
+    ) {
```
Wodann: Suggestion(non-blocking): consider using a named struct to self-document the fields of the return type

crates/edr_solidity_tests/tests/it/invariant.rs
```diff
@@ -971,3 +977,42 @@ async fn test_no_reverts_in_counterexample() {
...
+        .contains_key("set0(int256)"));
+    assert!(invariant_breaker_report
+        .functions
+        .contains_key("set1(int256)"));
+
+    let flag1_reports = invariant_breaker_report.functions.get("flag1()").unwrap();
+    assert!(!flag1_reports.is_empty());
+    assert!(flag1_reports
+        .iter()
+        .any(|r| r.gas > 0 && r.status == GasReportExecutionStatus::Success));
```
Wodann: Question: Similar to my other question, are we only interested in one case that reports gas usage? What about `Revert` and `Halt`?

js/helpers/src/index.ts
```diff
@@ -64,6 +65,49 @@ export async function runAllSolidityTests(
   });
 }
 
+/**
+ * Run all the given solidity tests and returns the whole results after finishing, including the gas report.
+ */
+export async function runAllSolidityTestsWithGasReport(
```
Wodann: Suggestion: Shouldn't we modify the original function? That one should also wait for the `Promise` to resolve, rather than waiting until `resultsFromCallback.length === testSuites.length`, since we now have a promise-based implementation rather than a function that kicks off a background processes and is informed using callback-driven results.

js/integration-tests/solidity-tests/test/testContext.ts
```diff
@@ -104,17 +106,36 @@ export class TestContext {
...
+        this.edrContext,
+        chainType,
+        this.artifacts,
+        testContracts,
+        this.tracingConfig,
+        {
+          ...this.defaultConfig(chainType),
+          ...config,
+        }
+      );
```
Wodann: Suggestion: As mentioned above, it feels like this duplication can be avoided by modifying the original function to return gas reports.

crates/edr_solidity_tests/src/multi_runner.rs
popescuoctavian: Hmm, very good point, I did not think of that. I did the changes, but I'm not sure if there's a better way to do it (without also having`mut gas_report: ...`.

crates/edr_solidity_tests/src/multi_runner.rs
popescuoctavian: That is correct, removed it, thanks for catching this!

crates/edr_solidity_tests/src/multi_runner.rs
```diff
@@ -373,11 +420,14 @@ impl<
     pub async fn test_collect(
         self,
         filter: impl TestFilter + 'static,
-    ) -> BTreeMap<String, SuiteResult<HaltReasonT>> {
+    ) -> (
+        SolidityTestResult,
+        BTreeMap<String, SuiteResult<HaltReasonT>>,
+    ) {
```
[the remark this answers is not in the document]
popescuoctavian: Added `SolidityTestsRunResult` which encapsulates both of them. Naming is also consistent with what we use in TS helpers.

js/helpers/src/index.ts
```diff
@@ -53,13 +62,15 @@ export async function runAllSolidityTests(
...
-            resolve(resultsFromCallback);
-          }
+          tryResolve();
         }
       )
+      .then((result) => {
+        testResult = result;
+        isTestComplete = true;
+        tryResolve();
+      })
```
popescuoctavian: We still need to check if all the callbacks have been processed, otherwise we sometimes end up resolving the promise without all the suite results.

I'm not sure if there is a better way to do this, so I'm very open to suggestions 😄

crates/edr_gas_report/src/gas_report.rs
```diff
@@ -122,6 +128,105 @@ impl GasReport {
     }
 }
 
+impl Display for GasReport {
```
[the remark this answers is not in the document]
Wodann: I am mostly concerned about whether the newly added dependency adds attack surface, but if you vetted it from a security perspective it's alright to keep it. I'll trust your judgement

crates/edr_solidity_tests/src/multi_runner.rs
```diff
@@ -432,16 +484,33 @@ impl<
             find_time,
         );
 
-        contracts.into_par_iter().for_each(|(id, contract)| {
```
agostbiro: It's important that we keep the parallel execution of test suites for performance. Could you put that back please?

Sharing a mutable `GasReport` won't work if parallelized. I can think of two ways to handle this:

1. Create a `GasReport` for each test suite and then merge them at the end.
2. Wait until all test suites have finished before generating the gas report.

I think whatever is simplest to implement is ok.

crates/edr_solidity_tests/src/multi_runner.rs
```diff
@@ -349,16 +371,43 @@ impl<
...
                 for (_, arena) in &mut result.traces {
-                    let mut trace_identifier =
-                        TraceIdentifiers::new().with_local(&self.known_contracts);
                     decoder.identify(arena, &mut trace_identifier);
                     tokio::task::block_in_place(|| {
                         handle.block_on(decode_trace_arena(arena, &decoder));
                     });
                 }
+
+                if let Some(gas_report) = &mut gas_report {
```
Wodann:
```suggestion
                if let Some(gas_report) = gas_report.as_mut() {
```

This should allow you to avoid the `mut: Option` in the function signature. `.as_mut()` is the way to get mutable access to the internals, whereas `&mut gas_report` would give you a mutable reference to the `Option`, hence why the function signature requires `mut`

js/helpers/src/index.ts
Wodann: > I'm not sure if there is a better way to do this, so I'm very open to suggestions 😄

I pondered the same. See https://github.com/NomicFoundation/edr/pull/1115#discussion_r2436502911

crates/edr_solidity_tests/src/multi_runner.rs
Wodann: See https://github.com/NomicFoundation/edr/pull/1115#discussion_r2436515644 for a suggestion to avoid the `mut gas_report`

crates/edr_solidity_tests/src/multi_runner.rs
popescuoctavian: IIUC `.as_mut()` still requires the variable to be mutable - converts `&mut Option<>` to `Option<&mut>`, so I can't get rid of the `mut` in signature.

In any case, `as_mut()` seems better suited here, since it explicitly gives access to the internals.

crates/edr_solidity_tests/src/multi_runner.rs
```diff
@@ -116,6 +125,8 @@ pub struct MultiContractRunner<
     on_collected_coverage_fn: Option<Box<dyn SyncOnCollectedCoverageCallback>>,
     #[allow(clippy::type_complexity)]
     _phantom: PhantomData<fn() -> (ChainContextT, EvmBuilderT, HaltReasonT, TransactionErrorT)>,
+    /// Whether to generate a gas report after running the tests.
+    generate_gas_report: bool,
```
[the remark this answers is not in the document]
popescuoctavian: Agreed, moved it

crates/edr_solidity_tests/src/multi_runner.rs
popescuoctavian: I went with the first solution, since the second required us to store the results. Thanks for catching this, I initially missed the parallel iterator!

Side note: removed `.finalize()` call since that is not needed as it only computes min/max/mean/median which we don't use. I will remove the function entirely after the refork.

agostbiro: Thanks! We could move initializing the gas report inside the method that fills if it's not too complicated with the return type. Not a merge blocker though

agostbiro (approved):

crates/edr_solidity_tests/src/multi_runner.rs
popescuoctavian: I avoided that since the return type `SuiteResult` is the structure which is passed to the callback and including a gas report in it seems counter-intuitive to me. Doing this also requires making an additional clone of the gas report.

We could have a wrapping structure that contains both the suite result and gas report, but I feel that it complicates the code further. I don't have a strong opinion though, let me know if you think I should do that. Thanks!

popescuoctavian:
Test failures are not related to these changes, merging.

PR #1119, opened 2025-10-15
docs: add docs for runSolidityTests arguments

agostbiro:
Add docs for `runSolidityTests` arguments

### DIFF
--- a/crates/edr_napi/src/context.rs
+++ b/crates/edr_napi/src/context.rs
@@ -166,7 +166,7 @@ impl EdrContext {
     ///   that access artifacts and other functionality (e.g. auto-linking, gas
     ///   reports) can break.
     /// - `testSuites`: the test suite ids that specify which test suites to
-    ///   execute. The test suite artifacts must present in `artifacts`.
+    ///   execute. The test suite artifacts must be present in `artifacts`.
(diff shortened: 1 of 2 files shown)


crates/edr_napi/src/context.rs
```diff
@@ -130,7 +130,7 @@ impl EdrContext {
         Ok(promise)
     }
 
-    #[doc = "Registers a new provider factory for the provided chain type."]
+    /// Registers a new provider factory for the provided chain type.
```
agostbiro: Replaced doc attributes with doc comments as `cargo fmt` will automatically enforce line lengths for the latter.

crates/edr_napi/src/context.rs
```diff
@@ -39,7 +39,7 @@ pub struct EdrContext {
 
 #[napi]
 impl EdrContext {
-    #[doc = "Creates a new [`EdrContext`] instance. Should only be called once!"]
+    /// Creates a new [`EdrContext`] instance. Should only be called once!
```
agostbiro: Replaced with doc comment for consistency with `run_solidity_tests`

popescuoctavian (approved):
LGTM, thanks for adding the info!

PR #1120, opened 2025-10-16
feat: backport Foundry 1.3

agostbiro:
Backport of changes for Solidity tests from Foundry 1.3.0.

https://github.com/foundry-rs/foundry/tree/v1.3.0


agostbiro:
The benchmark regression can be ignored, because that specific scenario is slower with Foundry 1.3 as well. Below is a performance evaluation of 9f0aa18 with the [EDR vs Forge comparison tool.
](https://github.com/NomicFoundation/edr/blob/main/js/benchmark/README.md#compare-edr-and-forge)

The ratio columns contain the ratio of the cumulative execution time of EDR over Forge, so lower is better for EDR. The zero values in the invariant column mean that the repo didn't have invariant tests.

|repo|successful_tests             |failed_tests|total_ratio                                  |unit_ratio|fuzz_ratio|invariant_ratio|
|----|-----------------------------|------------|---------------------------------------------|----------|----------|---------------|
|forge-std|177                          |0           |0.92                                         |0.98      |0.92      |0              |
|morpho-blue|145                          |0           |0.96                                         |0.93      |0.98      |0.95           |
|prb-math|314                          |0           |0.83                                         |0.99      |0.82      |0              |
|solady|1557                         |0           |1.01                                         |0.98      |1.01      |0.89           |
|v4-core|598                          |0           |0.98                                         |0.98      |0.98      |0              |

crates/edr_macros/src/console_fmt.rs
```diff
@@ -19,38 +21,36 @@ pub fn console_fmt(input: &DeriveInput) -> TokenStream {
...
     }
 
+    let fields = s.fields.iter().collect::<Vec<_>>();
     let first_ty = match &fields.first().unwrap().ty {
         Type::Path(path) => path.path.segments.last().unwrap().ident.to_string(),
         _ => String::new(),
     };
 
-    let args: Punctuated<TokenStream, Token![,]> = fields
+    let members = s.fields.members().collect::<Vec<_>>();
```
anaPerezGhiglia: similarly here, we don't need to create a new `Vec` here, we can use `s.fields.members()` directly in the following line 

```rust
let args: Punctuated<TokenStream, Token![,]> = s.fields.members()
        .into_iter()
        .map(
        // ...
...
```

crates/foundry/cheatcodes/Cargo.toml
```diff
@@ -4,48 +4,65 @@ description = "Foundry cheatcodes definitions and implementations"
 
 version.workspace = true
 edition.workspace = true
+#rust-version.workspace = true
+#authors.workspace = true
+#license.workspace = true
+#homepage.workspace = true
+#repository.workspace = true
+#exclude.workspace = true
```
Wodann: Question: Are these meant to be commented?

crates/edr_solidity_tests/src/multi_runner.rs
```diff
@@ -200,8 +203,6 @@ impl<
         .await
         .expect("Thread shouldn't panic")?;
 
-        let test_options: TestOptions = TestOptions { fuzz, invariant };
-
         if generate_gas_report {
             // Traces are needed to generate a gas report
             include_traces = IncludeTraces::All;
```
Wodann: Suggestion: The hack [introduced](https://github.com/NomicFoundation/edr/pull/1147/files#diff-bd0001d25e0dc485810c1b1e3f06130594c921cbe9a2fdfdd8200d4e7c041bd5R212) on line 213 should be resolved by this PR. As such, it would be good to remove the following lines:

```rs
        // HACK: When EVM isolation is enabled, we need to allow internal expectRevert
        // calls because we have a bug in isolation mode.
        if evm_opts.isolate {
            cheats_config_options.allow_internal_expect_revert = true;
        }
```

That way we can validate whether isolation mode runs correctly now.

Wodann: Merged

PR #1122, opened 2025-10-17
fix: broken JSON-RPC tests after Alchemy API change

Wodann:
This fixes two [test failures](https://github.com/NomicFoundation/edr/actions/runs/18598654079/job/53031762074?pr=1111) due to a change in the JSON-RPC API of Alchemy.

Previously they were treating these tests as JSON-RPC errors. Now they're treating them as bad HTTP requests.


popescuoctavian (approved):
LGTM, but it seems to be breaking the formatting in CI, not sure why.

PR #1126, opened 2025-10-21
Add manual prepublish step to publish

nebasuke:
This is meant to replicate the previous pre-publish behaviour. See e.g. https://github.com/NomicFoundation/edr/actions/runs/17685811865/job/50271882890#step:9:16 and note the 
```
> @nomicfoundation/edr@0.12.0-next.7 prepublishOnly /home/runner/work/edr/edr/crates/edr_napi
> bash ../../scripts/prepublish.sh
```

This was disabled for our [latest release](https://github.com/NomicFoundation/edr/actions/runs/18684463630/job/53274615884) due to the `ignore-scripts=true` in https://github.com/NomicFoundation/edr/pull/1114/files#diff-e813096d69c49812e40e00be9ac8fa14d77f0ec1f97ab4c45cb096b3bedd35de.

Add back commented out line to move `optionalDependencies` to `dependencies`.

### DIFF
--- a/scripts/prepublish.sh
+++ b/scripts/prepublish.sh
@@ -7,4 +7,4 @@ pnpm napi prepublish -t npm --skip-gh-release
 
 # temporarily commented to be able to publish in the @ignored org locally with verdaccio
 # rename optionalDependencies key to dependencies
-# jq 'with_entries(if .key == "optionalDependencies" then .key = "dependencies" else . end)' package.json | sponge package.json
+jq 'with_entries(if .key == "optionalDependencies" then .key = "dependencies" else . end)' package.json | sponge package.json


Wodann (approved):

anaPerezGhiglia:
@nebasuke I think that we should remove the `prepublishOnly` script definition from the `crates/edr_napi/package.json` as well (If the script is defined, then it's automatically included in the lifecycle of the publish command)

nebasuke:
> @nebasuke I think that we should remove the `prepublishOnly` script definition from the `crates/edr_napi/package.json` as well (If the script is defined, then it's automatically included in the lifecycle of the publish command)

Good point, addressed here: https://github.com/NomicFoundation/edr/pull/1127

PR #1127, opened 2025-10-21
Remove prePublishOnly step from package.json

nebasuke:
Remove prePublishOnly step from package.json as it is now explicitly covered in the EDR release workflow. 

I've added a changeset to force a version bump.


anaPerezGhiglia (approved):
LGTM 🚀

PR #1136, opened 2025-10-22
chore: improve pre-release CI workflow

anaPerezGhiglia:
## Context
Curretly the release workflow performs some last-mile changes to the napi `package.json` files that are not sync'ed back to the repository. The "lost" changes are:
 1. Update `version` field in platform-specific `package.json` files
 1. Update `napi` root package to list cross-platform packages as `dependencies`

## Proposed Changes
* Modify the current "Open Changeset PR" to include the modifications in point (1), transforming it into a "Open pre-release PR"
  * The goal for this new PR is to include the changes included by changeset action together with updating the platform-specific `package.json` files so that match with napi root `package.json` (which is being updated by the changeset command)

### Technical details
 * I leveraged the [`version` option in changeset action](https://github.com/changesets/action/blob/main/README.md) to execute other commands for generating the version output to be included in the PR
   * I tried different ways for this. The cleanest way I found was to define a new script for this use case and simply call that.
* I left out the `optionalDependencies`/`dependencies` field modifications to `edr_napi/package.json` file
   * My original intention was to include those changes as well in the PR but I stumbled with different complications and decided to leave this bit out
     *  I first tried to include the `dependencies`, but then the CI checks would fail because the `package.json` and `pnpm-lock.yaml` had diverged
     * I decided to run a `pnpm install` as final step then, but that would fail because it couldn't find the versions listed in `dependencies` (which makes a lot of sense, since those versions are not published yet)
     * so I figured "let's have them as `optionalDependencies` in the repo and just convert them to `dependencies` right before publishing". Guess what? it failed as well: Running `pnpm install` after it would pass, but it wouldn't update the `pnpm-lock.yaml` file with the optional dependencies, so when the CI check wanted to run `pnpm install --frozen-lockfile` they would fail because the lockfile and the package.json have diverged 🙄 
     * At this point I decided to timeout this effort since it's not crucial, and revisit it later on
       *  we might want to upgrade NAPI to version 3, which has explicit support por [cross-compilation](https://napi.rs/blog/announce-v3.en#cross-compilation) and [`pnpm` package template](https://napi.rs/blog/announce-v3.en#brand-new-pnpm-package-template) 

## Note for the reviewer
* There are a _lot_ of commits, but only a few changes, so my recommendation is to just review the changes as batch. The list of commit only evidences all the fun I had in the way 🥴  

## Evidence
I edited the workflow trigger configuration for this branch during changes so I could validate its execution
#1135 shows as evidence that the CI workflow works as expected

### DIFF
--- a/.changeset/some-turtles-throw.md
+++ b/.changeset/some-turtles-throw.md
@@ -1,5 +1,0 @@
----
-"@nomicfoundation/edr": patch
----
-
-"Patch EDR CI triggering release workflow"


.changeset/some-turtles-throw.md
```diff
@@ -0,0 +1,5 @@
+---
```
Wodann: todo: Changesets are only used to report user-facing changes to the published NPM packages, not for internal or CI changes. As such, there is no need for a changeset.

scripts/pre_release.sh
```diff
@@ -0,0 +1,35 @@
...
+# We don't want to include because adding it would make the 
+# pnpm-lock.yaml differ with package.json - and `pnpm install` does
+# not fix the lockfile in this case.
+# TODO: NAPI.rs v3 has a skipOptionalPublish option
+git restore ./package.json 
+
+cd ../.. 
+
+# Leave all changes not staged for commit
+# since Changesets action takes care of committing them
```
Wodann: Polish:
```suggestion
# Leave all changes unstaged for commit by the Changesets CI action
```

Wodann (approved):
One todo and a couple of suggestions to the comments, to clarify language. Otherwise LGTM. Feel free to merge after (at least) addressing the todo.

.changeset/some-turtles-throw.md
anaPerezGhiglia: oh I had created this one so that the CI action would trigger properly (changeset does not create the PR if there is no new changeset to include) - and I forgot to remove it afterward. 
Thanks for the catch 🙇

PR #1142, opened 2025-10-30
fix: Node v22 build issue for hardhat-tests

Wodann:
After upgrading to the latest Node version (v22), I was experiencing issues to a newly enabled feature in Node v22 that handles integrated TypeScript support, causing the Mocha test runner to fail.

I added a configuration option to mocha to disable the feature.


anaPerezGhiglia (approved):
LGTM 👍

Wodann:
It seems like CI is still running Node v20, which is incompatible with the flag. I'll need to upgrade the Node version before this can be merged.

PR #1143, opened 2025-10-31
feat(op stack): adopt isthmus

### LINKED ISSUE #906: Implement `Isthmus` hardfork
Since the original implementation of OP stack, the `Isthmus` hardfork has been added, but wasn't fully implemented in EDR yet.

**Definition of Done**
Implement support for Isthmus hardfork according to [spec](https://specs.optimism.io/protocol/isthmus/overview.html)

anaPerezGhiglia:
## Context 
On May 2025 Isthmus hardfork went live in the main OP stack chains.

**Quick overview of the changes introduced by [Isthmus specification ](https://specs.optimism.io/protocol/isthmus/overview.html)**
- implementation of Pectra execution layer in OP stack
- Set `L2ToL1MessagePasser` storage root as `withdrawals_root` in the block header
- New Operator Fee:  includes a new formula for calculating the `total_fee`. It adds a new component so that OP chains can better charge the resource consumptions. 
- Included a new `OperatorFeeVault` predeploy
- Modifications to existing predeploys:
  - L1Block
  - GasPriceOracle

More details on the EDR support analysis can be found [here](https://www.notion.so/nomicfoundation/OP-stack-support-241578cdeaf58040891efa97fe8eca63?source=copy_link#287578cdeaf580aabad2d6d8b7028015)

## Changes already implemented
 - EDR already implements Pectra, so that's already covered
 - Changes in `GasPriceOracle` predeploy has been addressed in #994
 - EDR overrides `withdrawals_root` field when replaying blocks  (`isthmus_header_overrides`)
 
### Isthmus spec 
* Add stubbed data for `OperatorFeeVault` predeploy introduced in this hardfork
 * Add tests to validate that EDR users can configure custom operator fee params via encoding the params in the right `L1Block` predeploy storage 
 * Set the right `withdrawals_root` value on LocalBlockchain genesis for OP spec
   * For doing this I had to change `EthLocalBlock.with_genesis_state` signature so it receives a built `PersistentStateTrie` instead of a diff
 * Extend the existent functionality of op block builder, to set the right `withdrawals_root` value before Isthmus, to match the spec
 * Version `l1_block` predeploy based on hardfork value
   * 👀 Previous version implementation of the proxy contract did not match the bytecode we had, so I assumed that the existent bytecode was the original implementation. So we now have 3 bytecode versions: bedrock? (original bytecode), isthmus (current bytecode) and ecotone (previous implementation of the proxy contract)

### Bonus ✨ 
- Simplified the return type of `StateDebug.account_storage_root`. It was unnecessarily returning a `Result` when all its implementations were returning an `Ok`

### Known limitations
- EDR in fork mode on Isthmus hardfork will not set the right `withdrawals_root` header field - unless overwritten
  - Isthmus hardfork specifies that the header `withdrawals_root` field should contain the storage_root of L2toL1MessagePasser predeploy address at that given block
  - EDR does not support `account_sotrage_root` for `ForkState` . It’s returning the KECCAK representing empty data instead. This is a product and technical decision.
  - Thus, EDR will not set the expected `widthrawals_root` when mining new blocks using Isthmus in fork mode.
- EDR will override `withdrawals_root` with the right value when replaying a block
  - When using the replay block tool, EDR is overriding the `withdrawals_root` value with the replay block value to ensure the block consistency 

## Note for the reviewer
I had started doing auto-contained commits, trying to leave an easy-to-follow trace, but then I had to rebase after #1139 🫣 and I had some hiccups there, so I recommend reviewing this PR as a batch with the description as guide

---

closes #906

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3095,7 +3095,6 @@ dependencies = [
  "edr_evm",
  "edr_evm_spec",
  "edr_gas_report",
- "edr_op",
  "edr_precompile",
  "edr_primitives",
  "edr_receipt",
--- a/crates/edr_provider/Cargo.toml
+++ b/crates/edr_provider/Cargo.toml
@@ -78,7 +78,6 @@ tracing = { version = "0.1.37", features = [
 [dev-dependencies]
 alloy-eips.workspace = true
 anyhow = "1.0.89"
-edr_op.workspace = true
 edr_runtime = { workspace = true, features = ["test-utils"] }
 edr_test_block_replay.workspace = true
 edr_test_utils.workspace = true


crates/edr_napi/src/chains/op.rs
```diff
@@ -208,6 +208,11 @@ pub fn op_genesis_state(hardfork: OpHardfork) -> Vec<AccountOverride> {
                 // uint256 blobBaseFee = 10 gwei
                 value: BigInt::from(0x00000002540be400_u64),
             },
+            StorageSlot {
+                // Operator fee parameters
+                index: BigInt::from(8u64),
+                value: BigInt::from(0u64),
+            },
```
Wodann: Issue: Since this storage value was not present before the Isthmus hardfork, the `AccountOverride` should be different depending on the hardfork too.

crates/edr_op/src/block/builder.rs
```diff
@@ -224,6 +210,22 @@ impl<'builder, BlockchainErrorT: std::error::Error>
     }
 }
 
+/// Prior to isthmus activation: the L2 block header's withdrawalsRoot field
+/// must be:
+///    - nil if Canyon has not been activated.
+///    - `keccak256(rlp(empty_string_code))` if Canyon has been activated.
+///
+/// After Isthmus activation, the withdrawalsRoot field be the
```
Wodann: Typo:

```suggestion
/// After Isthmus activation, the withdrawalsRoot field should be the
```

crates/edr_op/src/block/builder.rs
```diff
@@ -252,3 +254,77 @@ pub fn decode_base_params(extra_data: &Bytes) -> ConstantBaseFeeParams {
...
+mod tests {
+    use std::collections::BTreeMap;
+
+    use edr_block_api::{GenesisBlockFactory, GenesisBlockOptions};
+    use edr_block_header::BlockConfig;
+    use edr_blockchain_api::StateAtBlock as _;
+    use edr_chain_spec_provider::ProviderChainSpec;
+    use edr_provider::spec::LocalBlockchainForChainSpec;
+    // use edr_evm::{blockchain::LocalBlockchain, spec::GenesisBlockFactory,
+    // GenesisBlockOptions};
```
Wodann: Question: Can this comment be removed?

```suggestion
```

crates/state/api/src/lib.rs
```diff
@@ -79,7 +79,7 @@ pub trait StateDebug {
     type Error;
 
     /// Retrieves the storage root of the account at the specified address.
-    fn account_storage_root(&self, address: &Address) -> Result<Option<B256>, Self::Error>;
+    fn account_storage_root(&self, address: &Address) -> Option<B256>;
```
Wodann: Issue: This is a general-purpose API intended for extensibility by other `State` implementations. We cannot control whether they will or will not return an error, so we should keep this API returning a Result for future-compatibility; even if our current internal implementations don't return an error.

crates/edr_op/src/block/builder.rs
anaPerezGhiglia: totally, thanks for the catch!

PR #1149, opened 2025-11-03
fix(ci): release workflow 

anaPerezGhiglia:
## Changes
- Change the release workflow file so that the `check_commit` job is only required for `publish` job. As a consequence of this, the `prepare` and `release` jobs will run for all branches and commits, not only for release commits
- Change the `check_commit` job so that it directly determines the publish tag
  - Moved the if-else statement from publish job here
- Simplify the `publish` job so that it publishes using the tag defined by `check_commit` job
- Fix inexistent output variable

### Evidence
 - [If tag output is empty then publish job doesn't run](https://github.com/NomicFoundation/edr/actions/runs/19069502599)
 - [If tag in known, then it would execute the publish command](https://github.com/NomicFoundation/edr/actions/runs/19069360609)
 - [unknown tag fails](https://github.com/NomicFoundation/edr/actions/runs/19069192311)

### DIFF
--- a/.github/workflows/edr-npm-release.yml
+++ b/.github/workflows/edr-npm-release.yml
@@ -534,7 +534,7 @@ jobs:
         with:
           node-version: 24
       - name: Update npm to make sure it supports Trusted Publishing
-        run:  npm install -g npm@v11.6.2
+        run: npm install -g npm@v11.6.2
       - name: Download packed tarball
         uses: actions/download-artifact@v4
         with:


.github/workflows/edr-npm-release.yml
```diff
@@ -530,38 +536,22 @@ jobs:
         run: tar -xvzf ${{ needs.prepare.outputs.filename }}
       - name: Publish
         run: |
-          if git log -1 --pretty=%B | grep "^edr-[0-9]\+\.[0-9]\+\.[0-9]\+-";
+          if [ "${{ needs.check_commit.outputs.tag}}" = "next" ]
```
nebasuke: Shouldn't this be ==?

anaPerezGhiglia: No since it's using single `[ ]`
string comparison in bash when using `[[]]` requires `==` but when using `[]` it's just `=`

PR #1158, opened 2025-11-05
fix(ci): fix publish job 

anaPerezGhiglia:
The last publish job failed with a `pnpm: command not found`. We were missing to configure pnpm as well as node in the publish job

I also changed the review and publish jobs node config to use node 22 as the rest of the jobs


nebasuke (approved):
Thanks for also unifying the node versions!

nebasuke:
customSmallerIsBetter seems a bit too tuned at the moment and is failing on most PRs. Can ignore.

PR #1177, opened 2025-11-10
Don't persist credentials and use pinned version for action/checkout.

nebasuke:
Don't persist credentials and use pinned version for action/checkout.

Tackles https://github.com/alcuadrado/trusted-publishing-example/issues/5 for EDR (https://github.com/NomicFoundation/edr/issues/1170). 

(I'll pin further actions in a follow-up PR.)

### DIFF
--- a/.github/workflows/pre-release.yml
+++ b/.github/workflows/pre-release.yml
@@ -14,8 +14,7 @@ jobs:
       pull-requests: write
     steps:
       - name: Checkout Repo
-
-      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5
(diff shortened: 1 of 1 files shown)


.github/workflows/pre-release.yml
```diff
@@ -14,10 +14,12 @@ jobs:
       pull-requests: write
     steps:
       - name: Checkout Repo
-        uses: actions/checkout@v2
+
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5
```
anaPerezGhiglia: this change makes the file invalid. The `uses` key should be part of the same object that starts with `- name: Checkout Repo` , or remove the `name` key all together and just leave the `uses`

anaPerezGhiglia (changes requested):
Small but lethal comment. The rest looks good

PR #1178, opened 2025-11-10
Pin github actions to commit hashes

nebasuke:
Pin github actions to commit hashes. 

Tackles https://github.com/alcuadrado/trusted-publishing-example/issues/6 for EDR (https://github.com/NomicFoundation/edr/issues/1170).

(Git checkout actions are already pinned in https://github.com/NomicFoundation/edr/pull/1177.)


anaPerezGhiglia (commented):
To what extent do we want to enforce this? 
There are several actions still missing to be pinned, like all the `actions/checkout` and some others like `addnab/docker-run-action`, `dtolnay/rust-toolchain`, `taiki-e/install-action`, `changesets/action` (in `pre-release.yml` action that wasn't included here yet) and others.

PR #1185, opened 2025-11-13
test: set fuzz seed by default

agostbiro:
Set fuzz seed by default for `edr_solidity_test` integration tests to avoid spurious failures. Local tests passed 25 times in a row with this.


popescuoctavian (approved):
LGTM, thanks for taking care of this!

PR #1147, opened 2025-11-01
fix: prevent error when expectRevert is called in isolate mode

Wodann:
Resolves #1146.

This was an existing bug in the Foundry codebase. I was able to resolve it by backporting https://github.com/foundry-rs/foundry/pull/9018. During the backport, I used Foundry's master branch (https://github.com/foundry-rs/foundry/tree/ad9fcefde4dff489ee135ff5d280801b4f569d25).

PR #1188, opened 2025-11-13
fix: remove isolation mode hack

Wodann:
Removes the hack [introduced](https://github.com/NomicFoundation/edr/pull/1147/files#diff-bd0001d25e0dc485810c1b1e3f06130594c921cbe9a2fdfdd8200d4e7c041bd5R212) introduced to make isolation mode work. This allows us to validate that the Foundry refork addressed the core issue.


agostbiro (approved):
Thanks, good catch!

PR #1193, opened 2025-11-14
ci: migrate soltest benchmarks to microseconds

agostbiro:
Switch to microseconds resolution from milliseconds for soltest benchmarks, as ms resolution is too low and it causes spurious failures.

TODOs before merging:
- https://github.com/nomic-foundation-automation/edr-benchmark-results/pull/3
- Revert 16ff8a6ae0907bf44285bba74266c34addd412b3


anaPerezGhiglia (approved):
LGTM 👍 
I'm approving it so you can merge it as soon as the TODOs are done

PR #1194, opened 2025-11-16
feat: add Osaka blob parameters

Wodann:
This PR ensures that the Osaka blob parameters are used ([EIP-7918](https://eips.ethereum.org/EIPS/eip-7918)).

I also discovered an unused file that should have been removed. I deleted it as part of this PR.


anaPerezGhiglia (approved):
LGTM 🚀

PR #1196, opened 2025-11-17
refactor: split `edr_runtime` into smaller crates

Wodann:
This PR is a continuation of #1139, splitting the `edr_runtime` crate into more, smaller crates.

Most crates that depended on `edr_runtime` only depended on the logic in what's now called `edr_tracing`. Incremental compilation should be sped up for those crates.

I also ended up removing several tracked but unused files from the git repo.

### DIFF
--- a/book/src/04_multichain/01_chain_spec.md
+++ b/book/src/04_multichain/01_chain_spec.md
@@ -2,8 +2,7 @@
 
 EDR uses a concept called _chain specification_ to define all necessary types and functionality to support an Ethereum chain at compile-time.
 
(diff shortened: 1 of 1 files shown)


popescuoctavian (approved):
LGTM, thanks! Looks like there are some minor styling issues. We should probably also [update the docs](https://github.com/NomicFoundation/edr/blob/bb832ad91835f1ba8a59d4f9db135a439bdcaa38/book/src/04_multichain/01_chain_spec.md#L7-L14) to reflect the new structure.

PR #1197, opened 2025-11-17
feat: return error when block exceeds max RLP size

Wodann:
This PR implements the validity check of [EIP-7934](https://eips.ethereum.org/EIPS/eip-7934).


popescuoctavian (approved):
LGTM, thanks!

Issue #1212, opened 2025-12-03
Fully support OP Jovian hardfork


anaPerezGhiglia:
Hardfork spec [here](https://specs.optimism.io/protocol/jovian/overview.html)

PR #1213, opened 2025-12-03
feat(op): min jovian hardfork support

### LINKED ISSUE #1211: Minimum support for OP Jovian hardfork
- include new OP hardfork enum value
- implement hardfork breaking changes so that EDR can replay OP and base mainnet blocks

Hardfork spec [here](https://specs.optimism.io/protocol/jovian/overview.html)

anaPerezGhiglia:
## Context

- Since December 2 2025, Jovian hardfork is activated in OP Mainnet and Base Mainnet chains alongiside other OP stack chains.
- This hardfork introduces some breaking changes. See the [notice](https://docs.optimism.io/notices/upgrade-17) for further details.
- Since this hardfork went live, EDR replay-block tests have been failing.

## Changes
This PR aims to adapt EDR codebase the minimum to support new jovian changes, so that it can keep replaying blocks. 
The full adoption of Jovian hardfork is out of scope of this PR. We have #1212 to track that.

The main changes are:
 - support new version of the encoded `base_fee_params` in the `header.extra_data` field
 - clamp the `base_fee` calculation if the new `min_base_fee` is in included in the parent header
 - when replaying blocks, use a new `header_overrides` that uses the remote `blob_gas` field.
   - bypass `blob_gas` transaction validation when there are no `blob_hashes`. This way the validation of transactions on jovian activated blocks do not fail.
   - This new `header_override` may be removable down the road once EDR fully supports Jovian
  
closes #1211

### DIFF
--- a/crates/edr_op/src/eip1559.rs
+++ b/crates/edr_op/src/eip1559.rs
@@ -6,14 +6,44 @@ pub const HOLOCENE_BASE_FEE_PARAM_VERSION: u8 = 0x0;
 /// Post-Jovian version of the dynamic base fee parameter.
 pub const JOVIAN_BASE_FEE_PARAM_VERSION: u8 = 0x1;
 
-/// Encodes the dynamic base fee parameters into a byte array.
+/// Encodes the dynamic base fee parameters into a byte array following Holocene
+/// spec
 ///
 /// <https://specs.optimism.io/protocol/holocene/exec-engine.html#dynamic-eip-1559-parameters>
+pub fn encode_dynamic_base_fee_params_holocene(base_fee_params: &ConstantBaseFeeParams) -> Bytes {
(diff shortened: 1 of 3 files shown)


crates/edr_op/src/eip1559.rs
```diff
@@ -1,13 +1,19 @@
 use edr_eip1559::ConstantBaseFeeParams;
 use edr_primitives::Bytes;
 
+/// Holocene version of the dynamic base fee parameter.
+pub const HOLOCENE_BASE_FEE_PARAM_VERSION: u8 = 0x0;
 /// Currently used version of the dynamic base fee parameter.
```
Wodann: Suggestion(non-blocking)
```suggestion
/// Post-Jovian version of the dynamic base fee parameter.
```

PR #1222, opened 2025-12-11
build: upgrade to Slang v1.3.1

Wodann:
Upgraded Slang dependency to v1.3.1, with support for Solidity 0.8.31. This is used for instrumenting source code with code coverage hits.


nebasuke (approved):
Thanks for quickly doing this!

PR #1248, opened 2025-12-29
feat: add API to determine latest supported Solidity version

Wodann:
This adds an API to determine the latest supported Solidity version.

This will be helpful for Hardhat to fall back to the latest supported version of Solidity if a newer `solc` compiler is available than the currently supported version of Slang in EDR.


anaPerezGhiglia (approved):
LGTM 🚀

PR #1266, opened 2026-01-21
chore: fix inconsistent function-level config override test failures

popescuoctavian:
Test was inconsistent due to a contract misassumption - the invariant testing was finding counterexamples which caused the test to fail and no more runs to be executed. Fixed the test contract to make sure it can never fail.

### DIFF
--- a/js/integration-tests/solidity-tests/test-contracts/InvariantTest1.t.sol
+++ b/js/integration-tests/solidity-tests/test-contracts/InvariantTest1.t.sol
@@ -4,18 +4,10 @@ pragma solidity ^0.8.18;
 import {Test} from "forge-std/src/Test.sol";
 
 contract InvariantBreaker {
-    bool public flag0 = true;
     bool public flag1 = true;
 
-    function set0(int256 val) public returns (bool) {
-        if (val % 100 == 0) {
-            flag0 = false;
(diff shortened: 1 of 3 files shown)


anaPerezGhiglia (commented):
let me start by saying that I'm a complete newbie with Solidity & Solidity tests, so I might just be missing a lot of context & knowledge.

Are the `set0` or `set1` functions ever called when running the tests? Looking at the code I would assume that they don't, and so the `inv.flag1()` call would always return true since it's the initial value

anaPerezGhiglia:
thanks for the explanation! I was missing the "target contract" behavior. 
Knowing that, I have other silly questions
 - Having the condition that's impossible to satisfy, do we need this `if` statement at all?
 - Now that the contract is "stable" - can never break the invariant, the name `InvariantBreaker` becomes misleading 
 - We have another contract like this one in `crates/edr_solidity_tests/tests/testdata/default/fuzz/invariant/common/InvariantTest1.t.sol`, do we want them to have different behavior?
 
 ☝️ With all of that in mind, I wonder if we shouldn't just create a new "unbreakable" contract just for this test case

anaPerezGhiglia (approved):
LGTM ✨

PR #1273, opened 2026-01-27
chore: add op support tests

anaPerezGhiglia:
Add provider tests that validate what's written in [EDR-OP stack support document](https://www.notion.so/nomicfoundation/EDR-OP-stack-support-2f0578cdeaf5801085e0d93507542467#2f0578cdeaf580d997a8ccae1949a3fa)

Writing those I found a bug on `create_forked_blockchain_and_state`, so I fixed it here as well. 

Also, tweaked `edr_napi/test/op.ts` so TS does not complain about not matching types


Wodann (approved):
Some questions and suggestions, but overall LGTM! Thanks for the fixes!

PR #1278, opened 2026-01-30
refactor: unify provider & soldity test tracing architecture

Wodann:
This PR unifies the tracing architectures of the provider & solidity tests by introducing the `TracingInspector`—already used by the solidity test—in the JSON-RPC provider.

To get this to work, I needed to
- separate several files into new, smaller crates for reuse between both `edr_solidity_tests` and `edr_provider`
- implement `CallTraceArena` decoding using the `ContractDecoder`
- adapt the execution and inspector business logic to be able to handle `TracingInspector`'s per-transaction memory representation

PR #1287, opened 2026-02-06
feat: add the ability to specify the hardfork when running Solidity tests

popescuoctavian:
This PR adds the ability to specify the hardfork when running Solidity tests using the Typescript interface. It also adds support for the Osaka hardfork in Solidity tests.

### DIFF
--- a/crates/edr_napi/src/solidity_tests/config.rs
+++ b/crates/edr_napi/src/solidity_tests/config.rs
@@ -75,6 +75,9 @@ pub struct SolidityTestRunnerConfigArgs {
     /// Defaults to `31337`.
     #[serde(serialize_with = "serialize_optional_bigint_as_struct")]
     pub chain_id: Option<BigInt>,
+    /// The hardfork to use for EVM execution.
+    /// Defaults to the chain's default hardfork if not provided.
+    pub hardfork: Option<String>,
     /// The gas limit for each test case.
     /// Defaults to `9_223_372_036_854_775_807` (`i64::MAX`).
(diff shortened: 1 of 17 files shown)


crates/edr_napi/src/solidity_tests/config.rs
```diff
@@ -75,6 +75,9 @@ pub struct SolidityTestRunnerConfigArgs {
     /// Defaults to `31337`.
     #[serde(serialize_with = "serialize_optional_bigint_as_struct")]
     pub chain_id: Option<BigInt>,
+    /// The hardfork to use for EVM execution.
+    /// Defaults to the chain's default hardfork if not provided.
+    pub hardfork: Option<String>,
```
Wodann: Issue: The default value is controlled and updated by REVM. This means that we'd create a connection between upgrades of REVM and changing the default hardfork in Hardhat.

In EDR we broke this connection by always requiring the hardfork to be specified and two chain-specific functions: `l1HardforkLatest` and `opLatestHardfork`. The values returned from those functions are controlled by EDR. That way the EDR team can decide when we start supporting a new hardfork.

The full process and order of how we add support/upgrade a new hardfork for L1 can be found [here](https://www.notion.so/nomicfoundation/Ethereum-L1-Hardfork-Checklist-2c6578cdeaf580a497e0c3c7583f6ce5).

Wodann (changes requested):
Overall looks good. I do have one concern about the API surface

This PR will also introduce significant conflicts with https://github.com/NomicFoundation/edr/pull/1278 due to the changes in precompile decoding logic

Wodann (approved):

PR #1293, opened 2026-02-12
ci: fix release workflow

Wodann:
Fixes [CI release workflow failures](https://github.com/NomicFoundation/edr/actions/runs/21957607460/job/63425983634) due to https://github.com/addnab/docker-run-action/issues/62 by using a fork of the original GitHub action.


anaPerezGhiglia (approved):
Thanks for addressing this so quickly!

Would the idea be to maintain this fork even after the issue gets resolved in the upstream repository? If not, it might be worth adding a TODO comment to switch back to the original action once the issue is addressed.

PR #1298, opened 2026-02-17
fix: return error for mismatched names in gas snapshot cheatcodes

### LINKED ISSUE #1291: `startSnapshotGas`/`stopSnapshotGas` should throw clear errors on misuse instead of silently failing or panicking
While testing the snapshot gas cheatcodes ([context](https://github.com/NomicFoundation/hardhat/pull/7901#issuecomment-3871065162)), we found that some misuse scenarios either fail silently or crash with a panic message. The behavior is also inconsistent across scenarios. We'd expect all cases to throw with a clear error, since they're almost certainly unintentional.

Here's what we found:

### 1. Missing `vm.stopSnapshotGas()`

- **Hardhat (`hardhat test --snapshot`):** No error is thrown, the snapshot is silently removed.
- **Foundry (`forge snapshot`):** No error is thrown, the snapshot is just not recorded.

### 2. Mismatched names (`vm.startSnapshotGas(name1)` / `vm.stopSnapshotGas(name2)`)

For example:

```solidity
function testMultipleOperations() public {
    vm.startSnapshotGas("multiple-ops");
    for (uint8 i = 0; i < 18; i++) {
        counter.inc();
    }
    vm.stopSnapshotGas("wrong-closing");
}
```

- **Hardhat:** No error is thrown. `"multiple-ops"` snapshot is recorded, `"wrong-closing"` is not.
- **Foundry:** No error is thrown, the snapshot is not recorded.

### 3. Missing `vm.startSnapshotGas()`

Both Hardhat and Foundry crash with a panic:

```
called `Option::unwrap()` on a `None` value
```

popescuoctavian:
This PR modifies the way `group` and `name` is handled for  the `stopGasSnapshot` cheatcode. Previously, when either one of the parameters was missing, both values would default to the currently active gas snapshot.

However, for `stopSnapshotGas` this was not correct for cases where only one of the parameters is missing. For example, when using

```rust
function stopSnapshotGas(string calldata name)
```

the group is derived from the contract name, but the snapshot name should be the one provided in the cheatcode. In the previous implementation, the name would be overridden with the active snapshot name, resulting in the generation of a snapshot even for mismatched names.

This PR fixes this by using the active snapshot group & name values only for missing components, allowing us to identify mismatched snapshot names and produce an error as expected.

closes #1291

### DIFF
--- a/.changeset/mighty-geckos-change.md
+++ b/.changeset/mighty-geckos-change.md
@@ -1,0 +1,5 @@
+---
+"@nomicfoundation/edr": patch
+---
+
+Fixed error handling for mismatched names in gas snapshot cheatcodes
--- a/crates/foundry/cheatcodes/src/evm.rs
+++ b/crates/foundry/cheatcodes/src/evm.rs
@@ -3468,17 +3468,38 @@ fn inner_stop_gas_snapshot<
     group: Option<String>,
     name: Option<String>,
 ) -> Result {
-    // If group and name are not provided, use the last snapshot group and name.
-    let (group, name) = group.zip(name).unwrap_or_else(|| {
-        let (group, name) = ccx
+    // Resolve group and name, falling back to the active snapshot for any
+    // missing component.
+    let (group, name) = match (group, name) {
+        (Some(g), Some(n)) => (g, n),
+        (None, None) => ccx
             .state
             .gas_metering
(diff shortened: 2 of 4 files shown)


Wodann (approved):
LGTM. Thanks for the fix!

PR #1307, opened 2026-02-24
fix: return error instead of panic for missing `startSnapshotGas`

### LINKED ISSUE #1291: `startSnapshotGas`/`stopSnapshotGas` should throw clear errors on misuse instead of silently failing or panicking

popescuoctavian:
This PR changes the behavior when calling `stopSnapshotGas` without an active snapshot in Solidity tests. Previously, EDR would panic with an ambiguous message. Now the Solidity test fails with an appropriate message.

### DIFF
--- a/.changeset/swift-bees-enter.md
+++ b/.changeset/swift-bees-enter.md
@@ -1,0 +1,5 @@
+---
+"@nomicfoundation/edr": patch
+---
+
+Fixed error reporting when calling `stopSnapshotGas` without an active snapshot
--- a/crates/foundry/cheatcodes/src/evm.rs
+++ b/crates/foundry/cheatcodes/src/evm.rs
@@ -3470,35 +3470,20 @@ fn inner_stop_gas_snapshot<
 ) -> Result {
     // Resolve group and name, falling back to the active snapshot for any
     // missing component.
+    let (active_group, active_name) = match ccx.state.gas_metering.active_gas_snapshot.as_ref() {
+        Some((group, name)) => (Some(group), Some(name)),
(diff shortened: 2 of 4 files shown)


anaPerezGhiglia (approved):
LGTM 🚀 
thanks for taking the feedback into consideration!

PR #1311, opened 2026-02-25
chore(CI): fix post-setup-node step "Path Validation Error"

anaPerezGhiglia:
## Context
The `publish` job has been failing during the `post-setup-node` step, as seen in [this workflow run](https://github.com/NomicFoundation/edr/actions/runs/22329057297/job/64609694896).

## Fix
The root cause is that `setup-node` was configured with `cache: pnpm`, but the job does not install dependencies (it only needs Node.js to run `npm publish`). Because `cache` was set, the `post-setup-node` step attempted to save the pnpm store cache, which failed since the store path was never created (no `pnpm install` runs in this job).

Removing `cache` should prevent the post step to skip cache saving altogether, fixing the error.

## Bonus
Additionally, I set `package-manager-cache: false` explicitly. The `setup-node` action [warns about a breaking change in v5](https://github.com/actions/setup-node/blob/main/README.md):

> For workflows with elevated privileges or access to sensitive information, we recommend disabling automatic caching by setting `package-manager-cache: false` when caching is not needed for secure operation.

Currently, the auto-caching feature [does not detect pnpm](https://github.com/actions/setup-node/blob/774c1d62961e73038a114d59c8847023c003194d/__tests__/main.test.ts#L364), so it wouldn't activate for this project's setup. Still, setting it to `false` explicitly opts out in case support is expanded in the future, and makes the intent clear.

### DIFF
--- a/.github/workflows/edr-npm-release.yml
+++ b/.github/workflows/edr-npm-release.yml
@@ -462,7 +462,7 @@ jobs:
         run: |
           FILENAME="${{ env.APP_NAME }}_bundle.tar.gz"
           NAPI_EXTRA_FILES=("npm/" "LICENSE" "README.md" "package.json" "Cargo.toml")
-          readarray -t FILES < <(cat package.json | jq -r .files[] ) 
+          readarray -t FILES < <(cat package.json | jq -r .files[] )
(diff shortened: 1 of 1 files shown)


popescuoctavian (approved):
LGTM, thanks for fixing this!

PR #1318, opened 2026-03-03
build: upgrade to Rust v1.91.0

Wodann:
Upgraded to Rust v1.91.0.

This is a prerequisite for https://github.com/NomicFoundation/edr/pull/1301 to be able to upgrade to a newer version of `revm-inspectors` due to a transitive dependency on `alloy` crates which increased the minimal Rust version to v1.91.


anaPerezGhiglia (approved):
This PR made me realize that we are using the rust-toolchain file only for devcontainers. All the CI actions either default on `stable` or directly set `nightly`. I think we should fix the version for CI actions. WDYT?

I'm approving this PR since my comment goes beyond the scope of this PR

anaPerezGhiglia:
Oh, I wasn't aware of that, nice!

I double checked on the [action we are using](https://github.com/actions-rust-lang/setup-rust-toolchain), and it supports this

> All inputs are optional. If a [toolchain file](https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file) (i.e., rust-toolchain or rust-toolchain.toml) is found in the root of the repository and no toolchain value is provided, all items specified in the toolchain file will be installed. If a toolchain value is provided, the toolchain file will be ignored. If no toolchain value or toolchain file is present, it will default to stable

However, on [release workflow](https://github.com/NomicFoundation/edr/blob/1e40d8eaea3225793dcdc91b95affa4a0ad63635/.github/workflows/edr-npm-release.yml#L97) we are using [another action](https://github.com/dtolnay/rust-toolchain), and we are specifying `toolchain: stable`. In that same workflow, we are also installing rust "manually" and [also setting `--default-toolchain stable`](https://github.com/NomicFoundation/edr/blob/1e40d8eaea3225793dcdc91b95affa4a0ad63635/.github/workflows/edr-npm-release.yml#L148). We should change this workflow so we make sure we run everything with the same version.

PR #1321, opened 2026-03-05
fix(cheatcodes): ensure `CallerMode` compatibility with `forge-std`

anaPerezGhiglia:
## Summary

- Brought back `Broadcast` and `RecurrentBroadcast` variants to the `CallerMode` enum, which were removed in #1206.
  - Their absence broke ABI compatibility with forge-std
  - The side effect: `Prank (3)` and `RecurrentPrank (4)` were decoded as `Broadcast (1)` and `RecurrentBroadcast (2)` by test contracts using `forge-std`, breaking  standard Foundry test patterns that rely on pausePrank-style modifiers
- Added tests verifying the enum ordinals (`None=0, Prank=3, RecurrentPrank=4`)
- Updated the cheatcodes book to remind developers to keep hardhat-website docs in sync with `crates/foundry/cheatcodes/spec/src/vm.rs`

## Evidence
I validate that previously-failing aave-v4 tests now pass, by linking `aave-v4 -> local hardhat -> local edr_napi (with changes)`

https://github.com/user-attachments/assets/dccda0fa-4cf0-4351-8774-fbfd074c7a36

closes #1319

### DIFF
--- a/.changeset/olive-signs-cough.md
+++ b/.changeset/olive-signs-cough.md
@@ -1,0 +1,7 @@
+---
+"@nomicfoundation/edr": minor
+"solidity-tests": minor
+---
+
+Added `Broadcast` and `RecurrentBroadcast` variants to the `CallerMode` enum. These variants are unsupported by EDR cheatcodes implementation but are needed to  
+
--- a/.changeset/olive-signs-cough.md
+++ b/.changeset/olive-signs-cough.md
@@ -3,5 +3,4 @@
 "solidity-tests": minor
 ---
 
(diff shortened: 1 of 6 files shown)


Wodann (approved):
Nicely done and thank you for the detailed PR description ❤️

anaPerezGhiglia:
The Solidity test runner benchmark job fails, even after some re-runs. 
I don't think that any of the modifications in this PR justify a performance regression.
Looking at the [soltests graph](https://nomic-foundation-automation.github.io/edr-benchmark-results/soltests/) it looks like the last commit in `main` is an outlier. 

image

Comparing previous performance time in `main`, this branch does not imply a regression.
All that said, I'm considering this PR safe to be merged

PR #1327, opened 2026-03-11
fix: resolve proxied calls against the implementation's ABI

### LINKED ISSUE #7434: hardhat is not able to figure out function selector even though they are available as part of the source artifacts
### Version of Hardhat

3.0.6

### What happened?

I am using a proxy contract and when there are reverts the hardhat console show these :

```
eth_estimateGas                                                                                                                                                                                                                               
  Contract call:       ERC173Proxy#<unrecognized-selector>                                                                                                                                                                                    
  From:                0x258e5b287b1b75adc0f35abce119a3d61f839468                                                                                                                                                                             
  To:                  0xa513e6e4b8f2a923d98304ec87f64353c4d5c853                                                                                                                                                                             
  Value:               0 ETH                                                                                                                                                                                                                  
                                                                                                                                                                                                                                              
  Error: VM Exception while processing transaction: reverted with an unrecognized custom error (return data: 0x954de3349965507d1a55bcc2695c58ba16fb37d819b0a4dcaaf6496e74a5a3966fafc712) 
```

but these function selector should be known to hardhat as the implementation contract is one of source artifact hardhat knows about

### Minimal reproduction steps

- deploy a proxy with implementation
- call it in a way that it reverts

### Search terms

_No response_

wighawag:
fix https://github.com/NomicFoundation/hardhat/issues/7434

better than https://github.com/NomicFoundation/edr/pull/1324 as it supports any proxy, diamond ...

it is also simpler

Note that I modified the proxy chain detection logic to only match for the first 4 bytes so it supports things like clone-with-immutable-args that append data to the calldata

### DIFF
--- a/.changeset/chubby-masks-fall.md
+++ b/.changeset/chubby-masks-fall.md
@@ -1,0 +1,5 @@
+---
+"@nomicfoundation/edr": patch
+---
+
+universal proxy support for function identifier lookup#1327
--- a/.changeset/chubby-masks-fall.md
+++ b/.changeset/chubby-masks-fall.md
@@ -2,4 +2,4 @@
 "@nomicfoundation/edr": patch
 ---
 
-universal proxy support for function identifier lookup#1327
+Fixed calls through proxy contracts being reported as `<unrecognized-selector>` in logged transactions and calls. When a function selector is not found in the called contract's ABI, it is now resolved against the ABI of the implementation behind the proxy, and the call is labeled with the full delegation chain, e.g. `Proxy>Implementation#setValue`. This works for any proxy that forwards the selector via `DELEGATECALL`.
--- a/crates/edr_napi/test/proxyCalls.ts
+++ b/crates/edr_napi/test/proxyCalls.ts
@@ -1,0 +1,226 @@
+import { assert } from "chai";
+import * as fs from "fs";
+import {
+  ContractDecoder,
+  GENERIC_CHAIN_TYPE,
+  Provider,
+  SubscriptionEvent,
+  TracingConfigWithBuffers,
+} from "..";
+import {
+  DEFAULT_GENESIS_ADDRESS,
+  deployContract,
+  fundedGenesisState,
+  getContext,
+  l1ProviderConfig,
+  registerGenericProviderFactory,
+  sendTransaction,
+} from "./helpers";
+
+// Contract code in edr/data/contracts/ProxyMultipleImplementations.sol.
+// The proxies delegate through their fallback functions, so their ABIs do not
+// contain the implementations' functions. This makes calls through them hit
+// the proxy chain resolution in `ContractDecoder::populate_call_trace_arena`.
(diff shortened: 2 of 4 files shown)


wighawag:
Hey @Wodann any update on this ? was it integrated in other PR ?

Wodann:
> Hey @Wodann any update on this ? was it integrated in other PR ?

Sorry for the delay!

I incorporated the fix as part of #1333. This has since been released in Hardhat.

I do still need to rebase this PR to make sure that your tests are added and validated

wighawag:
Hey any update on this?
 #1333 does not seem to fix https://github.com/NomicFoundation/hardhat/issues/7434
 
 happy to help if needed

wighawag:
I created a reproduction repo showing that #1333 do not fix https://github.com/NomicFoundation/hardhat/issues/7434 but #1327 (not rebased yet) fix it

see: https://github.com/bug-reproduction/hardhat-7434/#readme


it also show that error are not decoded in console even with #1327, so this is something to improve

popescuoctavian:
Hey @wighawag, thank you for the contribution!

I've made some changes to the PR to make it ready to merge:

1. Rebased the branch onto main, since `proxy_detection.rs` has changed in the meantime and already includes a fix. The tests you added were ported over.

2. Refactor for `contract_decoder.rs`: instead of pre-computing proxy chains for every node in the call trace arena, detect them only when a function selector is not found in the called contract's ABI — the common case, where every selector resolves, does no proxy detection work at all.
- Split `populate_call_trace_arena` into an immutable decoding pass and a second pass that assigns the results, so that proxy chain detection can inspect the arena at the exact point of a selector miss.
- Use the `CallTrace` chain returned by `detect_proxy_chain` directly, instead of reversing it into a list of addresses.
- When decoding a proxied call's arguments, fall back to the calldata the implementation actually received, to support proxies that modify the forwarded calldata (e.g. clones-with-immutable-args).
- Deduplicate input argument decoding into `decode_input_args`.
- Add provider logging tests covering direct calls, proxied calls (transaction and eth_call), chained proxies, and the unrecognized-selector fallback.

This does not address the "unrecognized custom error" that you pointed [here](https://github.com/bug-reproduction/hardhat-7434/#defect-2-unfixed-by-either-pr). It will be addressed in a follow-up PR.

Wodann (approved):
Apologies for the delay @wighawag!

@popescuoctavian ended up pushing this over the finish line instead of me. Thanks to both of you for putting this fix together! 🙏

Wodann:
We've released [EDR v0.20.0](https://github.com/NomicFoundation/edr/releases/tag/%40nomicfoundation%2Fedr%400.20.0) with a fix for this issue. 

To upgrade to the latest EDR version immediately, you can remove your `node_modules` directory and npm/yarn/pnpm lock file and reinstall your dependencies. Otherwise you can wait for the next Hardhat release which will automatically use the latest EDR version.

You can verify that you're using the latest version of EDR by running the following command: `npm ls @nomicfoundation/edr`.

PR #1330, opened 2026-03-13
refactor: split execution and setup traces

Wodann:
While fixing bugs outlined in #1325 and adding support for multiple deployments of the same contract, this small refactor was necessary. I decided to split it into a separate PR to keep reviewing easier.

The goal is to split setup traces (either deployment or calls to the `setUp()` function) from execution traces. They used to be tracked together, but setup traces are generated once per test suite, whereas execution traces are generated per test.

This refactor creates a clear distinction between the two. User-facing APIs still return a combination of both types of traces for each test, for backwards compatibility.


popescuoctavian (approved):
LGTM, thanks for implementing this!

PR #1332, opened 2026-03-17
fix: allow multiple deployments per contract in gas report

Wodann:
This PR addresses a request to support multiple contract deployments per gas report.

**Note**

- While making changes, I removed the `Display`, `Deserialize`, and `Serialize` trait implementations for `GasReport` as they were not used and would require modifications to support multiple contract deployments.
- I removed the `comfy-table` dependency as it was no longer necessary after removing the `Display` trait.
- I added tests to validate that the status of `setUp` is correctly applied to the contract's deployment
- I cleaned up some iterator code to reduce allocations and conform idiomatic Rust.

### DIFF
--- a/.changeset/full-terms-call.md
+++ b/.changeset/full-terms-call.md
@@ -1,0 +1,5 @@
+---
+"@nomicfoundation/edr": patch
+---
+
+Added support for tracking multiple contract deployments in a gas report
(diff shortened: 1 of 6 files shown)


popescuoctavian (approved):
LGTM, thanks for taking care of this! Also, thanks for cleaning up the code. I'll merge this with the `edr_gas_report` crate as soon as I have some bandwidth.

PR #1336, opened 2026-03-19
feat: add runtime bytecode size to DeploymentGasReport

popescuoctavian:
This PR adds runtime bytecode size information to contract deployment gas reports.

Other changes:
- removed `comfy-table` dependency since `Display` trait was not used for `edr_gas_report::GasReport`

closes #1329

### DIFF
--- a/.changeset/spicy-cities-ask.md
+++ b/.changeset/spicy-cities-ask.md
@@ -1,0 +1,5 @@
+---
+"@nomicfoundation/edr": minor
+---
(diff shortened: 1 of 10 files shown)


Wodann (approved):
Thanks for implementing the new feature!

I only have one nitpick, but Copilot's has at least one suggestion that I think must be included before merging this.

Once that's done, feel free to merge without reapproval from me.

PR #1350, opened 2026-03-30
fix: add missing `show_logs` field to global fuzz config

popescuoctavian:
This PR adds `show_logs` to the global fuzz config which is exposed through NAPI. The field is already present on the Rust side of the configuration , but it couldn't be set through JS global config.


anaPerezGhiglia (approved):
LGTM!

Wodann (approved):
LGTM!

PR #1359, opened 2026-04-14
fix: replicate beacon root oracle state when replaying blocks

anaPerezGhiglia:
## Context
https://github.com/NomicFoundation/edr/issues/1346#issuecomment-4246035282

## Changes
Update `run_full_block` function so that the `replay-block` tool and `full_block` tests can replicate remote blocks that include transactions that query the EIP-4788 parent beacon block root oracle.

For Cancun+ blocks, before creating the block builder, the replay tool now fetches the beacon root contract's storage slots from the remote chain at the target block number and writes them into the local state. This ensures that any transaction querying the oracle reads the correct values, matching the remote chain's execution.

### Details
- Define `BeaconRootStorageSlots` and `beacon_root_storage_slots()` in the EIP-4788 module to compute the ring buffer slot indices for a given timestamp
- In `run_full_block`, fetch the two relevant storage slots from remote via `get_storage_at` rpc request and apply them to local state before block construction
- Add block replay test for mainnet block `24_735_736` (the block from the issue report)

Fixes #1346

### DIFF
--- a/crates/blockchain/fork/src/eips/eip4788.rs
+++ b/crates/blockchain/fork/src/eips/eip4788.rs
@@ -1,4 +1,4 @@
-use edr_primitives::{address, bytes, Address, Bytecode, Bytes};
+use edr_primitives::{address, bytes, Address, Bytecode, Bytes, U256};
 use edr_state_api::{account::AccountInfo, StateDiff};
 
 /// The address of the beacon roots contract.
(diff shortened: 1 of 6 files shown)


crates/blockchain/fork/src/eips/eip4788.rs
```diff
@@ -9,6 +9,38 @@ pub const BEACON_ROOTS_BYTECODE: Bytes = bytes!(
...
+/// - Second buffer to store the corresponding parent beacon block roots.
+pub fn beacon_root_storage_slots(timestamp: u64) -> BeaconRootStorageSlots {
+    let timestamp_slot = U256::from(timestamp % HISTORY_BUFFER_LENGTH);
+    let beacon_root_slot = timestamp_slot + U256::from(HISTORY_BUFFER_LENGTH);
+    BeaconRootStorageSlots {
+        timestamp_slot,
+        beacon_root_slot,
+    }
+}
+
```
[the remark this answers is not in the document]
anaPerezGhiglia: No need to add all these tests. This piece of code is only used for the replay tool & tests - which are not the core use of EDR. I think that a single test that validates that slots are properly encoded in the struct is enough

crates/test/block_replay/src/lib.rs
```diff
@@ -375,6 +388,53 @@ pub async fn run_full_block<
...
+        timestamp_slot,
+        beacon_root_slot,
+    } = beacon_root_storage_slots(replay_header.timestamp);
+    let timestamp_value = rpc_client
+        .get_storage_at(
+            BEACON_ROOTS_ADDRESS,
+            timestamp_slot,
+            Some(BlockSpec::Number(block_number)),
+        )
+        .await?;
```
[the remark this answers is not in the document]
anaPerezGhiglia: I think that splitting the function in two does not pay off for this

crates/test/block_replay/Cargo.toml
```diff
@@ -24,6 +24,7 @@ edr_receipt.workspace = true
 edr_rpc_eth.workspace = true
 edr_state_api.workspace = true
 edr_utils.workspace = true
+futures = "0.3"
```
Wodann, line 27: Question: Why is `futures` required here? Did I miss an import?

crates/test/block_replay/src/lib.rs
```diff
@@ -375,6 +389,50 @@ pub async fn run_full_block<
     Ok(())
 }
 
+async fn replicate_beacon_block_root_oracle_state<
+    ChainSpecT: 'static
```
Wodann, line 393: Question: What is the motivation to replicate the beacon root storage slot calculation rather than querying it directly from the state of block `N` and storing those values in the state of block `N-1`?

crates/test/block_replay/Cargo.toml
anaPerezGhiglia: `lib.rs` is importing it to use `try_join!` to avoid waiting on two async calls sequentially

crates/test/block_replay/src/lib.rs
anaPerezGhiglia: This function copies the storage from block n and updates the local state (of block n-1) with those values. 
The "calculation" is needed to know _which_ storage index to copy from remote into the local state. The contract storage is a two-ring-buffer, and the index depends on the timestamp of the block.

What would you suggest we do instead?

Wodann (approved):

crates/test/block_replay/src/lib.rs
Wodann: Thanks for the explanation! It clarifies my knowledge gap. No need to change anything. Your approach sounds good 👍

PR #1362, opened 2026-04-15
chore: document spec compliance

anaPerezGhiglia:
## Summary

Add a `spec-compliance/` directory documenting EIPs and chain specifications that EDR does not support or only partially supports.

The main goal is to have public records of what's not supported in EDR and why, serving as a reference for EDR contributors, Hardhat power users, and future EDR-library consumers who integrate EDR outside of Hardhat. This is a starting point, not an exhaustive list. Rhe expectation is that it grows alongside EDR as new hardforks and chain specs are adopted.

## Structure

- `spec-compliance/README.md` — index with status definitions and tables grouped by chain type (L1 Ethereum, OP Stack) and ordered by hardfork
- `spec-compliance/l1/` — detailed docs for L1 Ethereum specs
- `spec-compliance/op/` — detailed docs for OP Stack specs

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,10 @@ EDR finds its origins in Hardhat Network but incorporates the lessons we have le
 - **Source-level code coverage** via Solidity instrumentation.
 - **Per-function and per-deployment gas reports** with proxy delegation chain tracking.
 
+## Spec Compliance
+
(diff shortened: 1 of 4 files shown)


popescuoctavian (approved):
LGTM, thanks for adding detailed explanations and examples!

Wodann (approved):

PR #1363, opened 2026-04-16
feat(coverage): own the Solidity library

anaPerezGhiglia:
## Summary
[Related design doc](https://www.notion.so/nomicfoundation/Move-coverage-library-ownership-to-EDR-327578cdeaf5805ba8cbf715e5fc7f5c)

EDR now owns the Solidity coverage library that instrumented contracts depend on. The library is bundled with the npm package and exposed through a new TS helper at `@nomicfoundation/edr/coverage`, which returns the library's source and expected filename.

This removes a surface where Hardhat and EDR could drift out of sync: the library, its contract name, and its file name are all defined in a single place (`edr_instrument::coverage`) and shipped together.

### Breaking changes
- `addStatementCoverageInstrumentation` JS function no longer accepts a library path argument. Callers that passed one must remove it; the library is bundled with EDR.

### Updates
- Library has been renamed from `__HardhatCoverage` to `__NomicFoundationCoverage`
- Include `coverage.sol` file into EDR npm module
- New TS API
  - expose `COVERAGE_LIBRARY_FILE_NAME` const  
  - `@nomicfoundation/edr/coverage` exports:
     - `interface CoverageLib { content: string; filename: string }`
     - `function getCoverageLibrary(): CoverageLib` — reads the bundled `coverage.sol` at runtime
 
### Package configuration

- **Incidental fix: `@nomicfoundation/edr/solidity-tests` was a broken export.** The subpath would have been unusable for any consumer - Node's exports resolution points at `./dist/src/ts/solidity_tests.js`, but that path wasn't in the published tarball. Fixed by adding `dist/` to files.   
- The subpath pointed at `./dist/src/ts/solidity_tests.js`, but `package.json`'s `files` array didn't include `dist/`, so the compiled JS never shipped. Any consumer trying to use it would have hit "module not found". Fixed by adding `dist/` to `files` (the new `./coverage` subpath would have had the same issue).
- New `crates/edr_napi/tsconfig.prod.json` scopes the publication build to `src/ts/` (flat output, excludes tests). `scripts/build_edr_napi.sh` uses this config too, so local `pnpm build:dev` and CI produce identical dist layouts that match the `exports` paths.
- `crates/edr_napi/package.json`: flattened the `exports` paths (`./dist/src/ts/*.js` → `./dist/*.js`) to match the flat build layout.

### Release workflow

- `.github/workflows/edr-npm-release.yml` `prepare` job now copies `data/contracts/coverage.sol` → `crates/edr_napi/coverage.sol` before the tar is built, so the bundled tarball and `pnpm publish` include it.
- Added a `Compile TypeScript helpers` step in the same job, so the compiled helpers exist before the tar is built.

### `edr_tool_solidity` cleanup

- Output extension is now `.bin` (matching `solc --bin` convention) instead of `.in`.
- Output directory renamed `data/deployed_bytecode/` → `data/deployment_bytecode/` — the files are deployment bytecode, not runtime/deployed.

### Test fixture and test-module reorganization

- Re-instrumented and recompiled test contracts
- Moved `Increment` test contract (`data/contracts/increment.sol` → `data/contracts/test/Increment.sol`). 
- `crates/edr_provider/tests/integration/coverage_scenarios.rs` merged into `coverage.rs` as an inner `returndata` module.


crates/edr_napi/src/ts/coverage.ts
```diff
@@ -0,0 +1,18 @@
...
+import { COVERAGE_LIBRARY_FILE_NAME } from "@nomicfoundation/edr";
+
+export interface CoverageLib {
+  content: string;
+  filename: string;
+}
+
+export function getCoverageLibrary(): CoverageLib {
+  const packageRoot = path.dirname(require.resolve("@nomicfoundation/edr"));
+  const sourcePath = path.join(packageRoot, "coverage.sol");
```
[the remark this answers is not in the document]
anaPerezGhiglia: This is not correct. 

URL: https://nodejs.org/api/packages.html#self-referencing-a-package-using-its-name                                                                                               
                                                                                                                                                                                    
  Key quotes from the Node docs:                                                                                                                                                    
                  
  ▎ "Within a package, the values defined in the package's package.json "exports" field can be referenced via the package's name."                                                  
                  
  ▎ "Self-referencing is available only if package.json has "exports", and will allow importing only what that "exports" (in the package.json) allows." 
  
URL: https://www.typescriptlang.org/docs/handbook/modules/reference.html#packagejson-imports-and-self-name-imports                                                                
   
  Key quote:                                                                                                                                                                        
                  
  ▎ "when package.json "exports" lookups are enabled, TypeScript will attempt to resolve import paths beginning with the current package name — that is, the value in the "name"    
  ▎ field of the nearest ancestor package.json of the importing file — through the "exports" field of that package.json."
                                                                                                                                                                                    
  Supported in moduleResolution: node16, nodenext, or bundler.                                                                                                                      
   
**Why it works in our case**                                                                                                                                                          
                  
  Our package.json has:                                                                                                                                                             
  - "name": "@nomicfoundation/edr"
  - "exports": { ".": "./index.js", ... }                                                                                                                                           
                                         
  And @tsconfig/node20 sets moduleResolution: "node16" (node20 extends node16 resolution). So both Node and TypeScript find the package via its own name, walking up from wherever  
  the importing file lives — no node_modules involvement needed.

data/contracts/test/Increment.sol
```diff
@@ -0,0 +1,18 @@
+pragma solidity ^0.8.26;
+import "../coverage.sol";
```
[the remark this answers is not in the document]
anaPerezGhiglia: This comment is right in spirit but is not accurate. As the file is being compiled in a flat-structured directory, the `..` has nothing to pop off, it's effectively a no-op , and the import normalizes to `coverage.sol`.
However this was brittle, so I changed the import and left a comment explaining

crates/edr_napi/src/ts/coverage.ts
```diff
@@ -0,0 +1,18 @@
...
+  return {
+    content: fs.readFileSync(sourcePath, "utf-8"),
+    filename: COVERAGE_LIBRARY_FILE_NAME,
```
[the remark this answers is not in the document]
anaPerezGhiglia: I added a check and a meaningful error if is not present

crates/edr_napi/package.json
```diff
@@ -31,12 +31,15 @@
...
+    "./solidity-tests": "./dist/solidity_tests.js",
+    "./coverage-library": "./dist/coverage_library.js"
   },
   "files": [
     "index.js",
     "index.d.ts",
-    "src/"
+    "src/",
+    "dist/",
+    "coverage.sol"
```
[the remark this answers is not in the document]
anaPerezGhiglia: I added back into the `scripts/build_edr_napi.sh` so any local script copies the file as well

crates/edr_napi/src/ts/coverage.ts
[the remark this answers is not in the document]
anaPerezGhiglia: The file is being copied on production publish and local build now, so it should always me present, so a meaningful message is enough

.changeset/icy-bugs-live.md
```diff
@@ -0,0 +1,5 @@
+---
+"@nomicfoundation/edr": minor
+---
+
+Bundled the Solidity coverage library with EDR. Added a `getCoverageLibrary()` helper at the `@nomicfoundation/edr/coverage-library` subpath that returns the library's source and expected filename. `addStatementCoverageInstrumentation` no longer accepts a library path argument.
```
Wodann:
```suggestion
- Added a `getCoverageLibrary()` helper at the `@nomicfoundation/edr/coverage-library` subpath that returns the library's source and expected filename. 
- Bundled the Solidity coverage library with EDR.
- Changed `addStatementCoverageInstrumentation` to no longer accept a library path argument.
```

book/src/02_development/01_tools.md
```diff
@@ -65,7 +65,7 @@ This tool compiles and instruments Solidity source files for use in EDR tests.
 
 Some integration tests require pre-compiled bytecode. To avoid adding `foundry-compilers` as a test dependency (which can interfere with other test crates that use solc during `cargo test --workspace`), we compile contracts ahead of time. It can also instrument source files with coverage probes, producing instrumented `.sol` files used by the `edr_solidity_tests` crate.
 
-By convention, compiled bytecode in EDR is kept in `data/deployed_bytecode/` as hex-encoded `.in` files. Integration tests load them via `include_str!` so they don't need solc at test time.
+By convention, compiled creation bytecode in EDR is kept in `data/creation_bytecode/` as hex-encoded `.bin` files (matching `solc --bin` output). Integration tests load them via `include_str!` so they don't need solc at test time.
```
Wodann: Suggestion: Conventionally, in HH/EDR we use two names for bytecode:

- Deployment bytecode; i.e. the bytecode that's deployed using a create call
- Runtime bytecode; i.e. the bytecode that's returned from a create call after the deployment bytecode has been run

I agree `deployed_bytecode` is confusing, but if we want to change the name of the directory, for consistency, we might want to stay with `deployment_bytecode`, which I think avoids the ambiguity of whether this is "bytecode that is deployed" vs "bytecode that was deployed".

crates/edr_coverage/tests/integration/e2e.rs
```diff
@@ -22,8 +22,8 @@ use revm_context::BlockEnv;
 
 const CHAIN_ID: u64 = 31337;
 
-const INCREMENT_DEPLOYED_BYTECODE: &str =
-    include_str!("../../../../data/deployed_bytecode/increment.in");
+const INCREMENT_CREATION_BYTECODE: &str =
```
Wodann: Suggestion: See https://github.com/NomicFoundation/edr/pull/1363/changes#r3112299172

crates/edr_napi/src/ts/coverage.ts
```diff
@@ -0,0 +1,23 @@
+import * as fs from "fs";
```
Wodann, line 1: Nit: For simplicity, I'd prefer reducing the file name to `coverage.ts`

crates/edr_napi/test/coverage.ts
```diff
@@ -26,15 +26,15 @@ class CoverageReporter {
   public hits: Uint8Array[] = [];
 }
 
-function readDeployedBytecode(): string {
-  const filePath = `${__dirname}/../../../data/deployed_bytecode/increment.in`;
+function readCreationBytecode(): string {
```
Wodann: Suggestion: see https://github.com/NomicFoundation/edr/pull/1363/changes#r3112299172

crates/edr_napi/package.json
```diff
@@ -31,12 +31,15 @@
...
-    "./solidity-tests": "./dist/src/ts/solidity_tests.js"
+    "./solidity-tests": "./dist/solidity_tests.js",
+    "./coverage-library": "./dist/coverage_library.js"
   },
   "files": [
     "index.js",
     "index.d.ts",
-    "src/"
+    "src/",
+    "dist/",
```
Wodann: Question: @popescuoctavian you previously added the `./dist/solidity_tests.js`. Do you remember whether there was a reason to not include the `dist` folder that time?

crates/edr_instrument/src/coverage.rs
```diff
@@ -261,46 +267,31 @@ mod tests {
...
-            FIXTURE,
-            "instrumentation.sol",
-            version,
-            "\"path/with/quotes.sol\"",
-        )
-        .expect_err("Expected an error for invalid import path");
-
-        assert!(matches!(
-            result,
-            InstrumentationError::InvalidLibraryPath { .. }
```
Wodann, line 285: Question: Do we still need this error? Has it now become an invariant of the code as we define the path, which can be `expect`ed?

crates/edr_napi/tsconfig.prod.json
```diff
@@ -0,0 +1,12 @@
+{
```
Wodann: Note: I have limited familiarity with these configurations. After some research, this looks good, but it would be good to ask for a second opinion from a TS expert (e.g. Pato)

crates/edr_provider/tests/integration/coverage.rs
```diff
@@ -17,8 +17,8 @@ use edr_solidity::contract_decoder::ContractDecoder;
 use parking_lot::{Mutex, RwLock};
 use tokio::runtime;
 
-const INCREMENT_DEPLOYED_BYTECODE: &str =
-    include_str!("../../../../data/deployed_bytecode/increment.in");
+const INCREMENT_CREATION_BYTECODE: &str =
```
Wodann: Suggestion: see https://github.com/NomicFoundation/edr/pull/1363/changes#r3112299172

crates/edr_provider/tests/integration/coverage_scenarios.rs
```diff
@@ -1,192 +0,0 @@
-#![cfg(feature = "test-utils")]
```
Wodann, line 1: Question: Why are these tests removed?

crates/tool/solidity/src/main.rs
```diff
@@ -34,11 +36,11 @@ const COVERAGE_LIBRARY_SOURCE_ID: &str = "coverage_lib.sol";
 /// data/contracts/test/CoverageTest.sol`
 ///
 ///   # Write bytecodes to files:
-///   `cargo run -p edr_tool_solidity -- --instrument -o data/deployed_bytecode
+///   `cargo run -p edr_tool_solidity -- --instrument -o data/creation_bytecode
```
Wodann: Suggestion: see https://github.com/NomicFoundation/edr/pull/1363/changes#r3112299172

crates/tool/solidity/src/main.rs
```diff
@@ -131,15 +133,19 @@ fn main() -> Result<()> {
 
     let output = compile_project(&root)?;
 
-    // Extract and output deployed bytecodes.
+    // Extract and output creation bytecodes.
```
Wodann: Suggestion: see https://github.com/NomicFoundation/edr/pull/1363/changes#r3112299172

data/deployment_bytecode/Increment.bin
```diff
@@ -1 +1 @@
...
...6004016100ba90610226565b60405180910390fd5b6100cd60026100e7565b805f808282546100dd9190610271565b9250508190555050565b6100fe816100f661010161012d565b63ffffffff16565b50565b5f73c0bec0bec0bec0bec0bec0bec0bec0bec0bec0be90506040518281525f80602083855afa50505050565b610138819050919050565b6101406102a4565b565b5f819050919050565b61015481610142565b82525050565b5f60208201905061016d5f83018461014b565b92915050565b5f80fd5b61018081610142565b811461018a575f80fd5b50565b5f8135905061019b81610177565b92915050565b5f602082840312156101b6576101b5610173565b5b5f6101c38482850161018d565b91505092915050565b5f82825260208201905092915050565b7f496e6372656d656e742073686f756c6420626520706f736974697665000000005f82015250565b5f610210601c836101cc565b915061021b826101dc565b602082019050919050565b5f6020820190508181035f83015261023d81610204565b9050919050565b7f4e487b71000000000000000000000000000000000000000000000000000000005f52601160045260245ffd5b5f61027b82610142565b915061028683610142565b925082820190508082111561029e5761029d610244565b5b92915050565b7f4e487b71000000000000000000000000000000000000000000000000000000005f52605160045260245ffdfea2646970667358221220333153e7c80c8bccc99a299000d97336ae35f9522bcd5a675800683958fbcc5f64736f6c634300081a0033
```
Wodann, line 1: Thought: The only difference is in the IPFS metadata hash embedded at the tail of the bytecode. Otherwise, no changes were introduced in the bytecode.

This is likely caused by a change in the compilation metadata.

Wodann (changes requested):
Thanks for making this change!

Some questions and suggestions

data/deployment_bytecode/Increment.bin
anaPerezGhiglia: The difference comes because the coverage library changed the API function name. 
Now the `Increment.sol` file (which's the origin of this binary file) calls to `__NomicFoundationCoverage.sendHit` instead of `__HardhatCoverage.sendHit`.

crates/edr_instrument/src/coverage.rs
anaPerezGhiglia: Good catch! I removed it, moved the import statement construction to a new function and unit-tested that instead. 
See [55349ff](https://github.com/NomicFoundation/edr/pull/1363/commits/55349ff8e463d78dbe68aa040aab5b9770dadc95)

crates/edr_provider/tests/integration/coverage_scenarios.rs
anaPerezGhiglia: Not deleted, just reorganized. From the PR description: 
> `crates/edr_provider/tests/integration/coverage_scenarios.rs` merged into `coverage.rs` as an inner returndata module.

btw: sorry for the PR description being so long!

book/src/02_development/01_tools.md
anaPerezGhiglia: addressed by [8543197](https://github.com/NomicFoundation/edr/pull/1363/commits/85431977550d68c98cc549d456c50f805dc5dec8)

crates/edr_coverage/tests/integration/e2e.rs
anaPerezGhiglia: addressed by [8543197](https://github.com/NomicFoundation/edr/pull/1363/commits/85431977550d68c98cc549d456c50f805dc5dec8)

crates/edr_napi/test/coverage.ts
anaPerezGhiglia: addressed by [8543197](https://github.com/NomicFoundation/edr/pull/1363/commits/85431977550d68c98cc549d456c50f805dc5dec8)

crates/edr_provider/tests/integration/coverage.rs
anaPerezGhiglia: addressed by [8543197](https://github.com/NomicFoundation/edr/pull/1363/commits/85431977550d68c98cc549d456c50f805dc5dec8)

crates/tool/solidity/src/main.rs
anaPerezGhiglia: addressed by [8543197](https://github.com/NomicFoundation/edr/pull/1363/commits/85431977550d68c98cc549d456c50f805dc5dec8)

crates/tool/solidity/src/main.rs
anaPerezGhiglia: addressed by [8543197](https://github.com/NomicFoundation/edr/pull/1363/commits/85431977550d68c98cc549d456c50f805dc5dec8)

crates/edr_napi/src/ts/coverage.ts
anaPerezGhiglia: done [e3aeb99](https://github.com/NomicFoundation/edr/pull/1363/commits/e3aeb9976806b4cc4540386bd3da94e4996edf39)

crates/edr_napi/tsconfig.prod.json
anaPerezGhiglia: I decided to create this profile to scope the publication build to `src/ts/`. That way we don't build test files and have a flat `dist` folder as a outcome. The other reason I did this is to be able to include `dist/` directory in package.json `files` attribute (instead of having to add `dist/src/ts`) 

@alcuadrado can you please take a look into this? I did this looking for industry-standards, but your opinion would be really valuable here

crates/edr_napi/package.json
popescuoctavian: @Wodann @anaPerezGhiglia I think that was overlooked when I added the JS code. I believe `dist/src` should be included here instead of `dist`, since that would be the minimum required path.

anaPerezGhiglia:
After what we've discussed during the tech-sync, I have rolled back to use a single tsconfig file in the project, include `dist/src` into the `files` field and add the corresponding mappings to `exports` field.

anaPerezGhiglia:
@Wodann this is ready for another review 🙇

crates/edr_napi/tsconfig.prod.json
anaPerezGhiglia: This file was removed in [8162124](https://github.com/NomicFoundation/edr/pull/1363/commits/81621245b5c2f6276174fc1013867639458a13c9)
See https://github.com/NomicFoundation/edr/pull/1363#issuecomment-4290672038

Wodann (approved):
Thanks for making these changes. LGTM!

PR #1364, opened 2026-04-20
fix(op): Jovian+ `base_fee` calculation

anaPerezGhiglia:
## Context 
When implementing #1213 I missed the impact the following sentence had in the `base_fee` calculation

> Furthermore, from Jovian, the base fee update calculation now uses `gasMetered := max(gasUsed, blobGasUsed)` in place of the `gasUsed` value used before. As a result, blocks with high DA usage may cause the base fee to increase in subsequent blocks.

source: [ Jovian spec da-footprint section](https://specs.optimism.io/protocol/jovian/exec-engine.html#da-footprint-block-limit) 

Niran pointed us to it on https://github.com/base/node/issues/883#issuecomment-4269635982

## Changes
This PR introduces the changes necessary for tweaking the OP base fee calculation to follow the spec accordingly. 

Note: Even if Jovian is still not fully implemented in EDR (`da_footprint` is not calculated), this change guarantees that if the `blob_gas` is overwritten properly (like in block-replay tool) the calculation matches Jovian spec

### DIFF
--- a/crates/block/header/src/lib.rs
+++ b/crates/block/header/src/lib.rs
@@ -346,6 +346,7 @@ impl PartialHeader {
                 Some(if let Some(parent) = &parent {
                     calculate_next_base_fee_per_gas(
                         parent,
+                        u128::from(parent.gas_used),
                         overrides
                             .base_fee_params
                             .as_ref()
@@ -594,6 +595,7 @@ pub fn overridden_block_number<HardforkT>(
 /// Panics if the parent header does not contain a base fee.
 pub fn calculate_next_base_fee_per_gas<HardforkT: PartialOrd>(
     parent: &BlockHeader,
+    gas_used: u128,
     base_fee_params: &BaseFeeParams<HardforkT>,
     hardfork: HardforkT,
 ) -> u128 {
@@ -605,7 +607,6 @@ pub fn calculate_next_base_fee_per_gas<HardforkT: PartialOrd>(
     // Adapted from https://github.com/alloy-rs/alloy/blob/main/crates/eips/src/eip1559/helpers.rs#L41
     // modifying it to support `u128`.
     // TODO: Remove once https://github.com/alloy-rs/alloy/issues/2181 has been addressed.
-    let gas_used = u128::from(parent.gas_used);
     let gas_limit = u128::from(parent.gas_limit);
 
     // In reality, [EIP-1559] specifies an initial base fee block number at which to
--- a/crates/block/header/src/lib.rs
+++ b/crates/block/header/src/lib.rs
@@ -587,8 +587,15 @@ pub fn overridden_block_number<HardforkT>(
     })
 }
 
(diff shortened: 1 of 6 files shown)


crates/block/header/src/lib.rs
```diff
@@ -594,6 +595,7 @@ pub fn overridden_block_number<HardforkT>(
 /// Panics if the parent header does not contain a base fee.
 pub fn calculate_next_base_fee_per_gas<HardforkT: PartialOrd>(
     parent: &BlockHeader,
+    gas_used: u128,
     base_fee_params: &BaseFeeParams<HardforkT>,
     hardfork: HardforkT,
 ) -> u128 {
```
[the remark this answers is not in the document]
anaPerezGhiglia: addressed

crates/edr_op/src/spec.rs
```diff
@@ -273,27 +273,41 @@ pub(crate) fn op_base_fee_params_for_block(
...
+        let base_fee_per_gas =
+            calculate_next_base_fee_per_gas(parent_header, gas_metered, base_fee_params, hardfork);
         let min_base_fee = decode_min_base_fee(&parent_header.extra_data)
             .expect("Jovian should have min base fee defined in extra data");
-        if base_fee_per_gas < min_base_fee {
-            min_base_fee
-        } else {
-            base_fee_per_gas
-        }
+        core::cmp::max(base_fee_per_gas, min_base_fee)
```
Wodann, line 306: Nit: this feels very compressed. Having newlines between multi-line statements might make it easier to visually parse.

crates/edr_op/src/spec.rs
```diff
@@ -471,4 +485,198 @@ mod tests {
...
+            assert!(
+                result > PARENT_BASE_FEE,
+                "base fee should increase when blob_gas_used exceeds the target"
+            );
+
+            // Equivalent parent with the blob usage moved into gas_used should yield
+            // the same next base fee.
+            let extra_data = encode_dynamic_base_fee_params_jovian(&BASE_FEE_PARAMS, 1);
+            let equivalent_parent = parent_with(2 * GAS_TARGET, None, extra_data);
+            let equivalent_result = super::super::op_next_base_fee(
```
Wodann: Suggestion(non-blocking): Consider a `use` statement for `op_next_base_fee` as it's used multiple times.

Wodann (approved):
LGTM!

Did you retest this with all prior failed block replays?

crates/edr_op/src/spec.rs
anaPerezGhiglia: actually, no extra use statement is needed. This was Claude over doing-it. Thanks for the catch!

anaPerezGhiglia:
> Did you retest this with all prior failed block replays?

I did, they all pass now. 
I just pushed the `test/base-mainnet-base_fee_per_gas` rebased on top of this one in case you want to take a look.

PR #1367, opened 2026-04-23
feat(solidity-tests): support EIP-712 cheatcodes

### LINKED ISSUE #1365: Implement EIP-712 cheatcodes
## Context
Port the following three cheatcodes to EDR:

```solidity
#[cheatcode(group = Utilities)]
function eip712HashType(string calldata typeNameOrDefinition) external pure returns (bytes32 typeHash);

#[cheatcode(group = Utilities)]
function eip712HashStruct(string calldata typeNameOrDefinition, bytes calldata abiEncodedData) external pure returns (bytes32 typeHash);

#[cheatcode(group = Utilities)]
function eip712HashTypedData(string calldata jsonData) external pure returns (bytes32 digest);
```

Intentionally left out:

```solidity
#[cheatcode(group = Utilities)]
function eip712HashType(string calldata bindingsPath, string calldata typeName) external pure returns (bytes32 typeHash);

#[cheatcode(group = Utilities)]
function eip712HashStruct(string calldata bindingsPath, string calldata typeName, bytes calldata abiEncodedData) external pure returns (bytes32 typeHash);

```

**Rationale:**
use in-memory representation of the EIP-712 canonical type definitions instead of requiring file access.

## Tech detail
To allow configuration of EIP-712 canonical type definitions, we add a new configuration field for Solidity tests:

```tsx
export interface SolidityTestRunnerConfigArgs {
  /**
   * A list of EIP-712 canonical type definitions that can be referenced
   * by type name in the `eip712HashType` and `eip712HashStruct` cheatcodes.
   *
   * The type of a struct is encoded as:
   *
   * `name ‖ "(" ‖ member₁ ‖ "," ‖ member₂ ‖ "," ‖ … ‖ memberₙ ")"`
   *
   * where each member is written as `type ‖ " " ‖ name`.
   *
   * For example, the following struct:
   * 
   * ```
	 * struct Mail {
	 *   address from;
	 *   address to;
	 *   string contents;
	 * }
   *
   * is encoded as `Mail(address from,address to,string contents)`.
   */
  eip712CanonicalTypes: Array<string>,
  ..
}
```

EDR will use these configured `eip712CanonicalTypes` for the new eip712* cheatcodes implementation

anaPerezGhiglia:
related [design doc](https://www.notion.so/nomicfoundation/EIP-712-cheatcodes-31a578cdeaf58091ae82c295a39d88d7) 
closes NomicFoundation/edr#1365 

## Summary

Implements three [EIP-712](https://eips.ethereum.org/EIPS/eip-712) Forge cheatcodes, ported from upstream foundry:

  - `vm.eip712HashType(string)` - canonical typeHash of an EIP-712 type definition.
  - `vm.eip712HashStruct(string, bytes)` - EIP-712 hashStruct for a type + ABI-encoded payload.
  - `vm.eip712HashTypedData(string)` - full `keccak256("\x19\x01" ‖ domainSeparator ‖ hashStruct(message))` digest from a JSON typed-data payload.

The cheatcode bodies are an almost-exact copy-paste of the foundry implementation. One deliberate divergence is described below (see [Why reify `Eip712TypeDef`](#why-reify-eip712typedef)).

The filesystem-reading overloads (`eip712HashType(string, string)` / `eip712HashStruct(string, string, bytes)`) remain unsupported: EDR intentionally avoids disk access for these, per the scope discussed in the issue. Callers hitting those overloads will keep getting a targeted `UnsupportedCheatcode` error with the signature in the message.

## API specs

`eip712HashType` / `eip712HashStruct` accept either an inline type definition or a registered type name. Name resolution goes through a new runner config field:

```ts
solidityTestRunnerConfig.eip712CanonicalTypes: string[]
```

Each entry is an independent, self-contained EIP-712 type definition. Nested struct types referenced by the primary type must be inlined in the same string. Only the primary (leftmost) type is registered by name; nested sub-types are *not* auto-exposed. To reach a sub-type by name, the user should supply it as its own top-level entry:

```ts
eip712CanonicalTypes: [
  "Mail(Person from,Person to,string contents)Person(address wallet,string name)",
  "Person(address wallet,string name)",
]
```

Entries are validated at runner startup, not on first cheatcode call. Callers can expect:

- **Non-canonical input is accepted and normalized.** Entries don't need to already be in canonical form — whitespace and referenced-type ordering are fixed up on ingest, and the stored form is always the canonical one.
- **Every bad entry is reported in a single error.** If the list contains multiple unparseable or unresolvable entries, startup fails with all of them enumerated.
- **Duplicate primary-type names are rejected.** If the list contains two entries for the same primary type — even with identical bodies — startup fails. Silently letting the last one win would be dangerous: EIP-712 hashes are cryptographic, so a wrong typeHash produces a wrong signature with no visible trace back to the config.

### Not auto-exploding EIP-712 types
Callers setting the `eip712CanonicalTypes` config are already doing the full type-discovery walk on their side — that's how they know which types to register in the first place. 
Having EDR re-expand nested types would duplicate that work. It would also risk clashing with types the caller deliberately registered at the top level: EDR doesn't know where any of these types originate, so it's not in a position to make registration decisions about nested ones.

### Why reify `Eip712TypeDef`

Foundry resolves type names by reading a bindings file (generated by `forge bind-json`) on disk at cheatcode-call time. `get_canonical_type_def` returns a plain `String` each call, and the primary-type name is recovered where needed via a defensive string slice: `type_def[..type_def.find('(').unwrap_or(type_def.len())]`.

EDR can't do that — disk access is intentionally off the table — so name resolution has to go through the in-memory `eip712CanonicalTypes` runner config. Validation therefore happens up front at config load, not lazily per call, and the validated results get stored for reuse.

Once we're storing validated values, the question is *what* to store. A plain `HashMap<String, String>` works, but:

- anyone receiving a `&str` from the map has to trust the canonical-form property on faith;
- call sites still have to string-slice the value to recover the primary-type name.

Reifying `Eip712TypeDef` removes both at the type level:

```rust
pub struct Eip712TypeDef {
    name: String,                  // private
    canonical_definition: String,  // private
}

impl Eip712TypeDef {
    pub fn parse(input: &str) -> Result<Self, Eip712Error> { ... }
    pub fn name(&self) -> &str { ... }
    pub fn canonical_definition(&self) -> &str { ... }
}
```

Private fields plus a single public constructor means any `Eip712TypeDef` anywhere in the program is guaranteed to have gone through `parse`, which canonicalizes the input. What this buys us:

- **Compile-time guarantee that the canonical-form invariant holds.** Not a convention upheld by inspection. Anyone holding one can rely on it.
- **No string slicing at call sites** to recover the primary-type name. It's a field extracted from the parser's AST, not a substring re-carved defensively each time.
- **Single source of truth for parse + canonicalize + name-extract.** `Eip712TypeDef::parse` is the only path; both config-time ingestion and cheatcode-time inline parsing funnel through it, so the three-step logic lives in exactly one place.
- **No reconstruction at call sites.** The config stores `HashMap<String, Eip712TypeDef>` directly, so cheatcode impls clone from the map instead of rebuilding the type.

### DIFF
--- a/crates/foundry/cheatcodes/Cargo.toml
+++ b/crates/foundry/cheatcodes/Cargo.toml
@@ -51,7 +51,7 @@ revm.workspace = true
 revm-inspectors.workspace = true
 semver.workspace = true
 serde_json.workspace = true
-strum.workspace = true
+strum = { workspace = true, features = ["derive"] }
 thiserror.workspace = true
 toml = { workspace = true, features = ["preserve_order"] }
 tracing.workspace = true
--- a/crates/foundry/cheatcodes/assets/cheatcodes.json
+++ b/crates/foundry/cheatcodes/assets/cheatcodes.json
@@ -4336,7 +4336,7 @@
     {
       "func": {
         "id": "eip712HashStruct_0",
-        "description": "Generates the struct hash of the canonical EIP-712 type representation and its abi-encoded data.\nSupports 2 different inputs:\n 1. Name of a primary type previously declared via the\n    `eip712CanonicalTypes` runner configuration\n 2. String representation of the type (i.e. \"Foo(Bar bar) Bar(uint256 baz)\").\n    * Note: the cheatcode will use the canonical type even if the input is malformed\n            with the wrong order of elements or with extra whitespaces.",
+        "description": "Generates the struct hash of the canonical EIP-712 type representation and its abi-encoded data.\nSupports 2 different inputs:\n 1. Name of a primary type previously declared via the\n    `eip712CanonicalTypes` runner configuration\n 2. String representation of the type (i.e. \"Foo(Bar bar) Bar(uint256 baz)\").\n    Note: the cheatcode will use the canonical type even if the input is malformed\n          with the wrong order of elements or with extra whitespaces.",
         "declaration": "function eip712HashStruct(string calldata typeNameOrDefinition, bytes calldata abiEncodedData) external pure returns (bytes32 typeHash);",
         "visibility": "external",
         "mutability": "pure",
@@ -4376,7 +4376,7 @@
     {
       "func": {
         "id": "eip712HashType_0",
-        "description": "Generates the hash of the canonical EIP-712 type representation.\nSupports 2 different inputs:\n 1. Name of a primary type previously declared via the\n    `eip712CanonicalTypes` runner configuration\n 2. String representation of the type (i.e. \"Foo(Bar bar) Bar(uint256 baz)\").\n    * Note: the cheatcode will output the canonical type even if the input is malformed\n            with the wrong order of elements or with extra whitespaces.",
+        "description": "Generates the hash of the canonical EIP-712 type representation.\nSupports 2 different inputs:\n 1. Name of a primary type previously declared via the\n    `eip712CanonicalTypes` runner configuration.\n 2. String representation of the type (i.e. \"Foo(Bar bar) Bar(uint256 baz)\").\n    Note: the cheatcode will output the canonical type even if the input is malformed\n          with the wrong order of elements or with extra whitespaces.",
         "declaration": "function eip712HashType(string calldata typeNameOrDefinition) external pure returns (bytes32 typeHash);",
         "visibility": "external",
         "mutability": "pure",
--- a/crates/foundry/cheatcodes/assets/cheatcodes.json
+++ b/crates/foundry/cheatcodes/assets/cheatcodes.json
@@ -4336,7 +4336,7 @@
     {
       "func": {
         "id": "eip712HashStruct_0",
-        "description": "Generates the struct hash of the canonical EIP-712 type representation and its abi-encoded data.\nSupports 2 different inputs:\n 1. Name of the type (i.e. \"PermitSingle\")\n    * Requires previous `eip712CanonicalTypes` configuration\n 2. String representation of the type (i.e. \"Foo(Bar bar) Bar(uint256 baz)\").\n    * Note: the cheatcode will use the canonical type even if the input is malformated\n            with the wrong order of elements or with extra whitespaces.",
+        "description": "Generates the struct hash of the canonical EIP-712 type representation and its abi-encoded data.\nSupports 2 different inputs:\n 1. Name of a primary type previously declared via the\n    `eip712CanonicalTypes` runner configuration\n 2. String representation of the type (i.e. \"Foo(Bar bar) Bar(uint256 baz)\").\n    * Note: the cheatcode will use the canonical type even if the input is malformed\n            with the wrong order of elements or with extra whitespaces.",
         "declaration": "function eip712HashStruct(string calldata typeNameOrDefinition, bytes calldata abiEncodedData) external pure returns (bytes32 typeHash);",
         "visibility": "external",
         "mutability": "pure",
@@ -4376,7 +4376,7 @@
     {
       "func": {
         "id": "eip712HashType_0",
-        "description": "Generates the hash of the canonical EIP-712 type representation.\nSupports 2 different inputs:\n 1. Name of the type (i.e. \"Transaction\").\n    * Requires previous `eip712CanonicalTypes` configuration\n 2. String representation of the type (i.e. \"Foo(Bar bar) Bar(uint256 baz)\").\n    * Note: the cheatcode will output the canonical type even if the input is malformated\n            with the wrong order of elements or with extra whitespaces.",
+        "description": "Generates the hash of the canonical EIP-712 type representation.\nSupports 2 different inputs:\n 1. Name of a primary type previously declared via the\n    `eip712CanonicalTypes` runner configuration\n 2. String representation of the type (i.e. \"Foo(Bar bar) Bar(uint256 baz)\").\n    * Note: the cheatcode will output the canonical type even if the input is malformed\n            with the wrong order of elements or with extra whitespaces.",
         "declaration": "function eip712HashType(string calldata typeNameOrDefinition) external pure returns (bytes32 typeHash);",
         "visibility": "external",
         "mutability": "pure",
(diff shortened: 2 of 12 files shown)


crates/foundry/cheatcodes/spec/src/vm.rs
```diff
@@ -2525,6 +2525,34 @@ interface Vm {
...
 
+    /// Generates the hash of the canonical EIP-712 type representation.
+    ///
+    /// Supports 2 different inputs:
+    ///  1. Name of a primary type previously declared via the
+    ///     `eip712CanonicalTypes` runner configuration.
+    ///
+    ///  2. String representation of the type (i.e. "Foo(Bar bar) Bar(uint256 baz)").
+    ///     Note: the cheatcode will output the canonical type even if the input is malformed
+    ///           with the wrong order of elements or with extra whitespaces.
```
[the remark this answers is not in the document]
anaPerezGhiglia: addressed

crates/foundry/cheatcodes/assets/cheatcodes.json
```diff
@@ -4376,7 +4376,7 @@
     {
       "func": {
         "id": "eip712HashType_0",
-        "description": "EIP-712 cheatcodes are not supported.",
+        "description": "Generates the hash of the canonical EIP-712 type representation.\nSupports 2 different inputs:\n 1. Name of a primary type previously declared via the\n    `eip712CanonicalTypes` runner configuration.\n 2. String representation of the type (i.e. \"Foo(Bar bar) Bar(uint256 baz)\").\n    Note: the cheatcode will output the canonical type even if the input is malformed\n          with the wrong order of elements or with extra whitespaces.",
```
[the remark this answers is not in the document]
anaPerezGhiglia: addressed

Wodann:
> ported from upstream foundry

Question: What's the commit or version you used to port the cheatcodes?

anaPerezGhiglia:
> Question: What's the commit or version you used to port the cheatcodes?

I did it from `master` commit f15108c7c72b1fb93df1c7ef0f9d47bfceca45f2.

anaPerezGhiglia:
Looking at the compare between `v1.3.6` and that commit there wasn't any modifications to the logic of the eip712 cheatcodes

```sh
git diff v1.3.6 f15108c -- crates/cheatcodes/src/utils.rs
```

Or in GH https://github.com/foundry-rs/foundry/compare/v1.3.6...f15108c7c72b1fb93df1c7ef0f9d47bfceca45f2

crates/foundry/cheatcodes/spec/src/vm.rs
```diff
@@ -2775,25 +2803,15 @@ interface Vm {
 
     // -------- Utilities --------
 
-    /// EIP-712 cheatcodes are not supported.
-    #[cheatcode(group = Utilities, status = Unsupported)]
-    function eip712HashType(string calldata typeNameOrDefinition) external pure returns (bytes32 typeHash);
-
     /// EIP-712 cheatcodes are not supported.
     #[cheatcode(group = Utilities, status = Unsupported)]
     function eip712HashType(string calldata bindingsPath, string calldata typeName) external pure returns (bytes32 typeHash);
```
Wodann, line 2808: Question: @popescuoctavian is it possible to add a custom error message for unsupported cheatcodes. E.g. in this case, we could say:

> Providing a path to a bindings file is not supported. Please use the `eip712HashType(string calldata typeNameOrDefinition)` cheatcode instead.

crates/foundry/cheatcodes/Cargo.toml
```diff
@@ -51,7 +51,7 @@ revm.workspace = true
 revm-inspectors.workspace = true
 semver.workspace = true
 serde_json.workspace = true
-strum.workspace = true
+strum = { workspace = true, features = ["derive"] }
```
Wodann, line 54: Question: Is this fixing an existing issue, or is it related to this PR? If the latter, where is the strum derive used?

Wodann (approved):
Great implementation! I like the design, especially the usage of the `parse` function to guarantee that all canonical EIP-712 type definition are valid

crates/foundry/cheatcodes/Cargo.toml
anaPerezGhiglia: Sorry I forgot to mention it. This is fixing an existing issue. 
The crate depends on the derive feature but it was not including it. This is the output of running `cargo check -p foundry-cheatcodes` on `main`
```sh
error[E0433]: failed to resolve: could not find `Display` in `strum`
   --> crates/foundry/cheatcodes/src/error.rs:100:37
    |
100 | #[derive(Debug, Clone, Copy, strum::Display)]
    |                                     ^^^^^^^ could not find `Display` in `strum`
```

This didn't surface before since when compiling the whole workspace this doesn't happen because of feature unification, and the `cargo-hack` check we have in the CI only validates `edr_*` crates (which I think we should change)

Wodann: Thanks for the explanation!

crates/foundry/cheatcodes/spec/src/vm.rs
anaPerezGhiglia: For what I understand, not with the current implementation.
`CheatcodeErrorCode::UnsuportedCheatcode` in a unit variant and `CheatcodeErrorDetails` does not have a `details` (or similar) field. 

```rust
pub struct CheatcodeErrorDetails {
    pub code: CheatcodeErrorCode,
    pub cheatcode: String,
}


pub enum CheatcodeErrorCode {
    UnsupportedCheatcode,
    MissingCheatcode,
}
```

but let's wait for Octavian reply to confirm

Wodann: Feel free to merge. If it turns out to be possible, we can add it in a follow-up PR.

popescuoctavian: @anaPerezGhiglia @Wodann that's not possible with the current implementation. We can look into extending the `CheatcodeErrorDetails` struct for situations like this one, where we support another variant of the cheatcode.

Wodann: I've created an issue for this: https://github.com/NomicFoundation/edr/issues/1371

PR #1369, opened 2026-04-27
fix: avoid executing setUp for empty test suites

Wodann:
We currently have a behavioural discrepancy with Foundry. In EDR, if a user passes a `test_pattern` regex that matches no tests in a given suite, that suite's `setUp` still runs (and could revert / waste work). In Foundry, that `setUp` will be skipped.

After the fix, the suite must still be **reported as completed** (callback fires, suite appears in the collected results) but with zero work done.

Commit 976ca835 reorders work in `crates/edr_solidity_tests/src/runner.rs` so that test-function filtering happens **before** the setup phase. If the filter excludes every function in a suite, the runner now short-circuits with an empty `SuiteResult` (empty `setup_traces`, empty `test_results`) instead of running `setUp`.

## Testing

Added a new test that exercises the empty-suite path against the existing `FailingSetup.t.sol` fixture. That contract's `setUp` calls `require(false, "setup failed predictably")`, so it serves as a tripwire — if the regression returns, `setUp()` will appear in `test_results` with that reason.

We also already have a *positive* control covering the unfiltered case: `test_core` at core.rs:32-40 asserts that `FailingSetupTest` reports `setUp()` failure with reason `"setup failed predictably"` when the test runs normally. So the new test is the *negative* counterpart of that existing assertion.


popescuoctavian (approved):
LGTM, thanks for adding this!

PR #1374, opened 2026-04-30
ci/improve caching

nebasuke:
## Summary

Master caching/CI-hygiene branch. Will be split into smaller landable PRs once the experiment settles. Two main shifts:

- Replace Swatinem (via `actions-rust-lang/setup-rust-toolchain` bundled cache) with **slang-style `actions/cache` composites** for Rust deps + workspace target. Reason: newer Swatinem versions prune `target/` more aggressively (warm-cache compile time on Linux jumped 1m 40s → 5m 32s and the cache shrank 1171 MB → 714 MB, not coincidentally). slang's pattern saves the full `target/` archive verbatim — no surprise pruning.
- **Probe (Option C)**: enable `RUSTFLAGS=-Dwarnings` + `RUSTDOCFLAGS=-Dwarnings` on test/integration/hardhat jobs. Surfaces any latent warnings in test/cfg(test) code paths. Easily revertable if it goes red.

Plus several smaller cleanups (RPC cache restore-keys, redundant manual benchmark cache, Hardhat-tests joining the `napi-coverage` bucket, comments documenting cargo-hack scope, doctest features broadened).

Full research, methodology and follow-up backlog in `edr-caching-research.md` (workspace-local, not in this PR).

### Cache budget

- 10 GB → 50 GB → **100 GB** (bumped twice during this work).
- Pre-bump observed state: 11.7 GB used, 17 active entries — frequent LRU churn.
- Today: ample headroom, ~93 entries.

### Per-job Rust cache fragmentation (single ref)

Linux x86_64 entries on `refs/pull/1373/merge` (one PR, 13 distinct Rust caches, ~8 GB):

```
1171 MB  v0-rust-test-edr-rs-Linux-x64                 (edr-ci.yml)
 875 MB  v0-rust-js-scenario-benchmark-Linux-x64       (edr-benchmark.yml)
 862 MB  v0-rust-benchmarks-test-Linux-x64             (edr-benchmark.yml)
 802 MB  v0-rust-js-soltests-benchmark-Linux-x64       (edr-benchmark.yml)
 715 MB  v0-rust-test-edr-ts-Linux-x64                 (edr-ci.yml)
 715 MB  v0-rust-edr-integration-tests-Linux-x64       (edr-ci.yml)
 714 MB  v0-rust-run-hardhat-tests-Linux-x64           (hardhat-tests.yml)
 709 MB  v0-rust-cargo-hack-edr-Linux-x64              (edr-ci.yml)
[...]

PR #1375, opened 2026-05-01
ci: gate test/integration jobs on -Dwarnings

nebasuke:
Note: still stacked on #1376.

## Summary

Adds `RUSTFLAGS=-Dwarnings` to test/integration jobs and reworks where per-target rustflags live so the gate doesn't drop `+crt-static` from Windows test builds. Unifying `RUSTFLAGS` is also a precondition for #1374's caching work — Cargo's fingerprint hashes the value, so consistent env across jobs is needed before bucket consolidation can land.

Stacked on #1376, which fixed cargo-llvm-cov's contribution to the same `+crt-static`-clobbering bug; this PR closes the remaining sources.

## Changes

1. **Warning gate.** `RUSTFLAGS=-Dwarnings` + `RUSTDOCFLAGS=-Dwarnings` on `test-edr-rs`, `test-edr-ts`, `edr-integration-tests`, `run-hardhat-tests`. Previously only `check-edr` / `cargo-hack-edr` enforced this; warnings in `cfg(test)` code slipped through across all 3 OSes.
2. **Workspace-root rustflags.** Move `target.<triple>.rustflags` from `crates/edr_napi/.cargo/config.toml` to the workspace-root `.cargo/config.toml`.
3. **setup-rust augmentation.** On Windows runners, append `-C target-feature=+crt-static` to `RUSTFLAGS` so it survives any env value set later.

## Why (2) and (3)

Cargo's rustflags resolution is "first source wins, no merging." A job env `RUSTFLAGS` clobbers `target.<triple>.rustflags` from `.cargo/config.toml` wholesale. Two compounding problems made test builds diverge from the release ABI on Windows:

- **Pre-existing**: `test-edr-rs` runs cargo from the repo root, where the package-local config is never read — its Windows builds have always linked the CRT dynamically while the released NAPI links statically. Change (1) didn't introduce this; it surfaced it.
- **Would be introduced by (1) on this branch alone**: `test-edr-ts` and `run-hardhat-tests` invoke cargo from inside `crates/edr_napi/` (via napi build), where the package-local config *is* loaded — but the warning gate's env `RUSTFLAGS` would clobber it.

Change (2) fixes the first: workspace-root config is found by repo-root invocations. Change (3) fixes the second: the target feature is baked into `RUSTFLAGS` *before* the gate is set, so the env value already includes it. Stacked on #1376 because cargo-llvm-cov 0.7+ no longer mutates `RUSTFLAGS` itself, leaving the warning gate as the only remaining clobber source.

The released artifact is unaffected: `edr-npm-release.yml` sets no `RUSTFLAGS` and runs from `crates/edr_napi/`, so config loads cleanly. Same logic generalizes to `aarch64-unknown-linux-musl` (`-crt-static`) for completeness, though no test job runs on a musl host.

### DIFF
--- a/.github/workflows/hardhat-tests.yml
+++ b/.github/workflows/hardhat-tests.yml
@@ -13,6 +13,12 @@ jobs:
   run-hardhat-tests:
     name: Run Hardhat tests (${{ matrix.os }}, Node ${{ matrix.node }})
     runs-on: ${{ matrix.os }}
+    # See comment on test-edr-rs in edr-ci.yml. Same crt-static caveat
+    # applies on Windows native — release artifact is built by
+    # edr-npm-release.yml separately and is unaffected.
+    env:
+      RUSTFLAGS: -Dwarnings
+      RUSTDOCFLAGS: -Dwarnings
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/hardhat-tests.yml
+++ b/.github/workflows/hardhat-tests.yml
@@ -13,9 +13,7 @@ jobs:
   run-hardhat-tests:
     name: Run Hardhat tests (${{ matrix.os }}, Node ${{ matrix.node }})
     runs-on: ${{ matrix.os }}
-    # See comment on test-edr-rs in edr-ci.yml. Same crt-static caveat
-    # applies on Windows native — release artifact is built by
-    # edr-npm-release.yml separately and is unaffected.
+    # See comment on test-edr-rs in edr-ci.yml re crt-static handling.
     env:
       RUSTFLAGS: -Dwarnings
       RUSTDOCFLAGS: -Dwarnings
--- a/.github/workflows/hardhat-tests.yml
+++ b/.github/workflows/hardhat-tests.yml
@@ -63,7 +63,7 @@ jobs:
           DEBUG: "hardhat:core:hardhat-network:*"
         shell: bash
         run: |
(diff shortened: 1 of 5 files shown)


.github/workflows/hardhat-tests.yml
```diff
@@ -13,6 +13,10 @@ jobs:
   run-hardhat-tests:
     name: Run Hardhat tests (${{ matrix.os }}, Node ${{ matrix.node }})
     runs-on: ${{ matrix.os }}
+    # See comment on test-edr-rs in edr-ci.yml re crt-static handling.
+    env:
+      RUSTFLAGS: -Dwarnings
+      RUSTDOCFLAGS: -Dwarnings
```
Wodann, line 19: Question: Do we need to add these flags for the Hardhat TS tests, since we're merely building the binaries? Would the `edr-ci.yml` job not have caught all issues that could arise here?

Wodann (approved):
Thanks for catching this and your work on speeding up our CI! 🙏 

I have one question that could result in a small change but otherwise this LGTM!

.github/workflows/hardhat-tests.yml
nebasuke: No, we do not need these flags, and yes the job would have already caught them (so it's a good question). The reason to add the flags is to try and unify flags as much as possible (see also the llvm-cov change) so we can share as much of the cached artefacts as possible (as different flags force rebuilds).

PR #1377, opened 2026-05-03
ci: cap mocha test jobs at 30m and dump active handles on exit

nebasuke:
## Status

**Draft / debug branch.** Diagnostic experiments live alongside the actual CI fix. The experiment commits will be stripped before merge — once we have a real fix in `edr_napi`, this PR keeps only the `timeout-minutes:30` change and (TBD) the always-on diagnostics.

## What was happening

`Test EDR TS bindings` and `Run Hardhat tests` jobs intermittently hang for the full 6h GHA ceiling — observed across Windows, macOS, and at least once on Ubuntu. Per-test mocha timeout is 25s, but the failure mode is *total log silence* until the runner kills the job, so even mocha's own timer doesn't fire.

The hang lands in `Provider logs > Interval mining` in `crates/edr_napi/test/logs.ts`. Exact `it()` that wedges varies — sometimes test 3 (`should stop collapsing when a different method is called`), sometimes test 4 (`should print a block with one transaction`). Consistent with a timing-sensitive deadlock, not a deterministic bug location.

## Root cause

The `Interval mining` describe block declares **one shared `mockTimer`** at the outer scope and creates a fresh `Provider` in every `beforeEach` without disposing the previous one. `crates/edr_napi/src/provider.rs` defines no `Drop` impl and exposes no `close()` to JS, so the previous `Provider`'s tokio tasks (still subscribed to the shared mockTimer) keep running. With several providers all reacting to `mockTimer.addSeconds(...)` and contending on the mining notifier, one tick deadlocks deeply enough that the JS event loop itself wedges — neither mocha's per-test setTimeout nor any other timer can fire.

## Empirical confirmation

Commit `99d12d153` (afterEach that drops the provider reference, calls `globalThis.gc()`, and yields 100ms) suppressed the hang across multiple CI runs. Reverting just that commit — keeping the multiplied-trigger amplification and the diagnostic probes — produced a hang on the very next Windows TS-bindings run:

> <https://github.com/NomicFoundation/edr/actions/runs/25279523193/job/74132774239?pr=1377> — 30 m 18 s, killed at the new 30 m cap.

That swing is what proves the leaked-providers theory: with the afterEach the bug doesn't fire; without it, it does.

What the new probes captured on the failing run:

```
[mocha-watchdog] worker online
[mocha-watchdog] worker started, pid=8456
[mocha-start] should only print the mined block when there are no pending txs
✔ (1137ms)
[mocha-start] should collapse the mined block info
✔ (2168ms)
[mocha-start] should stop collapsing when a different method is called   ← test 3 starts
##[error] cancelled                                                       ← 22m 22s of silence
```

The watchdog *worker* started successfully and printed twice — but no further output from any thread arrives during the hang. No `[mocha-main-alive]` (expected every 10 s ≈ 130 lines) and no `[mocha-watchdog] main-thread silent` (expected every 30 s ≈ 44 lines). Speculation: on Windows the parent and worker share an stderr pipe and writes serialize; if the main thread deadlocks while holding the console handle, the worker blocks too. Out of scope for this PR — it's a *separate* observability problem with the diagnostic infra, not part of the bug. To peer into the hang itself we'd need an external watchdog (sidecar process polling `/proc`-equivalent on Windows).

## What's actually shipping

Two pieces, both already in the branch as commit `c615d6159`:

1. **`timeout-minutes: 30`** on `test-edr-ts` and `run-hardhat-tests`. Caps a hang at ~30 min instead of 6 h — already validated end-to-end by the failing run linked above (job died at exactly 30 m).
2. **Post-suite handle dump + force-exit** in each suite's `setup.ts`. After the last test, dumps `process.getActiveResourcesInfo()` and `process.exit(0)` 5s later. Defensive: doesn't help with the deadlock-style hang we just diagnosed (the loop is wedged before the suite ends), but covers the other failure shape where tests pass and the process won't exit.

The probes from `fc897f0f0` (`[mocha-start]` logging, worker watchdog) and `a6c8b785f` (worker lifecycle handlers + main-thread alive pings) are useful and could stay as permanent diagnostics — TBD whether to keep them or strip before merge.

## The actual fix lives elsewhere

The leaked-providers theory points at a missing API in `edr_napi`. To do this properly:

- **`crates/edr_napi/src/provider.rs`** — implement `Drop` on `Provider` that aborts in-flight tokio tasks (currently the default `Drop` just drops the `Arc`s without canceling anything spawned).
- **Expose `#[napi] pub async fn close(&self)`** — explicit JS-callable disposal, awaited so callers get deterministic cleanup. Doesn't rely on JS GC.
- Replace the `globalThis.gc()`-based `afterEach` in `logs.ts` with `await provider.close()`.

That fix will be a separate PR. This PR's only role is to (a) stop the bleeding for CI minutes via the 30 m cap and (b) document the diagnosis so the followup can land cleanly.

## Commit chronology (will be cleaned up before merge)

| SHA | What | Status |
|---|---|---|
| `c615d6159` | `timeout-minutes:30` + post-suite handle dump | **keep** |
| `fc897f0f0` | `[mocha-start]` logging + worker-thread watchdog | keep or strip |
| `99d12d153` | afterEach: drop provider + force GC + `--expose-gc` | drop — replaced by Rust fix |
| `0e51c3672` | isolated copy of failing test in own describe | drop (debug-only) |
| `a6c8b785f` | self-instrument the probes (lifecycle handlers, main-alive ping) | keep or strip |
| `1015d21e3` | multiply test 4 by 20× to amplify CI repro | drop (debug-only) |
| `4e1707892` | revert `99d12d153` to confirm the experiment was the fix | drop |

### DIFF
--- a/.github/workflows/edr-ci.yml
+++ b/.github/workflows/edr-ci.yml
@@ -159,6 +159,9 @@ jobs:
     name: Test EDR TS bindings (${{ matrix.os }})
     runs-on: ${{ matrix.os }}
     needs: check-edr
+    # Cap below GHA's 6h ceiling so the recurring mocha-process-hang surfaces
+    # within ~2× normal runtime instead of burning a full job slot.
+    timeout-minutes: 30
     # See comment on test-edr-rs above re crt-static handling.
     env:
       RUSTFLAGS: -Dwarnings
--- a/.github/workflows/hardhat-tests.yml
+++ b/.github/workflows/hardhat-tests.yml
@@ -13,6 +13,9 @@ jobs:
   run-hardhat-tests:
     name: Run Hardhat tests (${{ matrix.os }}, Node ${{ matrix.node }})
     runs-on: ${{ matrix.os }}
+    # Cap below GHA's 6h ceiling so the recurring mocha-process-hang surfaces
+    # within ~2× normal runtime instead of burning a full job slot.
+    timeout-minutes: 30
     # See comment on test-edr-rs in edr-ci.yml re crt-static handling.
     env:
       RUSTFLAGS: -Dwarnings
--- a/crates/edr_napi/.mocharc.json
+++ b/crates/edr_napi/.mocharc.json
@@ -1,4 +1,5 @@
 {
   "require": "ts-node/register/transpile-only",
+  "file": "./test/setup.ts",
   "timeout": 25000
 }
--- a/crates/edr_napi/package.json
+++ b/crates/edr_napi/package.json
@@ -74,7 +74,7 @@
     "lint:fix": "pnpm run prettier --write",
     "pretest": "pnpm build:dev",
     "prettier": "prettier --check \"test/**.ts\"",
-    "test": "node --max-old-space-size=8192 --expose-gc node_modules/mocha/bin/_mocha --recursive \"test/**/*.ts\"",
+    "test": "node --max-old-space-size=8192 node_modules/mocha/bin/_mocha --recursive \"test/**/*.ts\"",
     "testNoBuild": "node --max-old-space-size=8192 node_modules/mocha/bin/_mocha --recursive \"test/**/{,!(logs|mock)}.ts\"",
     "universal": "napi universal",
     "version": "napi version"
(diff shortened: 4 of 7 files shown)


nebasuke:
Closed in favour of https://github.com/NomicFoundation/edr/pull/1378

PR #1379, opened 2026-05-04
ci(npm-release): cache cargo on non-release runs only

nebasuke:
Two main changes:
* Update cache key to include hash of the lock file, as this is both a correctness and cache saving issue (`Cargo.lock` change would not update the cache making the cache more and more out of date unless it dropped out due to not being used for 7 days). 
* Stop relying on the cache for releases. We don't want to run the risk of a caching error given the sensitivity of NAPI builds, so when we detect there is a release, we don't use the cache. See https://github.com/napi-rs/napi-rs/issues/1297. This is likely fixed in napi 3.x.

# Claude summary

## Goal

The npm-release `build` job is also used for dry-runs on every PR and non-release push, where speed matters. Real release commits skip the cache so published artifacts are always built fresh.

## What changed

- **Cache enabled for non-release builds only.** Gated via `check_commit.outputs.tag` (mirrors how cooldown-check opts in to release commits). The `check_commit` job's author-association guard is removed so it runs on all events; downstream publish gates are unchanged (`event_name` / `GITHUB_REF` inside `check_commit` + the `edr-release` environment approval on publish).
- **Cache key includes `Cargo.lock`** with a `restore-keys` fallback. Without this, `actions/cache` would skip the save when the key already existed, so any deps added by later `Cargo.lock` changes would be re-fetched and re-compiled on every run, and the cached `target/` would drift further out of sync over time. The lock-hashed key rotates the cache naturally; the fallback key keeps cold-deps cost low when the lock changes.
- **Docker matrix entries now actually use the cache.** Mounts aligned to the host `/home/runner/.cargo` paths that the cache step writes to (previously they pointed at a `.cargo-cache` path that nothing populated). Dropped the unused `.cargo-cache` entry from the cache `path:` list.
- **Windows: cache `~/.cargo/registry/src/` too.** NTFS handles many small file creates ~5-10x slower than ext4/APFS, so napi-rs's pre-cargo `cargo metadata` extraction of ~500 `.crate` tarballs into `registry/src/` takes ~50s on warm runs without this. Other targets re-extract fast enough that caching the bloat isn't worth the size.

## Trade-off

Workspace crates recompile on warm-cache runs because fresh-checkout `mtime`s look newer to cargo than the cached `target/` artifacts. Dep caching (registry, git db, and the dep-built portion of `target/`) still applies — those are keyed by source hash, not workspace mtimes — so the lock-keyed key, docker mount fix, and Windows `registry/src/` cache all remain net wins on warm runs.

The alternative (`mtime` restoration via a composite action + `touch crates/edr_napi/src/lib.rs` to force napi-derive's `index.js` / `index.d.ts` regeneration, working around napi-rs#1297) is intentionally not included — it's brittle, depends on tight coupling between the checkout and mtime-restore steps, and the dep-cache wins on their own justify shipping this.

## Impact

Build-step durations (seconds, build step only — excludes setup, cache restore, upload). PR warm averages are 3 exact-key cache-hit attempts of run [25909359281](https://github.com/NomicFoundation/edr/actions/runs/25909359281) (commit `604df1d8`). Baseline is the mean of 4 recent main-push runs on the old constant-key cache (one x86-darwin outlier at 2139s excluded).

| target       |  baseline avg |  PR warm avg (n=3) |               delta |
| ------------ | ------------: | -----------------: | ------------------: |
| x86-windows  |         1131s |               609s |  **−522s (−46.2%)** |
| x86-musl     |         1180s |               692s |  **−488s (−41.4%)** |
| x86-darwin   |         1216s |               809s |  **−407s (−33.5%)** |
| aarch-musl   |          982s |               710s |  **−272s (−27.7%)** |
| aarch-darwin |          673s |               416s |  **−257s (−38.2%)** |
| x86-gnu      |          636s |               417s |  **−219s (−34.4%)** |
| aarch-gnu    |          615s |               470s |  **−145s (−23.6%)** |

### Caveats

- **Baseline reflects a semi-cold constant-key cache, not steady-state warm.** The headline gap is partly the lock-keyed rotation eliminating the old cache's accumulated drift (deps added since the cache was first written had to be re-fetched and re-compiled every run, and that overhead has grown as `Cargo.lock` churned). What remains after rotation is the docker mount fix and the Windows `registry/src/` cache.
- **macOS / Windows runs are noisy.** Run-to-run swings across the 3 PR attempts: x86-darwin 195s, aarch-darwin 159s, x86-windows 123s, x86-musl 148s. The Linux gnu builds are very stable (≤25s spread). All deltas comfortably exceed their per-target noise band.
- **Total job time moves less than the build step.** On Windows, the larger cache restore (~12s → ~170s, mostly `registry/src/`) eats some of the build-step savings. The build-step delta is the cleaner signal of what the cache buys; total job time still improves meaningfully.


.github/workflows/edr-npm-release.yml
```diff
@@ -74,10 +81,14 @@ jobs:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5
         with:
           persist-credentials: false
+          fetch-depth: 0 # required by ./.github/actions/restore-mtime
```
anaPerezGhiglia: perhaps instead of defining `restore-mtime` composite action that only restores the files timestamp,  we can encapsulate the checkout + restore times together (`checkout-preserve-time` action or something like that). That way we don't have this fragile coupling between this two steps.

.github/workflows/edr-npm-release.yml
```diff
@@ -128,6 +145,15 @@ jobs:
...
         run: pnpm install --frozen-lockfile --prefer-offline
 
+      # napi-derive emits index.js / index.d.ts as a side effect of
+      # proc-macro expansion, which cargo skips when the crate looks
+      # unchanged (napi-rs#1297). Combined with restore-mtime + a warm
+      # target/, that would silently produce stale bindings. Touching
+      # src/lib.rs invalidates edr_napi only and forces codegen to run.
+      - name: Force edr_napi rebuild so napi-derive regenerates bindings
+        shell: bash
+        run: touch src/lib.rs
```
anaPerezGhiglia: I see why this is needed, but I find it quite disturbing to be honest

nebasuke: Yeah, it's one of the reasons I've disabled cache for the actual release flow. I wouldn't mind parking this PR, and first attempting to update NAPI to v3 as that will likely fix this.

anaPerezGhiglia: This makes me think if it is ok to cache the whole target directory. Do we want to have EDR crates cached? 🤔 

For this particular case, I think that a clearer approach would be to `cargo clean --release -p edr_napi` instead of touching the file

If we want to make sure to clean all workspace crates before building we can do it by running

```sh
cargo clean --release $(cargo metadata --format-version 1 --no-deps | jq -r '.packages[].name | split(" ")[0]' | xargs -I {} echo -n " -p {}")
```

anaPerezGhiglia: IIUC, NAPI not running if target is up to date and no file is modified is a consequence of `cargo` build behavior - not of napi, so I wouldn't expect this to change after migrating to NAPIv3. 
The `edr_napi` crate triggers the napi build through the `build.rs` script, which is re-run as any cargo build script.

nebasuke: As discussed, removed mtimes now.

.github/workflows/edr-npm-release.yml
nebasuke: Removed mtimes and fetch-depth.

anaPerezGhiglia (approved):
LGTM 👍

PR #1380, opened 2026-05-04
ci: add explicit restore-keys for EDR RPC cache

nebasuke:
Add an explicit `restore-keys: edr-(rs|ts)-rpc-cache-v1-<os>-` fallback to EDR RPC cache. 

The save step writes keys with a hashFiles suffix (`edr-rs-rpc-cache-v1-<os>-<hash>`) but the restore step's `key:` was prefix only (`edr-rs-rpc-cache-v1-<os>`). Without an explicit `restore-keys`, the prefix-match was relying on actions/cache@v4 undocumented behaviour.


popescuoctavian (approved):
LGTM, thanks!

PR #1382, opened 2026-05-04
ci: share Rust cache for same-profile dev/check jobs

nebasuke:
## Summary

Carve-out from #1374. Adds Swatinem `cache-shared-key` plumbing to the local `setup-rust` composite, then uses it to merge the two clusters of jobs whose Rust compile profiles are actually identical. Earlier rev of this PR also merged `*-coverage` jobs by OS, but attempt-2 timings showed a fingerprint-mismatch regression there (see notes); that merger has been dropped.

## Buckets

| Key | Jobs |
|---|---|
| `linux-dev-strict` | `check-edr`, `cargo-hack-edr` (workspace `cargo check --locked` under `RUSTFLAGS=-Dwarnings`) |
| `linux-dev` | `edr-style`, `edr-docs`, `cooldown-check` (plain dev workspace, no coverage instrumentation) |

The strict bucket is separate from `linux-dev` because `RUSTFLAGS=-Dwarnings` hashes into Cargo's fingerprint; collapsing them would invalidate either side on every run. A `[workspace.lints]` deny-warnings refactor would let them merge.

`test-edr-rs`, `test-edr-ts`, `edr-integration-tests`, and `run-hardhat-tests` keep their per-job buckets. They share the same OS and key inputs but diverge on crate-type and feature resolution (`cargo llvm-cov --workspace` for `test-edr-rs` vs. `edr_napi` cdylib for the napi/hardhat path), which collides on the Swatinem key but mismatches cargo's per-unit fingerprint — the warm-cache penalty was +2 min/job on macOS in PR-internal A/B.

## Other changes

- `pnpm-safety` job: `cache: false` on `setup-rust` — no Rust compile happens there, so the cache step is dead weight.

## Test plan

- [ ] `linux-dev-strict` and `linux-dev` keys appear in `gh api repos/NomicFoundation/edr/actions/caches` after first run on the branch.
- [ ] Second run hits the consolidated buckets — confirm via Swatinem's "Cache hit" log line on the affected jobs and watch for absence of full `proc-macro2` recompile right after.
- [ ] `check-edr` / `cargo-hack-edr` / `edr-style` / `edr-docs` / `cooldown-check` wall-clocks match or beat current baseline.

### DIFF
--- a/.github/actions/setup-rust/action.yml
+++ b/.github/actions/setup-rust/action.yml
@@ -10,17 +10,34 @@ inputs:
   cache:
     description: Automatically configure Rust cache
     default: true
+  cache-shared-key:
+    description: |
+      Stable cache key shared across multiple jobs. Use to consolidate
+      similar jobs (same OS, same build profile, same RUSTFLAGS) into a
+      single Swatinem cache scope, avoiding per-job cache duplication.
+    required: false
+  cache-save-if:
+    description: |
+      When the cache should be saved. Default is to save on every run
+      (including PRs) so we can iterate on cache configuration. Once
+      stable, tighten to `${{ github.ref == 'refs/heads/main' }}` to
+      stop PR-scoped saves from polluting the cache budget and from
+      being used as a poisoning surface by external contributors.
+    required: false
+    default: "true"
 
 runs:
   using: composite
   steps:
     - name: Install Rust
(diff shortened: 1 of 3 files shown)


nebasuke:
Marginal wins, various regressions. Not really worth figuring this out given that we are using ~25GB of the cache and have loads of lee-way. Can revisit if we need to build much more and need the space.

PR #1384, opened 2026-05-05
misc: remove unused `solidity-tests` export

popescuoctavian:
This PR removes the `solidity-tests` export from `edr_napi` and related files/tests, since they are not used downstream.


anaPerezGhiglia (approved):
Thanks for taking care of this!

PR #1385, opened 2026-05-06
build(edr_napi): migrate to napi-rs v3

nebasuke:
PR to migrate to napi-rs v3 with accompanying Hardhat 3 test PR: https://github.com/NomicFoundation/hardhat/pull/8258

I also tested Hardhat 2 locally with Verdaccio and this works. I've left out the PR as it needs a good amount of fixes not relating to this PR to be able to match the newest EDR version. 

I've left one migration behind the `compat-mode`, as that might need some design. 

I've tried to keep APIs the same where possible, relying on a new napi-cli release to preserve const string enum behaviour. 

~~Note that the ThreadsafeFunction teardown bug gets worse in Node 20, and a little bit in Node 22. I would suggest merging this AFTER migrating away from Node 20 (https://github.com/NomicFoundation/edr/pull/1393), and to consider a reduced Node 22 test matrix or a test suite pass on failed exit given that we have now filed it as a known Hardhat 3 issue: https://github.com/NomicFoundation/hardhat/issues/8322.~~

**Update (2026-07-02):** the intermittent bindings crashes are now diagnosed and **fixed in this PR**. The known Node TSFN teardown bug was only the first of **three stacked memory-safety bugs** — the other two (a napi-rs custom-GC use-after-free, and JS-derived `napi::Error`s dropped off the JS thread) persisted on every Node version. See "The flaky bindings crash: three stacked bugs" below. A first-chance-gdb stress harness on native arm64 went from ~7% crashes per suite run to **0/200 (musl) and 0/300 (gnu)** with the two fix commits. 

Will follow-up if/when https://github.com/napi-rs/napi-rs/pull/3370/ is in. See https://github.com/napi-rs/napi-rs/issues/3368. 

# Claude summary 

## Migrate `edr_napi` to napi-rs v3

Migrates `edr_napi` and `edr_napi_core` from napi-rs v2 to v3 (`napi`/`napi-derive` 3.x, `@napi-rs/cli` ^3.7.0), moving the bindings onto napi-rs's current major version line and its lifetime-typed API. The JS API surface is preserved except for two documented changes (see "Consumer-visible changes"); both Hardhat 2 and Hardhat 3 work without modification, apart from a 14-line test-fixture adjustment in Hardhat 3 ([hardhat#8258](https://github.com/NomicFoundation/hardhat/pull/8258)).

The first commit is the bulk mechanical migration; every later commit is a single focused concern (rebase adaptations, error propagation, tests, style). Reviewing commit-by-commit should work well.

## Consumer-visible changes

The full list — everything else is byte-compatible at the API level:

1. **`TestResult.reason` / `.counterexample` / `.valueSnapshotGroups`** are now class getters typed `T | undefined` (previously optional class fields `readonly reason?: T`, runtime contract unchanged: absent → `undefined`). v3 emits *every* `TestResult` member as a getter, but only these three — formerly optional — change structural required-ness: TypeScript treats getters as required, so object literals satisfying `TestResult` must spell these keys out. The only consumer impact found (Hardhat 3 test fixtures).
2. **`SuiteResult`** is now a plain object (interface) instead of a class — same field shapes, read identically. It is no longer a runtime *value*, though: `module.exports.SuiteResult` is dropped, so `new SuiteResult` / `instanceof SuiteResult` stop working (Hardhat only reads fields).
3. **JS exceptions thrown by `decodeConsoleLogInputsCallback` / `printLineCallback`** now surface as JSON-RPC internal-error responses on the triggering request, carrying the JS error message. Previously they were swallowed (print) or crashed the process (decode). This is the one intentional runtime behavior change in Rust code; covered by new regression tests.

Not changed, by design: all 15 napi enums keep their runtime-enum shape (`--runtime-string-enum`, new in `@napi-rs/cli` 3.7.0 via [napi-rs/napi-rs#3284](https://github.com/napi-rs/napi-rs/pull/3284), restores it for the 3 string enums that v3 would otherwise demote to type-only unions); `MineOrdering.Fifo`-style value access keeps working everywhere.

### Typed surface (bulk of the diff, mechanical)
- `JsFunction` → `Function<'env, Args, Ret>`; the `'env` lifetime propagates through 7 `#[napi(object)]` config structs and 4 napi-exposed methods (which now take `env: &'env Env` so the returned `Object<'env>` can borrow from it).
- `JsObject` → `Object<'env>` for deferred-promise returns; `JsUnknown` → `Unknown<'_>`; `Vec<JsString>` → `Vec<JsString<'env>>` (the no-`Debug`/`Display`/`Serialize` private-key-leak guard is unchanged).
- v3 removed `Clone` from JS-typed wrappers: `#[derive(Clone)]` dropped from ~25 `#[napi(object)]` data carriers (verified none are cloned); manual deep-copy `Clone` impls on `FuzzCase`/`BaseCounterExample` via `Uint8Array::with_data_copied`.
- v3 removed `&str: FromNapiValue`: 4 `pub kind: &'static str` tag fields → `String` (round-trip pinned by a mocha test).
- `#[napi(async_runtime)]` on the 5 sync entry points that call `tokio::runtime::Handle::current()`, including the `EdrContext` constructor — v3 no longer enters the tokio runtime context implicitly for sync `#[napi]` functions.

### ThreadsafeFunction call sites (6 + 1)
All rewritten to the v3 builder (`build_threadsafe_function().weak::<true>().build_callback(...)`). Weak/strong semantics are preserved exactly: the 6 callbacks that v2 `unref(env)`'d are weak; the test-suite-completion TSFN stays strong (keeps the process alive during a Solidity test run). `ErrorStrategy::Fatal` maps to the v3 builder default (`CalleeHandled = false`). v3's `call_with_return_value` now delivers the JS outcome as a `napi::Result` — the logger sites forward it through their result channels (consumer-visible change 3 above); the coverage/gas-report sites already did.

### Build & packaging
- `napi`/`napi-derive` = 3 (lockfile: 3.10.0/3.5.7; 3.10.0 is a same-day release but carries the [napi-rs#3357](https://github.com/napi-rs/napi-rs/issues/3357) memory-safety fix, so it's cooldown-allowlisted — see the bindings-crash section); `napi-build` stays 2.3 (no v3 exists).
- **`dyn-symbols` re-enabled** (a napi default that our `default-features = false` silently dropped): N-API symbols resolve at runtime via libloading instead of link-time externs. Required for `cargo test` binaries to link now that `edr_napi_core` has unit tests; also how every default-configured napi-rs package (and all MSVC builds) already works. Failure-mode nuance: a host missing a symbol fails at first call (stderr message + error status) instead of at `dlopen` — unreachable in practice for `napi8` on Node ≥ 20.
- `build_edr_napi.sh`: `--no-const-enum` kept (const enums can't be value-imported under `isolatedModules`), `--runtime-string-enum` added, `-- --locked` separator replaces `--cargo-flags`.
- `prepublish.sh`: v3 flag names, plus explicit `--skip-optional-publish --no-gh-release` so publish/release safety doesn't depend on call-site ordering or an absent `GITHUB_TOKEN`.
- package.json: `napi.name` → `binaryName`, `triples` → `targets`, `napi universal` → `universalize`.

### CI
- `cargo llvm-cov` jobs pin `CARGO_BUILD_TARGET=$(rustc --print host-tuple)` (3 jobs, identical comment): v3's `napi build` implicitly passes `--target=<host>`, redirecting cdylib output to `target/<triple>/`; without the pin, profraws and the `.node` land in different subtrees and napi-crate coverage reports 0%.

## Design decisions to weigh in on

1. **`ArrayBuffer` annotation with `Uint8Array` runtime** for the two Hardhat-2-facing callbacks (`setCallOverrideCallback`, `decodeConsoleLogInputsCallback`): the `.d.ts` matches Hardhat 2's typings; producing a real `ArrayBuffer` is impossible (lifetime-carrying type vs `'static` TSFN args). `Buffer.from(x)` accepts both. Documented at both sites; exercised end-to-end by mocha tests asserting on the actual runtime values.
2. **`Either<T, ()>` instead of `Option<T>`** on the three `TestResult` getters: napi-rs serializes `Option::None` as `null`, but consumers test `=== undefined`; `()` serializes as `undefined`.
3. **`SuiteResult` POJO / `TestResult` class:** `SuiteResult` has no methods, and Hardhat reads three of its four fields multiple times per suite (`testResults` ×3, `id` ×3, `warnings` ×2; `durationNs` is currently unread) — as a class, each `testResults` read would re-clone the entire `Vec<TestResult>`. The eager conversion is shallow: `testResults` becomes an array of `TestResult` class handles, so per-test state stays lazy. `TestResult` keeps methods (`stackTrace()`, `callTraces()`) over un-serializable Rust state, so it stays a class with per-getter clones (same cost profile as v2's field codegen).
4. **`compat-mode` feature enabled:** v3's typed `Object<'_>` doesn't implement `ToNapiValue` by value, making it unusable as TSFN args (`CallJsBackArgs: 'static`); the deprecated `JsObject` is the remaining sanctioned way to build the subscription-event object inside a TSFN callback. Scoped to one module (`#![allow(deprecated)]` with rationale). Follow-up planned to remove it; first candidate is v3's `ObjectRef` (a `'static` reference wrapper that implements `ToNapiValue` by value).
5. **`Vec<JsString<'env>>` kept** for `owned_accounts` rather than a `String` newtype: `JsString` isn't deprecated, the leak-prevention property is intact, and a newtype costs ~20 lines of manual trait impls plus an allocation per element for no gain.

## Suggested review focus

- `crates/edr_napi/src/logger.rs` + `crates/edr_napi_core/src/logger.rs` — the error-propagation change (typed `LoggerError` variants, fallible `DecodeConsoleLogInputsFn`). The only deliberate runtime behavior change.
- `crates/edr_napi/src/solidity_tests/test_results.rs` — densest file: getter pattern, manual `Clone` deep-copies, `kind: String` tags.
- `crates/edr_napi_core/src/subscription.rs` + `crates/edr_napi/src/subscription.rs` — `compat-mode`/`JsObject` usage and the `Arc`-wrapped TSFN (v3 `ThreadsafeFunction` isn't `Clone`).
- The committed `index.d.ts` diff — fastest way to audit the consumer surface: `export enum` → `export declare enum` (cosmetic), the `TestResult`/`SuiteResult` changes above, new `SolidityStackTraceEntry`/`SolidityStackTrace` aliases (fixes a dangling type reference in the generated typings), no drops or renames.

## The flaky bindings crash: three stacked bugs (all fixed)

The intermittent bindings-test crashes on this PR (SIGSEGV / SIGTRAP / `Check failed: node->IsInUse()`, arm64-clustered, scattered across Node versions) were **three independent memory-safety bugs** sharing one symptom pool. Diagnosed with a first-chance-gdb stress harness on native arm64 runners — post-mortem cores were useless, because `signal-exit` re-raises fatal signals via `process.kill` and destroys the faulting stack.

1. **Node: `napi_threadsafe_function` teardown data race** ([nodejs/node#55706](https://github.com/nodejs/node/issues/55706)) — fixed upstream by Mika Fischer (landed [`bff6ea49`](https://github.com/nodejs/node/commit/bff6ea49b860a14817097e66fb6edd08e99434cb), Node 25+, backported to 24.13.1; never to the 20/22 LTS lines). Covered by the Node 24/26 matrix bump (#1507); user-facing tracker [hardhat#8322](https://github.com/NomicFoundation/hardhat/issues/8322). This was the originally-suspected (and only known) cause; it masked the two below, which kept crashing on Node 24/26.
2. **napi-rs ≤ 3.9.4: custom-GC use-after-free for buffers dropped off the JS thread** ([napi-rs#3357](https://github.com/napi-rs/napi-rs/issues/3357)) — napi-rs recycles the `napi_ref` behind every `Buffer`/`TypedArray` dropped off the JS thread through an internal TSFN with a racy destroyed-flag check, and deletes refs the env already invalidated at teardown → V8 GlobalHandles corruption. EDR triggers this constantly (`Uint8Array`s dropped on tokio/deallocator threads). Fixed upstream in **napi 3.10.0**; adopted here via lockfile bump + a cooldown-allowlist entry (same-day release, but it *is* the memory-safety fix).
3. **JS-derived `napi::Error` dropped off the JS thread** — a `napi::Error` from a JS throw (TSFN callback) or promise rejection owns a `napi_ref` to the JS error object, is (unsafely) `Send`, and its `Drop` deletes that ref on whatever thread runs it. This PR's error bridges shipped such errors across channels and awaited rejections on the tokio runtime → the same GlobalHandles corruption; this was the **dominant** cause. v2 was immune only because throwing callbacks were swallowed or crashed the process, so the errors never crossed threads. Fixed in this PR: errors cross threads as `String`s only — stringified in the TSFN result closures (JS thread), and via `napi_error::reason_and_forget` (extract message + `mem::forget`, a bounded leak of one JS error object per failed callback) where rejections materialize on tokio threads. An upstream napi-rs issue will follow; the fix shape is routing `Error`-ref deletion through their new per-env custom-GC handle, like buffers.

Evidence (bindings suite looped on native arm64, per configuration):

| Configuration | musl (200×) | gnu (300×) |
|---|---|---|
| napi 3.9.4 (bugs 2 + 3 live) | 19 crashes | 15 crashes |
| + napi 3.10.0 (bug 2 fixed) | 8 | 14 |
| + `napi::Error` fix (bug 3 fixed) | **0** | **0** |

One bad `napi_delete_reference` corrupts V8's GlobalHandles free-list, and the *next* handle operation crashes — wherever it happens to be. That's why crash sites wandered (mid-run inside `createProvider`, at-exit in the TSFN close sweep) and why each single-cause theory only lowered the rate instead of eliminating it.

## Verification

- Hardhat 3 end-to-end: a release-profile build of this branch published via local Verdaccio into Hardhat 3 — build clean with zero TS errors and no patches, 272 tests passing across `network-manager/edr` (incl. JSON-RPC provider e2e), `gas-analytics`, and `solidity-test` (incl. real EVM execution). Hardhat-side diff: [hardhat#8258](https://github.com/NomicFoundation/hardhat/pull/8258) (test fixtures only).
- Bindings crash: root-caused and fixed — see "The flaky bindings crash: three stacked bugs" above. Validated by a first-chance-gdb stress harness on native arm64 (musl + gnu): 0 crashes in 500 combined suite iterations with both fix commits, versus ~7% per iteration before.

## Follow-ups (out of scope)

- File the upstream napi-rs issue for `napi::Error`'s unsound `Send` impl (env-bound `napi_ref` deleted in `Drop` on whatever thread runs it — see bug 3 above); once fixed upstream, `napi_error::reason_and_forget` can be retired.
- Remove `compat-mode` once an alternative TSFN-args idiom is chosen (see design decision 4); `ObjectRef` is the first candidate to spike.
- Evaluate v3's experimental `napi build --cross-compile` to replace the Docker-based Linux release legs.
- `#[napi(module_exports)]` instead of `napi-build::setup()` (cosmetic).


anaPerezGhiglia:
@nebasuke I didn't look into any of the changes so far, but in case it helps, here is the [design doc](https://www.notion.so/nomicfoundation/EDR-migrate-to-NAPIv3-Summary-31f578cdeaf580c8b6a0cd3244920e50) related to this

nebasuke:
> @nebasuke I didn't look into any of the changes so far, but in case it helps, here is the [design doc](https://www.notion.so/nomicfoundation/EDR-migrate-to-NAPIv3-Summary-31f578cdeaf580c8b6a0cd3244920e50) related to this

Yeah I used that as the starting point :-) @anaPerezGhiglia

pnpm-lock.yaml
```diff
@@ -733,15 +741,62 @@ packages:
...
     engines: {node: '>=12.22'}
 
   '@humanwhocodes/object-schema@2.0.3':
     resolution: {integrity: sha512-93zYdMES/c1D69yZiKDBj0V24vqNzB/koF26KPaagAfd3P/4gUlh3Dys5ogAK+Exi9QyzlD8x/08Zt7wIKcDcA==}
-    deprecated: Use @eslint/object-schema instead
+
+  '@inquirer/ansi@2.0.7':
+    resolution: {integrity: sha512-3eTuUO1vH2cZm2ZKHeQxnOqlTi9EfZDGgIe3BL3I4u+rJHocr9Fz86M4fjYABPvFnQG/gGK551HqDiIcETwU6Q==}
+    engines: {node: '>=23.5.0 || ^22.13.0 || ^20.17.0'}
+
```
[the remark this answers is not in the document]
nebasuke: Fixed in 7bd3172cb — raised the root `package.json` Node floor to `>=20.17.0` to match the `@inquirer/*` transitive requirement (`^20.17.0` on the 20.x line) now that `@napi-rs/cli` v3 pulls them in. Scoped to the root only: `@napi-rs/cli` is a dev dependency, so `crates/edr_napi/package.json` stays at `>= 20` to avoid imposing the floor on consumers of the published binding. Thanks!

.github/actions/setup-llvm-cov/action.yml
```diff
@@ -0,0 +1,18 @@
...
+runs:
+  using: composite
+  steps:
+    - name: Install cargo-llvm-cov
+      uses: taiki-e/install-action@7a79fe8c3a13344501c80d99cae481c1c9085912 # v2.81.10
+      with:
+        tool: cargo-llvm-cov@0.8.5
+
+    - name: Pin CARGO_BUILD_TARGET to the host triple
+      if: inputs.builds-napi-crate == 'true'
```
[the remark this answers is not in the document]
nebasuke: False positive

crates/edr_napi/Cargo.toml
```diff
@@ -45,14 +45,18 @@ edr_utils_sync.workspace = true
...
 ] }
-# The `async` feature ensures that a tokio runtime is available
-napi = { version = "2.16.17", default-features = false, features = [
+# The `async` feature ensures that a tokio runtime is available.
+# `compat-mode` is a temporary napi v2->v3 shim for passing objects through threadsafe functions.
+# `dyn-symbols` lets test binaries that keep napi code alive link outside a Node host.
+napi = { version = "3", default-features = false, features = [
     "async",
+    "compat-mode",
+    "dyn-symbols",
```
Wodann, line 53: Question: Based on
> `dyn-symbols` lets test binaries that keep napi code alive link outside a Node host.
is this test-only code?

If so, I'd recommend adding the feature flag the `test-mock` (or similar) feature flag to avoid using it for the published build.

.changeset/silver-baths-fold.md
```diff
@@ -0,0 +1,8 @@
+---
+"@nomicfoundation/edr": minor
+---
+
+- Migrated `edr_napi` to napi-rs v3.
```
Wodann: Suggestion: I think the below changes already cover the user-facing API changes. I think the migration itself is an internal detail.
```suggestion
```

.changeset/silver-baths-fold.md
```diff
@@ -0,0 +1,8 @@
...
+- The `reason`, `counterexample`, and `valueSnapshotGroups` fields on `TestResult` are now class getters returning `T | undefined`.
+- `SuiteResult` is now a plain object instead of a class; field shapes are unchanged.
```
Wodann: Style: conform to changelog format
```suggestion
- Changed the `reason`, `counterexample`, and `valueSnapshotGroups` fields on `TestResult` to class getters returning `T | undefined`.
- Changed `SuiteResult` from a class to a plain object; field shapes are unchanged.
```

.changeset/silver-baths-fold.md
```diff
@@ -0,0 +1,8 @@
...
+- Exceptions thrown by the `decodeConsoleLogInputsCallback` and `printLineCallback` logger callbacks now surface as JSON-RPC internal-error responses carrying the JS error message, instead of being swallowed or crashing the process.
```
Wodann:
```suggestion
- Fixed exceptions thrown by the `decodeConsoleLogInputsCallback` and `printLineCallback` logger callbacks from being swallowed or crashing the process. They now surface as JSON-RPC internal-error responses carrying the JS error message.
```

crates/edr_napi/src/solidity_tests/test_results.rs
```diff
@@ -48,21 +48,30 @@ pub struct ValueSnapshotEntry {
...
+/// `.testResults` multiple times per suite. As a class, each read would
+/// re-clone the entire `Vec<TestResult>`; as an object, the data is serialized
+/// to a plain JS object once when the suite-completion TSFN fires.
+///
+/// `object_from_js = false`: napi-derive's default for `#[napi(object)]` emits
+/// both `ToNapiValue` and `FromNapiValue`. We never receive a `SuiteResult`
+/// from JS (it's only ever constructed Rust-side and pushed through the
+/// progress TSFN), and the `Vec<TestResult>` field can't satisfy
+/// `FromNapiValue` because `TestResult` is a `#[napi]` class and not a POJO.
+/// Disabling the from-JS direction lets the to-JS direction compile cleanly.
```
Wodann, line 63: praise: Nice! I didn't know about this option. I've added https://github.com/NomicFoundation/edr/issues/1531 to apply this pattern across our N-API surface.

crates/edr_napi/src/solidity_tests/test_results.rs
```diff
@@ -480,6 +548,28 @@ pub struct BaseCounterExample {
     pub args: Option<String>,
 }
 
+// Manual impl; auto-derive doesn't work because `Uint8Array` does not
+// implement `Clone`, so the `sender`, `address`, and `calldata` buffers
+// are deep-copied instead.
+impl Clone for BaseCounterExample {
```
Wodann, line 554: thought: ideally we'd be able to avoid all of these `Clone` implementations, as cloning the `Uint8Array` is expensive, especially when the JS side can in theory reuse the same object by reference.

I had a quick look, but it's not trivial. We'd likely have to do a breaking change in the API to split the `TestResult` into an object for all plain-old result data (that doesn't need to be copied) and a class for the lazily copied traces. Potentially these can be combined in a new class `TestResultWithTraces`.

No necessity to do that in this PR though. I've added a TODO for myself to explore when I have time.

crates/edr_napi/src/config.rs
```diff
@@ -491,136 +491,145 @@ impl TryFrom<MiningConfig> for edr_provider::config::Mining {
...
+/// blocking Rust callback the provider core requires.
+///
+/// Builds a weak threadsafe function from `callback`, then returns a closure
+/// that, on each call, converts the core value via `to_js`, invokes the JS
+/// function on its own thread, and blocks until the returned promise resolves.
+/// The three failure points — the threadsafe call not being scheduled, a
+/// synchronous exception in the callback, and a promise rejection — are all
+/// funneled into the returned `Result`. `label` names the callback in error
+/// messages.
+fn blocking_promise_callback<Core, Js, ToJs>(
```
Wodann: Style: We use the convention of a postfix `T` for all generic types to make it clear that we're dealing with a generic type vs a normal type.

In this case that would be `CoreT, JsT, ToJsT`

crates/edr_napi/src/call_override.rs
```diff
@@ -92,25 +96,58 @@ impl CallOverrideCallback {
...
+                }
                 Ok(())
             },
         );
 
-        assert_eq!(status, Status::Ok, "Call override callback failed");
+        // Distinct from the callback-failure panic below: a non-`Ok` status
+        // means the threadsafe call itself was not scheduled (e.g. `Closing`
+        // during environment teardown), not that the user's callback failed.
+        assert_eq!(status, Status::Ok, "Call override threadsafe call failed");
```
Wodann, line 141: Suggestion(non-blocking): To avoid any panics, we can mirror this pattern by returning an `Err` when this invariant happens:

https://github.com/NomicFoundation/edr/pull/1385/changes#diff-54cff23897508e1cabb049e549e3efbdda93ac17039d72efb835a32e79554bc2R569

crates/edr_napi/src/config.rs
```diff
@@ -652,8 +660,8 @@ impl ProviderConfig {
                 #[allow(deprecated)]
                 use edr_signer::DangerousSecretKeyStr;
 
-                static_assertions::assert_not_impl_all!(JsString: Debug, Display, serde::Serialize);
-                static_assertions::assert_not_impl_all!(JsStringUtf8: Debug, Display, serde::Serialize);
+                static_assertions::assert_not_impl_all!(JsString<'_>: Debug, Display, serde::Serialize);
+                static_assertions::assert_not_impl_all!(napi::JsStringUtf8<'_>: Debug, Display, serde::Serialize);
```
Wodann: Nit: `napi::JsStringUtf8` can be imported at the top of the file

crates/edr_napi/src/napi_error.rs
```diff
@@ -0,0 +1,19 @@
...
+/// cross-thread, corrupting V8's handle pool (`GlobalHandles`), which
+/// surfaces later as `Check failed: node->IsInUse()` aborts or segfaults in
+/// unrelated `napi_ref` operations. Until napi-rs routes the deletion
+/// through its custom-GC machinery, leak the error instead: one small
+/// allocation plus one JS error object kept alive per failed callback.
+#[expect(
+    clippy::mem_forget,
+    reason = "skipping `Drop` is the point: it must not run off the JS thread"
+)]
+pub(crate) fn reason_and_forget(error: napi::Error) -> String {
```
Wodann, line 15: thought: Consider marking this `unsafe` to signal that we're doing something that breaks Rust's memory system guarantees.

crates/edr_napi/src/provider.rs
```diff
@@ -118,18 +122,35 @@ impl Provider {
...
+        // `ArrayBuffer<'env>` carries a lifetime, while threadsafe-function
+        // arguments must be `'static`.
+        //
+        // Hardhat 2 reads the arguments with `Buffer.from(x)`, which accepts
+        // both shapes identically. Any new consumer should treat this
+        // argument as a typed-array view: `Buffer.from(x)` and
+        // `new Uint8Array(x)` work, while ArrayBuffer-specific operations
+        // (`new DataView(x)` with no offset, ArrayBuffer `.slice(start, end)`
+        // semantics) would silently behave like `Uint8Array`. The same caveat
+        // applies to `decodeConsoleLogInputsCallback` in `logger.rs`.
```
Wodann, line 141: Question: Is this comment implying we should modify HH2 to pass in a `Uint8Array`? If you'd like, I can make that change on this PR as a patch, to keep the benchmarks and tests working

crates/edr_napi/test/provider.ts
```diff
@@ -99,23 +100,58 @@ describe("Provider", () => {
     printLineCallback: (_message: string, _replace: boolean) => {},
   };
 
-  it("initialize local generic provider", async function () {
-    const provider = context.createProvider(
+  // Used by the callback tests below.
+  async function createL1Provider(
```
Wodann: typo: `createGenericProvider`

Wodann: Fixed in https://github.com/NomicFoundation/edr/pull/1385/changes/f4803a427d34f9d1302b473b709bcb67b6fdf8e7

crates/edr_napi/tsconfig.typings-check.json
```diff
@@ -0,0 +1,14 @@
+{
+  "extends": "@tsconfig/node22/tsconfig.json",
+  "compilerOptions": {
+    "noEmit": true,
+    "composite": false,
+    "declaration": false,
+    "declarationMap": false,
+    "sourceMap": false,
+    // The node22 base enables this, which hides dangling type refs in
```
Wodann: issue: comments are not supported in JSON

Wodann: Addressed by https://github.com/NomicFoundation/edr/pull/1385/changes/1b42dd31c07e3e8a3e425a96b8bad16b5fdb60b6

Wodann (changes requested):
Some suggestions. I'm working on resolving them

crates/edr_napi/src/config.rs
Wodann: Addressed by https://github.com/NomicFoundation/edr/pull/1385/changes/3710b99c973ca0a9ed6f5650695c15c27cce597f

crates/edr_napi/src/provider.rs
Wodann: Addressed by https://github.com/NomicFoundation/edr/pull/1385/changes/1708730c8b6bbd68083691883eaf9f4628cb20fc

crates/edr_napi/src/config.rs
Wodann: Addressed by https://github.com/NomicFoundation/edr/pull/1385/changes/d81d4174df9df7e64866def0157666c4705ad4b4

Wodann (approved):

Wodann:
I addressed the most pressing comments I left. The other ones are nice-to-haves; so feel free to merge, if you're okay with the changes.

Wodann:
> I've left one migration behind the compat-mode, as that might need some design.

I was able to remove `compat-mode`

nebasuke:
Looks good to me! I'll rebase/merge in main to make sure we're not hitting any of the new ESLints, and then I'm happy to move forward.

crates/edr_napi/src/napi_error.rs
nebasuke: I've had a look, and this makes using this function a lot harder at callpoint. 

I looked around and I don't think it violates the memory system guarantees, see https://doc.rust-lang.org/std/mem/fn.forget.html#safety.

crates/edr_napi/Cargo.toml
nebasuke: I've filed #1533 and will address this shortly after

crates/edr_napi/src/napi_error.rs
Wodann: Good to know.

Ideally we still somehow keep track of these purposeful memory leaks (they are leaking memory, if I'm not mistaken?)

In that case, it would be good to have a GitHub issue that tracks this and is referenced from the code.

PR #1388, opened 2026-05-06
ci: upload Hardhat-tests coverage to Codecov

nebasuke:
## Why

The `hardhat-tests` workflow already runs `cargo llvm-cov report` after the JS test run and produces a `codecov.json`, but it never had an upload step — the file was generated and discarded. That suite exercises the napi binding code paths end-to-end (subscriptions, provider lifecycles, Solidity test runs, stack-trace decoding), so without its upload those files look under-covered on every PR even though they are tested.

Most visible on #1385 (napi v3 migration): patch coverage 52% with `call_override.rs`, `provider.rs`, and `trace/debug.rs` showing 0% — exactly the files Hardhat-tests would cover.

## Change

Add a `codecov-action` upload step to `hardhat-tests.yml` after `cargo llvm-cov report`, matching the pattern already used by the three uploads in `edr-ci.yml` (matrix-named, SHA-pinned, `fail_ci_if_error: false`, `CODECOV_TOKEN`).

### DIFF
--- a/.github/workflows/hardhat-tests.yml
+++ b/.github/workflows/hardhat-tests.yml
@@ -75,6 +75,14 @@ jobs:
           # `--release` included to match `build:dev` compilation flags
           cargo llvm-cov report --release --locked --codecov --output-path codecov.json
 
+      - name: Upload coverage to Codecov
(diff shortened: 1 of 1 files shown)


popescuoctavian (approved):
Great find, thanks!

PR #1392, opened 2026-05-07
fix: coverage instrumentation interfering with single-call `vm.prank`

### LINKED ISSUE #1391: Coverage diagnostics interfere with `vm.prank`
In Solidity Test the `vm.prank` cheatcode (https://www.getfoundry.sh/reference/cheatcodes/prank) allows you to impersonate an address for the next call. When running with `--coverage`, an instrumentation static call is injected before that call, effectively negating the effect of `vm.prank` and causing the test to fail.

This is happening on the Nomic fork of OpenZeppelin Contracts:

```shell
git clone https://github.com/NomicFoundation/openzeppelin-contracts.git
cd openzeppelin-contracts
git checkout chore/update-to-hardhat-3-4
npm install
npx hardhat test solidity --coverage

# Which will show up failing tests like:

1) ERC4626StdTest#test_previewWithdraw((address[4],uint256[4],uint256[4],int256),uint256)
    Error: `vm.assume` rejected too many inputs (100 allowed)

2) ERC4626StdTest#test_previewRedeem((address[4],uint256[4],uint256[4],int256),uint256)
    Error: `vm.assume` rejected too many inputs (100 allowed)
```

Swapping to `vm.startPrank` and `vm.stopPrank` works around the issue in the OpenZeppelin tests.

This may affect other cheatcodes that rely on a next call mechanism, but I have not investigated.

popescuoctavian:
This PR fixes coverage instrumentation interfering with single-call `vm.prank`. Previously, the `STATICCALL` injected by the coverage instrumentation before each statement consumed the `prank`.

closes #1391

### DIFF
--- a/crates/edr_solidity_tests/tests/testdata/default/coverage/InstrumentedCoveragePrankTest.sol
+++ b/crates/edr_solidity_tests/tests/testdata/default/coverage/InstrumentedCoveragePrankTest.sol
@@ -1,0 +1,60 @@
+// Auto-generated from data/contracts/test/CoveragePrank.sol — do not edit manually.
+// Regenerate with:
+//   cargo run -p edr_tool_solidity -- --instrument-only data/contracts/test/CoveragePrank.sol
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.26;
+
+import "cheats/Vm.sol";
+
+/// @dev Verifies that coverage instrumentation does not interfere with the
+/// single-call `vm.prank` cheatcode. The instrumenter injects a `STATICCALL`
+/// to the coverage address before each statement which should not consume
+/// the single-call `prank`.
+contract CoveragePrankHelper {
+    Vm constant vm =
+        Vm(address(uint160(uint256(keccak256("hevm cheat code")))));
+
+    /// Sets a single-call prank for `pranked`, then immediately calls
+    /// `target.recordSender()`. The the recorded sender should be `pranked`.
+    function prankAndRecordSender(
+        SenderRecorder target,
+        address pranked
+    ) external returns (address) {
+__NomicFoundationCoverage.sendHit(0x5b1bae0df22ecaf575aecbdf8cbb50dee5231ee532ff064624e1929c956b344b);         vm.prank(pranked);
+__NomicFoundationCoverage.sendHit(0xc7d6f1bb9c09169a7eb25121686028c9dfabe35ee6446f8185a70e6095eb379d);         return target.recordSender();
+    }
+
+    /// Same as above but with `vm.prank(addr, origin)` so we also catch
+    /// premature restoration of `tx.origin` by the cheatcodes inspector.
+    function prankAndRecordSenderAndOrigin(
+        SenderRecorder target,
+        address pranked,
+        address newOrigin
+    ) external returns (address sender, address origin) {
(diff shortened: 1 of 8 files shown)


crates/foundry/cheatcodes/src/inspector.rs
```diff
@@ -937,6 +938,18 @@ impl<
...
+        // `COVERAGE_ADDRESS`. They are part of the test runner's bookkeeping —
+        // not the user's contract logic — and must be invisible to cheatcode
+        // state. In particular, applying or consuming a single-call `vm.prank`
+        // here would defeat the prank for the user's intended next call.
+        // `CoverageHitCollector` short-circuits these calls earlier in the
+        // inspector stack so this guard is normally unreachable; we keep it
+        // for correctness under any future inspector reordering.
+        if call.target_address == COVERAGE_ADDRESS {
+            return None;
+        }
```
anaPerezGhiglia, line 951: perhaps is worth having a `SYSTEM_ADDRESS` slice that contains both addresses so we don't have to check on both separately every time

crates/foundry/cheatcodes/src/inspector.rs
```diff
@@ -1432,8 +1445,14 @@ impl<
...
     ) {
+        // System addresses that are transparent to cheatcode state.
+        // Treating `COVERAGE_ADDRESS` as a system address prevents the prank cleanup
+        // below from running for coverage instrumentation probes — which would
+        // otherwise wipe a single-call `vm.prank` before the user's intended next call
+        // could consume it.
         let cheatcode_call = call.target_address == CHEATCODE_ADDRESS
-            || call.target_address == HARDHAT_CONSOLE_ADDRESS;
+            || call.target_address == HARDHAT_CONSOLE_ADDRESS
+            || call.target_address == COVERAGE_ADDRESS;
```
anaPerezGhiglia, line 1455: this would be the other place

anaPerezGhiglia (approved):
LGTM! Thanks for tackling this so quickly

crates/edr_solidity_tests/tests/testdata/default/coverage/InstrumentedCoveragePrankTest.sol
```diff
@@ -0,0 +1,60 @@
...
+    function recordSender() external view returns (address) {
+__NomicFoundationCoverage.sendHit(0xa6effab53389158e07b84454221a66a7a9a5ea65890d20813c29d562d1cea9ca);         return msg.sender;
+    }
+
+    function recordSenderAndOrigin() external view returns (address, address) {
+__NomicFoundationCoverage.sendHit(0x50020f1a925abb966e10d34cb078ddc6659c8289b29f53efc10771c12dbd4ae3);         return (msg.sender, tx.origin);
+    }
+}
+
+import "__NomicFoundationCoverage-1fe87c59-dedc-4831-8918-604bc223bbfa.sol";
```
Wodann, line 60: Nit: missing newline at end of file

Wodann (approved):
Thanks for the quick fix! LGTM

crates/edr_solidity_tests/tests/testdata/default/coverage/InstrumentedCoveragePrankTest.sol
popescuoctavian: This was generated automatically by `cargo run -p edr_tool_solidity -- --instrument-only`. We can modify the tool to also include a newline in the output, I'm leaving this out of this PR for now.

PR #1393, opened 2026-05-07
build: upgrade to new Node LTS version

Wodann:
@nebasuke added:
* Rebased on main, and absorbed the changes. old changes of the PR basically became, drop node 20, and get rid of the 3.20 suffix for alpine. `mocha` changes were already done in https://github.com/NomicFoundation/edr/pull/1507. 

In addition I did the following:
* Raised the TypeScript compile baseline to Node 22, including for the napi packages, as we don't want infra/builds to misalign.  
* Regenerated lockfiles (so our libraries use Node 22).
* I've left our code on `@tsconfig/node20` as changing that affects our linter. I'll do that in a follow-up PR to leave the scope contained. 

# Wodann summary
This PR upgrades the supported Node LTS versions, as Node v20 is [end-of-life](https://endoflife.date/nodejs) and a new LTS version has since been released: 24.

While upgrading docker images, I noticed that the alpine 3.20 version we are using [has severe security vulnerabilities](https://hub.docker.com/layers/library/node/22-alpine3.20/images/sha256-c845dee47e9667a9b5a61f641d79356a2e13719b555941b5e482d4b0f5045bd9). To ensure that we receive automatic upgrades of the docker image, when this happens in the future, I dropped the `3.20` postfix. `22-alpine` maps to alpine v3.23 (see [complete list of node versions](https://hub.docker.com/_/node/)).

After the transition to Node v22, a fix was needed for the Mocha test runner in the `hardhat-tests` project. This fix was repurposed from PR #1142.


nebasuke:
FYI: I discussed this with John. We should keep supporting Node 20, as we commit to supporting Node majors for 2 months after end of life.

.github/workflows/edr-npm-release.yml
```diff
@@ -205,8 +205,8 @@ jobs:
           - host: windows-2022
             target: x86_64-pc-windows-msvc
         node:
-          - "20"
```
anaPerezGhiglia, line 225: considering @nebasuke  comment, we should keep it until ends on June

anaPerezGhiglia (changes requested):
I'm blocking this after Bas comment. 
If we keep node 20 test bindings, we should remove the `bash -c "wget -qO- 'https://unpkg.com/@pnpm/self-installer'` command from docker builds

.github/workflows/edr-npm-release.yml
anaPerezGhiglia: From https://github.com/nodejs/release#end-of-life-releases, Node 20 reached EOL 2026-04-30, so we should keep its support at least until 2026-06-30

nebasuke: We might also want to consider supporting [22,24,26] as a window, as Hardhat already added explicit CI support for Node 26. (Node 26 was released early this month.)

nebasuke: It's now July :-), I've adapted the PR!

anaPerezGhiglia (approved):
LGTM

Issue #590, opened 2024-08-08
Github action linux + hardhat setup deterministically panics


n3xt-auz:
Using the following versions 

```
"devDependencies": {
    "@nomicfoundation/hardhat-network-helpers": "^1.0.11",
    "@nomicfoundation/hardhat-toolbox": "^5.0.0",
    "@nomiclabs/hardhat-ethers": "^2.2.3",
    "@openzeppelin/contracts": "^5.0.2",
    "@openzeppelin/contracts-upgradeable": "^5.0.2",
    "@openzeppelin/hardhat-upgrades": "^3.2.0",
    "@typechain/ethers-v6": "^0.5.1",
    "@typechain/hardhat": "^9.1.0",
    "ethers6": "npm:ethers@6",
    "hardhat": "^2.22.6",
    "hardhat-deploy": "^0.12.4",
    "hardhat-deploy-ethers": "^0.4.2",
    "hardhat-gas-reporter": "^1.0.8",
    "solidity-coverage": "^0.8.12",
    "tsx": "^4.16.2",
    "typechain": "^8.3.2"
  }
```

with a pretty straightforward Hardhat configuration:

```
import 'hardhat-deploy';
import 'hardhat-deploy-ethers';
import '@nomicfoundation/hardhat-toolbox';
import '@openzeppelin/hardhat-upgrades';

import { HardhatUserConfig } from 'hardhat/config';
import { accounts, namedAccounts } from './test-wallets';

const config: HardhatUserConfig = {
  defaultNetwork: 'hardhat',
  typechain: {
    discriminateTypes: true,
    outDir: './src/typechain-types',
    node16Modules: false,
    target: 'ethers-v6',
  },
  solidity: {
    compilers: [{ version: '0.8.20', settings: {} }],
  },
  namedAccounts: namedAccounts,
  networks: {
    hardhat: {
      mining: {
        auto: true,
        interval: 1000,
      },
      throwOnTransactionFailures: true,
      throwOnCallFailures: true,
      allowUnlimitedContractSize: true,
      blockGasLimit: 30000000,
      gas: 12450000,
      initialBaseFeePerGas: 30000000,
      gasPrice: 30000000,
      accounts: accounts,
    }
  },
};

export default config;
```

a simple ERC20 contract with Proxy and running some tests using the `loadFixture()` construct, I reliably get (100+ runs) the following panic:

```
# thread '<unnamed>' panicked at /build/crates/edr_provider/src/interval.rs:92:18:
[...]

PR #1395, opened 2026-05-07
feat(edr_napi): add Provider.close() for explicit ThreadsafeFunction cleanup

### LINKED ISSUE #1394: napi-rs v3: ThreadsafeFunction cleanup race during Node env teardown (atexit crash)
## Symptom

After the napi-rs v3 migration (PR #1385), the `edr_napi` mocha suite crashes during process teardown on `macos-15`:

```
46 passing (...)
SIGBUS / SIGSEGV   (run-to-run alternation)
```

The crash fires *after* the last test prints, inside Node's `atexit` cleanup — well past mocha's success line. EDR #590 reports the same crash on Linux with a stack trace pointing at napi-rs's `ThreadsafeFunction` finalizer; macOS-15 is just a more visible expression of the same race.

## Root cause

The Node bug is [nodejs/node#55706](https://github.com/nodejs/node/issues/55706): a UAF on `ThreadsafeFunction` references during env teardown. Fixed only in Node 25 ([nodejs/node#55877](https://github.com/nodejs/node/pull/55877), commit `350b0ea895`); no LTS backport.

The napi-rs v3 migration did not introduce the bug — but it dropped v2's explicit `tsfn.unref(env)` escape hatch in favor of `weak::<true>()`, removing the workaround that was masking it. The race becomes visible against any ThreadsafeFunction (TSFN) dropped during atexit, which is every TSFN the EDR provider holds (subscriptions, console-log decode, print-line, coverage, gas-report, call-override, suite-completion).

## Fix

Add an explicit async `Provider.close()` that runs the cleanup cascade — `IntervalMiner::Drop` → `Runtime::shutdown_background()` → TSFN release — on the JS event loop, *before* atexit runs. Hardhat's network manager (and EDR's mocha suite) call it in `afterEach` / shutdown.

## Background

Full investigation (Node bug timeline, why macOS-15, why EDR's tokio runtime amplifies the race, alternatives considered, and the design of `Provider.close()`):

[edr-provider-close-investigation.md (on the fix branch)](https://github.com/NomicFoundation/edr/blob/bas/edr-provider-close/edr-provider-close-investigation.md)

## Related

- EDR #590 — same bug, Linux manifestation
- PR #1385 — napi-rs v3 migration; skips the three TSFN-exercising tests on `macos-15` to keep CI green pending this fix
- (fix PR opened separately, will reference here)

nebasuke:
## Summary

Adds an async `Provider.close()` to `edr_napi` that releases the inner `Arc<dyn SyncProvider>` (and the cascading `IntervalMiner::Drop` → `Runtime::shutdown_background()` → `ThreadsafeFunction` (TSFN) release) on the JS event loop, before Node's `atexit` runs. Works around [nodejs/node#55706](https://github.com/nodejs/node/issues/55706), a TSFN UAF during env teardown that's only fixed in Node 25.

Closes #1394. Stacked on #1385.

## Why

Without explicit cleanup, every TSFN the provider holds (subscriptions, decode-console-log, print-line, coverage, gas-report, call-override, suite-completion — 7 sites) is dropped during Node's `atexit` and races the very `napi_env` they reference. The napi-rs v2 → v3 migration didn't introduce the race, but dropped v2's `tsfn.unref(env)` escape hatch (replaced by `weak::<true>()`) that had been masking it. Outcomes:

- macOS-15 mocha CI: SIGBUS / SIGSEGV after `46 passing`, deterministically.
- Linux: same race, allocator-dependent visibility — surfaces as SIGABRT in [EDR #590](https://github.com/NomicFoundation/edr/issues/590).

Full investigation (Node bug timeline, why macOS-15 expresses it deterministically, why EDR's tokio runtime amplifies the race, alternatives considered, design of `Provider.close()`):

[edr-provider-close-investigation.md](https://github.com/NomicFoundation/edr/blob/bas/edr-provider-close/edr-provider-close-investigation.md) (temporary, removed before merge)

## What changed

- **`edr_napi/src/provider.rs`** — wrapped `Arc<dyn SyncProvider>` in `Mutex<Option<...>>`. New `pub async fn close()` does `lock().take()` then `runtime.spawn_blocking(move || drop(provider))` so blocking `IntervalMiner::Drop` runs on a tokio worker (where `block_in_place` is valid). After close, `inner()` returns `Provider has been closed`. Three method bodies (`handle_request`, `set_call_override_callback`, `set_verbose_tracing`) updated to use the helper.
- **`edr_napi/test/helpers.ts`** — `useProviderCleanup()` helper: registers an `afterEach`, returns a `track(provider)` that pushes onto a per-suite list and returns the provider for chaining.
- **`edr_napi/test/{provider,coverage,gasReport,issues,op,logs}.ts`** — all `createProvider` call sites wrapped in `track(...)` (~13 sites). `logs.ts`'s prior `evm_setIntervalMining(0)` afterEach (the workaround added in #1378) replaced by `provider.close()`.
- **Reverts the [BISECT] skip on `macos-15`** added in #1385's tip commit — those three tests now run on every platform.

## Verification

- Linux: `pnpm -C crates/edr_napi test` — 43 tests passing with all three previously-bisected tests un-skipped.
- macOS-15: this PR's CI is the proof point. The Linux pass is necessary but not sufficient since the macos-15 race was the original symptom.

## Hardhat integration (follow-ups)

`provider.close()` should be wired into both, pending a published `@nomicfoundation/edr` build with this PR landed:

- HH3 PR [hardhat#8258](https://github.com/NomicFoundation/hardhat/pull/8258) — network manager teardown.
- HH2 follow-up — same shape.

## Cleanup before merge

- Drop the temporary `edr-provider-close-investigation.md` doc commit (`4618e68d3`).


Wodann:
Did the release of https://github.com/NomicFoundation/edr/pull/1436 by any chance resolve this problem?

nebasuke:
> Did the release of #1436 by any chance resolve this problem?

I think it did! I will do some more testing, but I haven't been able to reproduce the issue any more :)

nebasuke:
Did some more testing on https://github.com/NomicFoundation/edr/pull/1385, https://github.com/NomicFoundation/edr/pull/1436 seems to have fixed this issue! Closing PR.

PR #1398, opened 2026-05-11
fix(CI): write to correct path in forge-std patch

popescuoctavian:
This PR fixes [a sporadic CI failure](https://github.com/NomicFoundation/edr/actions/runs/25680744357/job/75392667537?pr=1397).

This is very likely caused by `test_writeToml` writing to the wrong `readPath` instead of `writePath`, which resulted in a read-write race during parallel execution of the read/write tests: even though the written format is consistent with the `test_readToml` expectation, there is a small windows where the file is truncated during `fs::write` which can cause the `test_readToml` to fail as it is trying to read it. This has a small probability to happen, hence the sporadic failures.


Wodann (approved):
LGTM! Thanks for the quick fix!

PR #1424, opened 2026-05-18
Bas/fix ci test flakes

nebasuke:
Two flakes:
* More fun with `UnixTimeTest` time drift. I don't think we particularly care about precision, so I've increased the allowed drift. 
* Alchemy started returning a JsonRpcError for future block requests. I've adapted the test to allow both. 

See https://github.com/NomicFoundation/edr/actions/runs/26032135043/job/76521143259?pr=1420 for an example of the two flakes.


popescuoctavian (approved):
LGTM, thanks for adding this!

PR #1425, opened 2026-05-18
edr_solidity: parse solx DWARF debug info for Solidity stack traces

marianfe:
Adds a DWARF v5 parser for solx-built artifacts so EDR can render Solidity stack traces for tests compiled with solx 0.1.4+. Routes between the solc source-map decoder and the new DWARF decoder based on the existence of the debugInfo field; existing solc artifacts decode unchanged.

Companion hardhat-solx plugin PR: https://github.com/NomicFoundation/hardhat/pull/8296

## Notes for Reviewer

Could someone please have a cursory look at the following topics:

- DWARF Parser
- Connecting to `compiler.rs`
- Test infra + test cases (should we skip the sweep tests that compare solc to solx in the CI? currently skipped) 

Recommend reviewing by commit


crates/edr_solidity/src/artifacts.rs
```diff
@@ -10,6 +10,41 @@ use indexmap::IndexMap;
...
 use serde::{Deserialize, Serialize};
 
+/// Compiler that produced a Hardhat build-info. Absent on older build-infos
+/// and the EDR in-process flow; absent is treated as `Solc`.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Serialize, strum::Display, strum::EnumString)]
+#[serde(rename_all = "camelCase")]
+#[strum(serialize_all = "camelCase")]
+pub enum CompilerType {
+    /// Reference Solidity compiler; uses `evm.{deployed,}Bytecode.sourceMap`.
+    Solc,
```
Wodann: Suggestion: add a derived `Default` implementation to be used by the serde implementation

```suggestion
#[derive(Clone, Copy, Debug, Default, PartialEq, Eq, Serialize, strum::Display, strum::EnumString)]
#[serde(rename_all = "camelCase")]
#[strum(serialize_all = "camelCase")]
pub enum CompilerType {
    /// Reference Solidity compiler; uses `evm.{deployed,}Bytecode.sourceMap`.
    #[default]
    Solc,
```

crates/edr_solidity/src/artifacts.rs
```diff
@@ -10,6 +10,41 @@ use indexmap::IndexMap;
...
+        Some(s) => match CompilerType::from_str(&s) {
+            Ok(ct) => Ok(Some(ct)),
+            Err(_) => {
+                log::warn!("Unknown build-info compilerType {s:?}; treating as \"solc\".");
+                Ok(None)
+            }
+        },
+    }
+}
+
```
Wodann: Suggestion: This code can be avoided by using `#[serde(default)]`

```suggestion
```

crates/edr_solidity/src/artifacts.rs
```diff
@@ -143,11 +179,15 @@ pub struct BuildInfoWithOutput {
     pub id: String,
     pub solc_version: String,
     pub solc_long_version: String,
+    /// Producing compiler. Absent on older builds and the Hardhat 2 flow;
+    /// `None` is treated as [`CompilerType::Solc`].
+    #[serde(default, deserialize_with = "deserialize_optional_compiler_type")]
+    pub compiler_type: Option<CompilerType>,
```
Wodann: suggestion: `serde(default)` signals: "If the value is not present when deserializing, use the Default::default()."

```suggestion
    #[serde(default)]
    pub compiler_type: CompilerType,
```

crates/edr_solidity/src/artifacts.rs
```diff
@@ -157,13 +197,20 @@ pub struct BuildInfo {
     pub id: String,
     pub solc_version: String,
     pub solc_long_version: String,
+    /// See [`BuildInfoWithOutput::compiler_type`].
+    #[serde(default, deserialize_with = "deserialize_optional_compiler_type")]
+    pub compiler_type: Option<CompilerType>,
```
Wodann: suggestion: `serde(default)` signals: "If the value is not present when deserializing, use the Default::default()."

```suggestion
    #[serde(default)]
    pub compiler_type: CompilerType,
```

crates/edr_solidity/src/artifacts.rs
```diff
@@ -157,13 +197,20 @@ pub struct BuildInfo {
...
 
-/// A `BuildInfoOutput` contains all the output of a solc run.
+/// A `BuildInfoOutput` contains all the output of a compiler run.
 #[derive(Clone, Debug, PartialEq, Eq, Deserialize, Serialize)]
 #[serde(rename_all = "camelCase")]
 pub struct BuildInfoOutput {
+    /// Mirrored from the input file (canonical source). See
+    /// [`BuildInfoWithOutput::compiler_type`].
+    #[serde(default, deserialize_with = "deserialize_optional_compiler_type")]
+    pub compiler_type: Option<CompilerType>,
```
Wodann: suggestion: `serde(default)` signals: "If the value is not present when deserializing, use the Default::default()."

```suggestion
    #[serde(default)]
    pub compiler_type: CompilerType,
```

crates/edr_solidity/src/artifacts.rs
```diff
@@ -311,6 +361,99 @@ mod tests {
     fn serde_compiler_output() {
```
Wodann: Suggestion: rename to make it clear that this test is validating solc compiler output
```suggestion
    fn serde_solc_output() {
```

Wodann (commented):
Review of https://github.com/NomicFoundation/edr/pull/1425/changes/ce78fc272fdbf03a37ef3288a764495e0c1150db

If compiler output of `solc` and `solx` are completely different, e.g. as evidenced by `solx` leaving `bytecode.source_map` and `deployed_bytecode.source_map` strings empty, I would recommend having two completely separate data types altogether and wrapping those in an enum that decides between them.

E.g.

```rs
// `#[serde(untagged]`: There is no explicit tag identifying which variant the data contains.
// Serde will try to match the data against each variant in order and the first one that
// deserializes successfully is the one returned.
#[serde(untagged)
pub enum CompilerOutput {
  Solc(SolcOutput),
  Solx(SolxOutput),
}

pub struct SolcOutput {
  // All member fields
}

pub struct SolxOutput {
  // All member fields
}
```

This allows developers to leverage the Rust type system at compile-time rather than to have to assert things on runtime. Plus, it saves memory.

crates/edr_solidity/fixtures/sources/Scenarios.t.sol
```diff
@@ -0,0 +1,430 @@
...
+  }
+}
+
+contract InvalidOpcodeTest is Test {
+  function testInvalidOpcode() public pure {
+    assembly { invalid() }
+  }
+}
+
+// ===== F.5 — vm.expectRevert cheatcode-violation traces =====
```
Wodann: Question: What do the letter-number combinations stand for? E.g. `F.5`, G.3`?

crates/edr_solidity/src/debug_info.rs
```diff
@@ -0,0 +1,5 @@
+//! Per-compiler debug-info parsers. [`crate::source_map`] (solc) and [`dwarf`]
```
Wodann: Style: Only on exception do we use `mod.rs` files. Instead our styling is to use `src/debug_info.rs`

crates/edr_solidity/src/debug_info/dwarf.rs
```diff
@@ -0,0 +1,1626 @@
...
+
+/// Decode a solx-emitted DWARF blob into the same
+/// [`Instruction`] vector that [`crate::source_map::decode_instructions`]
+/// produces for solc artifacts.
+pub fn decode_instructions(
+    bytecode: &[u8],
+    debug_info_hex: &str,
+    build_model: &Arc<BuildModel>,
+    _is_deployment: bool,
+) -> anyhow::Result<Vec<Instruction>> {
```
Wodann: For new code in the reusable crates, we use `thiserror` error enums.

In Rust you use `anyhow` for binaries (such as a CLI) as the output format is homoegenous. For libraries, you use `thiserror` (or equivalent) for typed enums, such that callers of your APIs can make assumptions about the errors and potentially recover.

I must admit that when the `edr_solidity` crate was ported, to expedite the process, it used homogenous errors. Hence, why you might see some leftover usage of `anyhow` in library code. However, if new code use `thiserror`, it can automatically be cast to an `anyhow::Error` when using `?` or `From`.

crates/edr_solidity/src/debug_info/dwarf.rs
```diff
@@ -0,0 +1,1626 @@
...
+
+    // 4. Per-file line-start caches, populated on demand.
+    let mut line_starts_by_file_id: HashMap<u32, Vec<usize>> = HashMap::new();
+
+    // 5. Walk PCs, mirroring `source_map::decode_instructions` so PUSH operands
+    //    are skipped consistently.
+    let mut instructions = Vec::new();
+    let mut pc: usize = 0;
+    while pc < bytecode.len() {
+        let Some(&raw_op) = bytecode.get(pc) else {
```
Wodann: Suggestion: This way of iterating does not leverage the Rust compiler's type system to rule out "error states". Instead, I'd recommend something like:

```rs
let mut it = bytecode.iter(); // or .into_iter()
while let Some(raw_op) = it.next() {
  if let Some(push_size) = push_size {
    // pc += push_size;
    it = it.skip(push_size);
  }
}
```

This is a recurring pattern throughout the `dwarf.rs` code

crates/edr_solidity/src/debug_info/dwarf.rs
```diff
@@ -0,0 +1,1626 @@
...
+            break;
+        };
+        let Some(opcode) = OpCode::new(raw_op) else {
+            // Invalid opcode = start of the trailing CBOR metadata region.
+            log::debug!("DWARF decoder: invalid opcode {raw_op} at pc={pc}, stopping");
+            break;
```
Wodann: Question: Is it expected that we receive invalid DWARF files?

This feels like an error state that we should return as such; i.e. `Result::Err`.

crates/edr_solidity/src/debug_info/dwarf.rs
```diff
@@ -0,0 +1,1626 @@
...
+                    // `directory_index()` into `include_directories`.
+                    let directories: Vec<String> = header
+                        .include_directories()
+                        .iter()
+                        .map(|d| {
+                            unit_ref
+                                .attr_string(d.clone())
+                                .ok()
+                                .and_then(|s| {
+                                    s.to_string_lossy().ok().map(std::borrow::Cow::into_owned)
```
Wodann, line 368: Question: Are we okay ignoring errors by doing `ok`?

crates/edr_solidity/src/debug_info/dwarf.rs
```diff
@@ -0,0 +1,1626 @@
...
+            // Recent PUSH = JUMP destination (low 8 bytes; JUMPDESTs fit in u64).
+            // SWAP/DUP between PUSH and JUMP → bail to InternalJump.
+            let mut dest: Option<u64> = None;
+            for j in (0..i).rev().take(8) {
+                let Some(prev) = instructions.get(j) else {
+                    break;
+                };
+                if matches!(
+                    prev.opcode,
+                    OpCode::SWAP1
```
Wodann: Suggestion: For legibility, it would be good to move this to a descriptive helper function to improve legibility/self-documenting.

crates/edr_solidity/src/debug_info/dwarf.rs
```diff
@@ -0,0 +1,1626 @@
...
+            };
+            if !matches!(inst.opcode, OpCode::JUMP | OpCode::JUMPI) {
+                continue;
+            }
+            // Recent PUSH = JUMP destination (low 8 bytes; JUMPDESTs fit in u64).
+            // SWAP/DUP between PUSH and JUMP → bail to InternalJump.
+            let mut dest: Option<u64> = None;
+            for j in (0..i).rev().take(8) {
+                let Some(prev) = instructions.get(j) else {
+                    break;
```
Wodann, line 761: Thought: `break`s like these feel as though code can silently break (i.e. no error, no log) while it appears to be an issue in the provided compilation output.

In EDR, we try to bubble up errors as close to the cause as possible to allow us to provide helpful error messages that the end-user can leverage to resolve the bug.

crates/edr_solidity/src/debug_info/dwarf.rs
```diff
@@ -0,0 +1,1626 @@
...
+            .deployed_bytecode;
+        let raw = hex::decode(&bc.object).unwrap();
+        let dbg = bc.debug_info.as_deref().unwrap();
+        decode_instructions(&raw, dbg, model, false).unwrap()
+    }
+
+    /// solx flattens a modifier into its enclosing function as a single
+    /// inlined-subroutine — pin one frame (vs. solc's duplicate setIfPositive).
+    #[test]
+    fn modifier_revert_has_single_set_if_positive_inline_call_site() {
```
Wodann: Thought: It's unclear to me whether these tests comprehensively cover the logic. From scanning, they feel like a random selection of scenarios that test some of the behaviour rather than a (minimal) set of tests that tests the complete functionality of the DWARF parser.

Wodann (commented):
Review of https://github.com/NomicFoundation/edr/pull/1425/changes/a9dcc622b9bd0a16711641fbf9c047553ff39c9a#r3299834213

I commented about the DWARF parser code in-line. I'm unfamiliar with the DWARF spec, so a lot of the code feels like a black box—that's hard to review—to me. Most of my comments are thus around the usage of Rust rather than the code logic itself. Overall, it gives the impression of not fully leveraging Rust's features.

I have three questions/concerns about the testing approach:

1. Can we easily run a compilation job with `solx` for the fixture's sources from Rust? If so, it would be worthwhile to generate the `fixtures/solx_compiler_*.json` files programatically.
   - For reference, doing so using `solc` is non-trivial. Hence, why we have pre-generated files submitted in the repo.
2. Currently, we're only testing equality of stack traces for Solidity tests. Is stack trace generation for providers explicitly left out of scope?
3. Orthogonal to my prior question: I'm wondering whether testing of equality of stack traces for every possible cheatcode error (and provider error) makes sense; as it means that we'd scale our number of tests per cheatcode/provider error.
    
    Rather, Solidity stack trace generation failures due to adoption of `solx` would stem from the logic in `edr_solidity` failing. To that end, it might make more sense to base test scenarios on the different types of variability that exist in `edr_solidity` itself.

    I recognise that there is no clear test matrix, but a good starting point would be:
     - The variance in `StackTraceEntry`: https://github.com/NomicFoundation/edr/blob/main/crates/edr_solidity/src/solidity_stack_trace.rs#L47
     - Proxy detection: https://github.com/NomicFoundation/edr/blob/main/crates/edr_solidity/src/proxy_detection.rs
     - Control-flow logic in the construction of a stack trace: https://github.com/NomicFoundation/edr/blob/main/crates/edr_solidity/src/nested_trace/conversion.rs#L33; i.e.
       - Detection of cheatcode calls (your tests can be reused)
       - Detection of `console.log` calls
       - CALL vs CREATE vs PRECOMPILE

crates/edr_napi/src/provider.rs
```diff
@@ -66,8 +66,11 @@ impl Provider {
                 let compiler_output = serde_json::from_value(compiler_output)
                     .map_err(|error| napi::Error::from_reason(error.to_string()))?;
 
+                // `addCompilationResult` is Hardhat 2 only — solc artifacts.
```
Wodann: Thought: AFAIK, HH3 uses the `addCompilationResult` method as well. The difference between HH2 and HH3 being that HH2 still supports the `hardhat_addCompilationResult` JSON-RPC method. HH3 no longer supports it but still internally calls the `addCompilationResult` TS method.

I could be wrong though.

crates/edr_solidity/src/build_model.rs
```diff
@@ -68,6 +103,17 @@ impl SourceFile {
         self.functions.push(contract_function);
     }
 
+    /// Returns the [`ContractFunction`] declared at the given 1-based line
```
Wodann, line 75: Thought: `solx`-specific code is being added to heterogeneous types for `solx` and `solc`. It feels as though we should keep those separate, so that only `solx`-specific use cases can call into them, guaranteed by Rust's type system.

Wodann (commented):
Review of https://github.com/NomicFoundation/edr/pull/1425/changes/1042897670178a6076a48df5f1842a3a774f4a0f

Rather than taking the approach of patching the `solx` compiler input/output onto the `solc` types, I'd propose using runtime polymorphism by means of dynamic trait objects.

Something like:

```rs
pub trait CompilerInput {
  // trait functions required by the stack trace generation
}

// #[derive(serde::Deserialize)] with custom deserialization logic
pub struct SolxCompilerInput {
}

impl CompilerInput for SolxCompilerInput {
  // trait implementations for `solx`-specific data
}
```

This leverages the Rust type system, keeps code easier to maintain, uses less memory per object, and allows decoupling of the data format and logic.

Wodann (commented):
Review for https://github.com/NomicFoundation/edr/pull/1425/changes/78cc0bf3bbf6fe92bf505aedba83153d69af4275

EDR leverages Hardhat 2's vast legacy test suite for stack traces: https://github.com/NomicFoundation/edr/tree/main/hardhat-tests/test/internal/hardhat-network/stack-traces

Ideally, we'd either port that test suite to Rust or Hardhat 3 and run it for `solx` in addition to `solc`. If the test suite also succeeds for `solx`, that would give a strong indicator of its quality.

crates/edr_solidity/src/debug_info/dwarf.rs
marianfe: thanks, added a log statement, does this work?

crates/edr_solidity/src/artifacts.rs
marianfe: thanks, fixed. kept the custom deserializer separately for unknown-variant fallback, see reply on the deserializer comment.

crates/edr_solidity/src/artifacts.rs
marianfe: thanks, fixed. kept the custom deserializer separately for unknown-variant fallback, see reply on the deserializer comment.

crates/edr_solidity/src/debug_info/dwarf.rs
marianfe: thanks, added a comment to explain the intent.

crates/edr_solidity/src/artifacts.rs
marianfe: thanks for the comment, the idea here was to be able to fallback gracefully, as 3rd party plugins can add new/unknown compilers.

crates/edr_solidity/src/artifacts.rs
marianfe: thanks, fixed

crates/edr_napi/src/provider.rs
marianfe: thanks, fixed

crates/edr_solidity/src/debug_info.rs
marianfe: thanks, fixed

crates/edr_solidity/src/debug_info/dwarf.rs
marianfe: thanks, fixed

crates/edr_solidity/fixtures/sources/Scenarios.t.sol
marianfe: removed these, they were from an internal plan mode doc. not relevant to the PR.

crates/edr_solidity/src/artifacts.rs
marianfe: thanks, fixed. kept the custom deserializer separately for unknown-variant fallback, see reply on the deserializer comment.

crates/edr_solidity/src/artifacts.rs
marianfe: thanks, fixed

crates/edr_solidity/src/debug_info/dwarf.rs
marianfe: thanks, fixed

crates/edr_solidity/src/debug_info/dwarf.rs
marianfe: thanks, fixed

crates/edr_solidity/src/build_model.rs
marianfe: thanks, fixed

crates/edr_solidity/src/debug_info/dwarf.rs
marianfe: thanks, fixed

crates/edr_solidity/src/debug_info/dwarf.rs
marianfe: thanks, improved the coverage with a more methodical organization around the variability points you raised:
- DWARF parser tests reorganized into 6 groups: `bottom_frame_resolution`, `inline_call_sites`, `jump_type_assignment`, `source_length_resolution`, `dwarf_file_matching`, `edge_cases`
- New `parse_errors` sub-module exercising the `DwarfError` variants (`HexDecode`, `ElfParse`, `RangeEscapesBytecode`)
- New `nested_trace/conversion.rs` tests covering the CALL / CREATE / PRECOMPILE / cheatcode / console.log seams
- New provider-flow integration tests covering 3 `StackTraceEntry` variants (`RevertError`, `PanicError`, `CustomError`) — limited coverage here for now but should be good enough. We can add more with a follow-up PR where we can also look at the HH2 stack-traces suite.

crates/edr_provider/tests/integration/solx_stack_trace.rs
```diff
@@ -0,0 +1,344 @@
...
+}
+
+// ---------- variance-axis tests ----------
+
+/// Counter.set(0) reverts via `require(v > 0, "must be positive")`.
+/// Pin: stack trace surfaces a [`StackTraceEntry::RevertError`] referencing
+/// Counter.sol. Covers the provider-flow plumbing end-to-end ([R3.2]) and
+/// the `RevertError` axis ([R3.3]).
+#[tokio::test(flavor = "multi_thread")]
+async fn revert_error_variant_surfaces_for_counter() -> anyhow::Result<()> {
```
marianfe, line 243: @Wodann re your question on whether stack-trace generation for providers was explicitly out of scope — it wasn't intentionally, just not yet exercised. Added this provider integration test: deploys a solx-built contract through the JSON-RPC provider, triggers a revert, asserts the rendered stack trace surfaces with the expected `StackTraceEntry::RevertError` and references the source file. Two more variant tests in this file cover `PanicError` and `CustomError`.

crates/edr_solidity/src/nested_trace/conversion.rs
```diff
@@ -210,3 +210,165 @@ fn is_calllike_op(step: &CallTraceStep) -> bool {
             | opcode::CREATE2
     )
 }
+
+/// Control-flow axis tests for [`convert_node`]. Each test isolates one
+/// dispatch branch (CALL / CREATE / PRECOMPILE / cheatcode / console.log)
+/// against a hand-built `CallTraceArena`, so a regression in any branch
+/// surfaces without needing a full provider round-trip.
+#[cfg(test)]
+mod tests {
```
marianfe: @Wodann re your suggestion to base test scenarios on the variability that exists in `edr_solidity` itself rather than per-error-class — adopted that direction. The most direct artifact is in this file: new tests for the CALL / CREATE / PRECOMPILE / cheatcode / console.log code paths you listed (previously zero unit tests). See also the DWARF parser test reorganization and `StackTraceEntry` variant tests covered in the reply on your other comment.

crates/edr_solidity/src/artifacts.rs
```diff
@@ -270,17 +311,59 @@ pub struct CompilerOutputSource {
     pub ast: serde_json::Value,
 }
 
-/// The bytecode output for a given compiled contract.
+/// Bytecode output for a compiled contract. Wraps an `Arc<dyn CompilerArtifact>`
+/// so the stack-trace pipeline dispatches dynamically over compiler-specific
+/// implementations ([`SolcBytecode`], [`SolxBytecode`]).
+#[derive(Clone, Debug)]
+pub struct CompilerOutputBytecode(pub Arc<dyn crate::debug_info::CompilerArtifact>);
```
marianfe: @Wodann re your comments in the review summary, addressed in this commit: split at the bytecode level rather than top-level CompilerOutput (which is the same for both solx and solc with no compiler-specific fields). The `CompilerArtifact` trait is implemented on `SolcBytecode` / `SolxBytecode` as `Arc<dyn CompilerArtifact>`. what do you think, does this work?

crates/edr_solidity/src/debug_info.rs
```diff
@@ -1,5 +1,185 @@
...
+    /// [`Instruction`] vector consumed by the stack-trace pipeline.
+    fn decode_instructions(
+        &self,
+        normalized_code: &[u8],
+        build_model: &Arc<BuildModel>,
+        is_deployment: bool,
+    ) -> anyhow::Result<Vec<Instruction>>;
+
+    /// Downcast hook so tests can recover the concrete type.
+    fn as_any(&self) -> &dyn Any;
```
Wodann: issue: downcasting of a trait object signals that the polymorphic API is leaking.

If it's only needed for tests, I'd recommend hiding the `fn` behind a `#[test]` feature flag. However, a rewrite that addresses my larger concern around following SOLID principles would void the need for `fn as_any`.

I was able to achieve this locally by propagating the compiler artifact type `ArtifactT` using this pattern:

```rs
/// The main output of the Solidity compiler.
#[derive(Clone, Debug, Deserialize)]
pub struct CompilerOutput<ArtifactT: CompilerArtifact> {
    // Retain the order of the sources as emitted by the compiler.
    // Our post processing relies on this order to build the codebase model.
    pub sources: IndexMap<String, CompilerOutputSource>,
    pub contracts: HashMap<String, HashMap<String, CompilerOutputContract<ArtifactT>>>,
}

#[cfg(test)
mod tests {
    #[test]
    fn serde_solx_output() {
        let compiler_output_json = include_str!("../fixtures/solx_compiler_output.json");
        let output: CompilerOutput<SolxBytecode> = serde_json::from_str(compiler_output_json)
            .expect("should deserialize solx compiler output");

        let contract = output
            .contracts
            .get("Counter.sol")
            .and_then(|m| m.get("Counter"))
            .expect("Counter.sol::Counter should be in the solx fixture");

        // \x7fELF magic, hex-encoded.
        assert!(contract.evm.bytecode.debug_info.starts_with("7f454c46"));
        assert!(
            contract.evm.bytecode.debug_info.len() >= 200
                && contract.evm.deployed_bytecode.debug_info.len() >= 200
        );
        assert!(contract
            .evm
            .deployed_bytecode
            .debug_info
            .starts_with("7f454c46"));
    }
}
```

crates/edr_provider/tests/integration/solx_stack_trace.rs
```diff
@@ -0,0 +1,352 @@
...
+        .get(file)
+        .and_then(|m| m.get(contract))
+        .with_context(|| format!("fixture missing {file}::{contract}"))?
+        .evm;
+    let artifact = evm.bytecode.as_artifact();
+    let object = artifact
+        .as_any()
+        .downcast_ref::<SolxBytecode>()
+        .map(|b| &b.object)
+        .context("fixture must carry SolxBytecode")?;
```
Wodann: suggestion: to avoid downcasting, I recommend using `solx`-specific types for tests that validate `solx`.

I created an example of how this can be done using Rust's type system: https://github.com/NomicFoundation/edr/commit/7e619dfed0ebbe5bbb6545a613cb7c5816dbbc6a

crates/edr_solidity/src/artifacts.rs
```diff
@@ -270,17 +313,59 @@ pub struct CompilerOutputSource {
     pub ast: serde_json::Value,
 }
 
-/// The bytecode output for a given compiled contract.
+/// Bytecode output for a compiled contract. Wraps an `Arc<dyn
+/// CompilerArtifact>` so the stack-trace pipeline dispatches dynamically over
+/// compiler-specific implementations ([`SolcBytecode`], [`SolxBytecode`]).
+#[derive(Clone, Debug)]
+pub struct CompilerOutputBytecode(pub Arc<dyn crate::debug_info::CompilerArtifact>);
```
Wodann: suggestion: an `Arc` should not be required as we only ever expect there to be one owner per `dyn CompilerArtifact`.

This might have resulted from the `BuildInfoOutput` deriving the `Clone` implementation, but I think that is unnecessary.

If I'm wrong and `Clone` is used somewhere, you can require `DynClone` to be implemented on `CompilerArtifact`. Here's an [example](https://github.com/NomicFoundation/edr/blob/main/crates/edr_coverage/src/reporter.rs#L14) of us using `DynClone` elsewhere in the codebase.

Wodann: thought: I have an example showing how you can guarantee the compiler type at the type system. At that point, you don't need this wrapper type any more to determine which type of compiler the output bytecode is from.

See example: https://github.com/NomicFoundation/edr/commit/7e619dfed0ebbe5bbb6545a613cb7c5816dbbc6a

crates/edr_provider/tests/integration/solx_stack_trace.rs
```diff
@@ -0,0 +1,352 @@
...
+    ))?;
+    let tx_hash: B256 = serde_json::from_value(response.result)?;
+    let receipt_response = provider.handle_request(ProviderRequest::with_single(
+        MethodInvocation::GetTransactionReceipt(tx_hash),
+    ))?;
+    let receipt: L1RpcTransactionReceipt = serde_json::from_value(receipt_response.result)?;
+    receipt
+        .contract_address
+        .context("deployment receipt must carry contract_address")
+}
```
Wodann: nit: I think there is code duplication here with other test code (utilities): https://github.com/NomicFoundation/edr/blob/main/crates/edr_provider/src/test_utils.rs#L248

There is a `Selector` type that might be able to do some of the work of `fn selector`: https://github.com/NomicFoundation/edr/blob/main/crates/foundry/evm/traces/src/decoder/mod.rs#L417

crates/edr_provider/tests/integration/solx_stack_trace.rs
```diff
@@ -0,0 +1,352 @@
...
+
+/// Sends a transaction and expects [`ProviderError::TransactionFailed`] to
+/// be returned — i.e. the call reverted under `bail_on_transaction_failure`.
+/// Pulls the stack trace out of the failure and returns it directly to
+/// avoid naming `TransactionFailureWithCallTraces` (its module is private).
+fn expect_failed_call_stack_trace(
```
Wodann, line 161: praise: nice test helper!

crates/edr_provider/tests/integration/solx_stack_trace.rs
```diff
@@ -0,0 +1,352 @@
...
+        | StackTraceEntry::UnrecognizedContractError { .. } => None,
+    }
+}
+
+// ---------- variance-axis tests ----------
+
+/// Counter.set(0) reverts via `require(v > 0, "must be positive")`.
+/// Pin: stack trace surfaces a [`StackTraceEntry::RevertError`] referencing
+/// Counter.sol. Covers the provider-flow plumbing end-to-end ([R3.2]) and
+/// the `RevertError` axis ([R3.3]).
```
Wodann: nit: seemingly lingering references to local docs (`R3.2`)

crates/edr_solidity/src/debug_info/dwarf.rs
Wodann: Looks good! Thanks for addressing

crates/edr_solidity/src/debug_info/dwarf.rs
```diff
@@ -0,0 +1,1790 @@
...
+    }
+}
+
+/// Resolved row from the DWARF `.debug_line` program.
+#[derive(Clone, Debug)]
+struct LineRow {
+    /// Combined `<dir>/<name>` from the DWARF file table.
+    file: String,
+    /// `0` is DWARF's "no line" sentinel.
+    line: u64,
```
Wodann: nit: when possible, at least in Rust, you try to avoid encoding sentinels like these and instead leverage the type system by using `Option<u64>` (using a conversion step from file format upon parsing: `0` -> `None`).

In this case you can even use `Option<NonZeroU64>` because you're guaranteed 0 will never exist.

This [usage site](https://github.com/NomicFoundation/edr/pull/1425/changes#diff-cd6595b1092dc92d5e62b8c7f90b5bb254c0fe604661531ed1a2bcb2b3cfa9c2R681) illustrates the benefit well:

```rs
        if let Some(row) = self.location_for_pc(pc)
            && row.line != 0
            && let Some((file_id, offset)) = resolve_location_by_name(
                &row.file,
                row.line,
                row.column,
                name_to_file_id,
                build_model,
                line_starts_cache,
            )
```

Currently, the type system would allow us to pass the `row.line` number `0` to the `resolve_location_by_name` whereas it would be impossible if `row.line: Option<NonZeroU64>`. Similarly, if `resolve_location_by_name` can receive a `NonZeroU64` to guarantee that it never receives the line number 0—if it knows that that line number never exists.

crates/edr_solidity/src/debug_info/dwarf.rs
```diff
@@ -0,0 +1,1790 @@
...
+
+    use super::*;
+    use crate::{
+        artifacts::{CompilerOutput, CompilerOutputBytecode, SolxBytecode},
+        build_model::SourceFile,
+    };
+
+    /// Test helper: every fixture in this module is solx-emitted, so the
+    /// dyn-trait artifact must downcast to the [`SolxBytecode`] concrete type.
+    fn as_solx(bc: &CompilerOutputBytecode) -> &SolxBytecode {
```
Wodann: suggestion: See https://github.com/NomicFoundation/edr/commit/7e619dfed0ebbe5bbb6545a613cb7c5816dbbc6a for how downcasting can be avoided

crates/edr_solidity/src/debug_info/dwarf.rs
```diff
@@ -0,0 +1,1790 @@
...
+    for (i, b) in content.bytes().enumerate() {
+        if b == b'\n' {
+            starts.push(i + 1);
+        }
+    }
+    starts
+}
+
+#[cfg(test)]
+mod tests {
```
Wodann, line 981: question: I must admit that not all of the terminology in the DWARF test scenarios or the tested behavbiour made sense to me. I'm probably lacking understanding of DWARF-details.

Is there some reference document that would be helpful to understand better?

crates/edr_solidity/src/nested_trace/conversion.rs
```diff
@@ -210,3 +210,174 @@ fn is_calllike_op(step: &CallTraceStep) -> bool {
             | opcode::CREATE2
     )
 }
+
+/// Control-flow axis tests for [`convert_node`]. Each test isolates one
+/// dispatch branch (CALL / CREATE / PRECOMPILE / cheatcode / console.log)
+/// against a hand-built `CallTraceArena`, so a regression in any branch
+/// surfaces without needing a full provider round-trip.
+#[cfg(test)]
```
Wodann: question: are these test meant to catch regressions for

> CALL vs CREATE vs PRECOMPILE

mentioned in review comment https://github.com/NomicFoundation/edr/pull/1425#pullrequestreview-4357740737?

If so, how does this test that `solx` build info works correctly for call, create, and precompile instructions?

AFAICT, this is testing existing (unmodified) code that doesn't directly interface with `solx` build info at all.

crates/edr_solidity/src/compiler.rs
```diff
@@ -17,17 +17,16 @@ use crate::{
...
-/// For the Solidity compiler version and its standard JSON input and
-/// output, creates the source model, decodes the bytecode with source
-/// mapping and links them to the source files.
-///
-/// Returns the decoded bytecodes that reference the resolved source model.
+/// For the Solidity compiler version and its standard JSON input and output,
+/// creates the source model, decodes the bytecode, and links them to the
+/// source files. The producing compiler is derived from each contract's
+/// bytecode variant (see [`CompilerOutputBytecode`]) — no external tag
+/// needs to be threaded in.
```
Wodann: issue: I believe this method still needs to be abstracted for `solx` vs `solc` (more details given at the top-level comment)

crates/edr_solidity/src/debug_info.rs
```diff
@@ -0,0 +1,188 @@
...
+    /// Reverse-index of `file_id_to_source_file` keyed by source name —
+    /// the DWARF parser uses this to resolve a DWARF file string back to
+    /// the [`BuildModel`]'s `file_id`.
+    fn name_to_file_id(&self) -> &HashMap<String, u32>;
+
+    /// Smallest (leafmost) AST `(offset, length)` span containing `offset` —
+    /// the DWARF parser uses this to widen a zero-length `(file, line)`
+    /// hit into the surrounding AST node's span for the renderer.
+    fn smallest_enclosing_span(&self, file_id: u32, offset: u32) -> Option<(u32, u32)>;
+}
```
Wodann: issue: this breaks the abstraction of the `solx` vs `solc` build info formats from the Solidity stack trace implementation, as `BuildModel` still needs to know about `solx`-specific functionality and types (as well as potential `solc`-specifics)

See top-level review comment for more information.

crates/edr_solidity/src/error_inferrer.rs
```diff
@@ -118,7 +118,9 @@ impl<HaltReasonT: HaltReasonTrait> From<alloy_dyn_abi::Error> for InferrerError<
 
 pub(crate) fn filter_redundant_frames<HaltReasonT: HaltReasonTrait>(
     stacktrace: Vec<StackTraceEntry>,
+    compiler_type: crate::artifacts::CompilerType,
```
Wodann: issue: changes in this file break the abstraction of the `solx` vs `solc`-specific build info formats from the Solidity stack trace implementation.

See top-level review comment for more information.

crates/edr_solidity/src/library_utils.rs
```diff
@@ -23,9 +23,9 @@ pub fn normalize_compiler_output_bytecode(
 }
 
 /// Retrieves the positions of the library addresses in the bytecode.
-pub fn get_library_address_positions(bytecode_output: &CompilerOutputBytecode) -> Vec<u32> {
-    bytecode_output
-        .link_references
+pub fn get_library_address_positions(artifact: &dyn CompilerArtifact) -> Vec<u32> {
+    artifact
```
Wodann, line 27: praise: clean abstraction!

js/integration-tests/solx-parity-sweep/.gitignore
```diff
@@ -0,0 +1,7 @@
+node_modules/
```
Wodann: nit: the `.gitignore` duplicates the entries `node_modules/` and `*.tsbuildinfo` that already exist at the workspace root

js/integration-tests/solx-parity-sweep/hardhat.config.ts
```diff
@@ -0,0 +1,22 @@
+import type { HardhatUserConfig } from "hardhat/config";
+import HardhatSolx from "@nomicfoundation/hardhat-solx";
+
+// `as HardhatUserConfig` tells TypeScript to accept the `type: "solx"`
+// solidity profile — the hardhat-solx plugin augments the config types
+// at runtime via module declaration merging, which the TypeScript
+// compiler doesn't always pick up across plugin boundaries.
```
Wodann: question: is this an issue because the `solx` plugin hasn't been released yet? Or is this a general problem?

Wodann (changes requested):
Thank you for addressing the feedback. The detailed error reporting looks good, as well as the added test suite for the provider. The tests suite for DWARF also looks better and—as discussed—suffices for now as a stand-in for the HH2 Solidity stack trace tests.

I still see one concern, though. Despite a trait having been implemented for `CompilerArtifact`, I think the actual problem I raised in https://github.com/NomicFoundation/edr/pull/1425#pullrequestreview-4358650299 has not been addressed.

The resulting code still does not adhere to SOLID principles. This mainly shows in two ways:
1. the `compiler_type` is needed (and stored) throughout the code rather than at the type level (`BuildInfoWithOutput`, `BuildInfo`, `BuildInfoOutput`; .e.g `BuildInfoWithOutput<SolxBytecode>`)
   - At a type-level, we know for a fact that metadata for a single build info should have been generated by the same compiler; i.e. only complete build infos can be generated by different compilers.
2. code that should be agnostic to the compiler metadata still needs to be aware of the details of the compiler implementation:
    ```rs
    if compiler_type == CompilerType::Solx { /* do A */ } else { /* do B */ }
    ```

The current polymorphism implementation only abstract part of the generated build info. The way (parts of) `create_models_and_decode_bytecodes` and the stack trace heuristics interact with that build info has not been abstracted yet.

By intertwining the two concepts, the code is harder to understand, less flexible, and harder to maintain; impacting our ability to debug & fix bugs or extend the existing stack trace functionality.

The suggestion of using runtime polymorphism should be able to abstract this too; allowing the Solidity stack trace heuristic to be agnostic to the compiler and its build info format.

I created an example that illustrates how item 1 can be done using Rust's type system: https://github.com/NomicFoundation/edr/commit/7e619dfed0ebbe5bbb6545a613cb7c5816dbbc6a
- The example uses a combination of compile-time and run-time polymorphism to tailor to the different (test) use cases.

Item 2 might require an additional API on the polymorphic trait that converts the compiler-specific build info to a heterogeneous `ContractMetadata`, which now lives in `create_models_and_decode_bytecodes` and associated methods.

crates/edr_napi/src/provider.rs
```diff
@@ -69,16 +69,16 @@ impl Provider {
...
-
                 // `addCompilationResult` is called by both HH2 (via the legacy
                 // `hardhat_addCompilationResult` JSON-RPC method) and HH3
                 // (internally, for its provider's in-process compile flow).
                 // Both feed solc artifacts; solx reaches EDR through HH3's
                 // `BuildInfoConfig` (`runSolidityTests` / `withContracts`).
-                // The compiler is derived from the bytecode variant — no
-                // separate tag needs to be threaded in.
+                let compiler_output: edr_solidity::artifacts::CompilerOutput<
+                    edr_solidity::artifacts::SolcBytecode,
```
marianfe: TODO: why is the solcbytecode type hardcoded here?

crates/edr_solidity/benches/decode_instructions.rs
```diff
@@ -0,0 +1,200 @@
+//! Head-to-head benchmark of the two parsers consumed by
```
marianfe: TODO: remove this file and associated temporary benchmark files from the commit.

crates/edr_solidity/src/debug_info/dwarf.rs
marianfe, line 760: TODO: in an earlier reply to a comment we got (line R772) , we said we added a log statement to not break silently, but i dont see that log now anywhere, we lost it.

crates/edr_solidity/src/error_inferrer.rs
```diff
@@ -327,10 +339,10 @@ pub(crate) fn infer_before_tracing_create_message<HaltReasonT: HaltReasonTrait>(
 }
 
 pub(crate) fn instruction_to_callstack_stack_trace_entry<HaltReasonT: HaltReasonTrait>(
-    contract_meta: &ContractMetadata,
+    ctx: TraceContext<'_>,
```
marianfe: TODO: name it context

crates/edr_solidity/src/error_inferrer.rs
```diff
@@ -501,15 +471,15 @@ fn check_custom_errors<HaltReasonT: HaltReasonTrait>(
...
+    if let Some(loc) = &last_instruction.location
         && let Some(failing_function) = loc.get_containing_function()?
+        && let Some(strategy) = trace.trace_strategy()
     {
-        stacktrace.extend(build_solx_inline_callstack_frames::<HaltReasonT>(
-            &contract_meta,
-            last_instruction,
-            &failing_function,
-        )?);
+        let ctx = TraceContext {
```
marianfe: TODO: name it context - same everywhere

crates/edr_solidity/src/trace_strategy.rs
```diff
@@ -0,0 +1,368 @@
...
+pub struct PanicHelperContext<'a> {
+    /// EVM-only step PCs in call order, extracted from `NestedTrace::steps()`.
+    pub step_pcs: &'a [u32],
+    /// User-facing function-start reference for the calldata-selector
+    /// function on the current Call, or `None` if the trace is a Create,
+    /// the calldata is empty, or the selector doesn't match a known function.
+    pub call_selector_function_start_ref: Option<SourceReference>,
+}
+
+/// Polymorphic stack-trace behaviour. Each site in
```
marianfe: TODO: refactor this comment to simplify it and stop mentioning 'used to branch on compiler type' - that code was never checked in, it was just a PR iteration. comment should describe what this trait is, not reference earlier iterations and no explanations of why we got here. 

also TODO: all the comments in this file need to be trimmed down / made concise / single line.

js/integration-tests/solx-parity-sweep/hardhat.config.ts
```diff
@@ -0,0 +1,22 @@
+import type { HardhatUserConfig } from "hardhat/config";
+import HardhatSolx from "@nomicfoundation/hardhat-solx";
+
+// `as HardhatUserConfig` tells TypeScript to accept the `type: "solx"`
```
marianfe: TODO: remove this entire comment from lines 4-7, it's incorrect.

crates/edr_napi/src/provider.rs
```diff
@@ -69,16 +69,16 @@ impl Provider {
                 let compiler_input = serde_json::from_value(compiler_input)
                     .map_err(|error| napi::Error::from_reason(error.to_string()))?;
 
-                let compiler_output = serde_json::from_value(compiler_output)
+                // Solc-only bridge: HH2's legacy `hardhat_addCompilationResult`
+                // JSON-RPC and HH3's in-process solc compile flow both feed
+                // solc artifacts here. Solx reaches EDR through
+                // `BuildInfoConfig` (`runSolidityTests` / `withContracts`),
+                // dispatched by `BuildInfoBuffers::parse`.
```
Wodann: Question: Does the solc plugin not send new compilation results for a node that runs in the background (`npx hardhat node`?

If a new contract is compiled and deployed to EDR after instantiation with `BuildInfoBuffers`, we'd not have any `ContractMetadata`.

crates/edr_solidity/src/build_model.rs
```diff
@@ -41,9 +41,36 @@ pub struct BuildModel {
     pub(crate) ast_spans: HashMap<u32, Vec<(u32, u32)>>,
 }
 
-// Solc-only paths reach for none of the surface declared on
-// [`crate::debug_info::SolxBuildModelExt`]; that trait holds the
-// methods that exist solely to support DWARF resolution.
+impl BuildModel {
```
Wodann: issue: this is still `solx` mixing with `solc`-specific types.

crates/edr_solidity/src/compiler.rs
```diff
@@ -94,16 +98,21 @@ pub(crate) fn create_sources_model_from_ast(
...
-        collect_ast_spans(&source.ast, &mut ast_spans);
-    }
-    for spans in ast_spans.values_mut() {
-        spans.sort_unstable();
-        spans.dedup();
-    }
+    // A build info is produced by a single compiler, so any artifact in the
+    // output can collect the AST spans its decoder needs. No contracts means
+    // no bytecode will be decoded, so no spans are needed either.
+    let ast_spans = compiler_output
```
Wodann: issue: `solx`-specific code mixing with `solc` code.

crates/edr_solidity/src/contracts_identifier.rs
```diff
@@ -12,8 +12,31 @@ use edr_primitives::{bytecode::opcode::OpCode, Address};
...
+/// per-bytecode data.
+#[derive(Clone)]
+pub struct IdentifiedContract {
+    /// Resolved metadata for the contract's bytecode.
+    pub metadata: Arc<ContractMetadata>,
+    /// Compiler-specific stack-trace strategy.
+    pub trace_strategy: &'static dyn TraceStrategy,
+}
+
+impl std::fmt::Debug for IdentifiedContract {
```
Wodann: suggestion: avoid manually deriving of `Debug`, instead favoring `#[derive(Debug)]`

crates/edr_solidity/src/debug_info.rs
```diff
@@ -101,15 +53,21 @@ pub trait CompilerArtifact: std::fmt::Debug + Send + Sync {
...
+    fn trace_strategy(&self) -> &'static dyn TraceStrategy;
+
+    /// Per-file AST `src` spans (`file_id` → sorted `(offset, length)`) this
+    /// artifact's debug-info decoder needs from the compilation output's
+    /// ASTs. The sourceMap decoder resolves spans directly and collects
+    /// none; the DWARF decoder derives `SourceLocation.length` from them.
+    fn collect_ast_spans(
+        &self,
+        sources: &IndexMap<String, CompilerOutputSource>,
+    ) -> HashMap<u32, Vec<(u32, u32)>>;
```
Wodann: issue: This is `solx`-specific functionality that's exposed through a general-purpose `trait` API

crates/edr_solidity/src/nested_trace.rs
```diff
@@ -146,6 +146,10 @@ pub struct CreateMessage<HaltReasonT> {
     pub steps: Vec<NestedTraceStep<HaltReasonT>>,
     /// Resolved metadata of the contract that is being executed.
     pub contract_meta: Option<Arc<ContractMetadata>>,
+    /// Compiler-specific stack-trace strategy for the contract that is
+    /// being executed. Set together with `contract_meta` at frame
+    /// construction time; `None` when the contract is unrecognised.
+    pub trace_strategy: Option<&'static dyn crate::trace_strategy::TraceStrategy>,
```
Wodann: issue: This `Option` is always the same as that for `contract_metadata`. As such; i.e. if one is `Some`, the other should be `Some`. If one is `None`, the other should be `None`.

This is evident from the way this is constructed; i.e. we split a single `Option` into two.

Instead of separating these on this type, they should be bundled together, like in the original `IdentifiedContract`.

crates/edr_solidity/src/trace_strategy.rs
```diff
@@ -0,0 +1,349 @@
...
+}
+
+/// Bundles `ContractMetadata` + [`TraceStrategy`] into one call-site handle.
+#[derive(Clone, Copy)]
+pub struct TraceContext<'a> {
+    /// Contract metadata for the current frame.
+    pub contract_meta: &'a ContractMetadata,
+    /// Compiler-specific stack-trace strategy for that contract.
+    pub strategy: &'a dyn TraceStrategy,
+}
```
Wodann: suggestion: this type essentially recreates `IdentifiedContract` but as two internal references. Instead you can use `&IdentifiedContract`.

crates/edr_solidity/src/debug_info/dwarf.rs
```diff
@@ -617,11 +614,11 @@ impl ParsedDwarf {
         let _ = dwarf_file_names; // unused on this path
         let line_program_func: Option<Arc<crate::build_model::ContractFunction>> = self
             .location_for_pc(pc)
-            .filter(|row| row.line != 0)
             .and_then(|row| {
+                let line = row.line?.get();
```
Wodann: suggestion: the idea of using `NonZeroU64` is that at the type level you can guarantee that the integer is never zero. This is useful for `resolve_location_by_name`, which does a `-1` which is only valid for non-zero integers.

Instead of doing `.get()` here, consider passing the `NonZeroU64` into the function.

Wodann (commented):
The adoption of the generic `ArtifactT` and `TraceStrategy` are a large step forwards. It isolates a lot of the storage and control-logic flow variance between `solx` and `solc`. However, as illustrated by my comments, there is still `solx`-specific data and logic during the construction of the `ContractMetadata`.

My intuition is that the `BuildModel` should be the part of a factory pattern. I'm trying to work out that API and clean up its usage locally, but I've submitted this review with comments for other, smaller issues in the interim. I already addressed and committed changes for those.

crates/edr_napi/src/provider.rs
Wodann: Support for `solx` is added by https://github.com/NomicFoundation/edr/pull/1425/changes/28d02d7ab51d74de2b5c6ee318312e2fa7cf7bce

crates/edr_solidity/src/contracts_identifier.rs
Wodann: Addressed by https://github.com/NomicFoundation/edr/pull/1425/changes/488686cc670117c92310b95f53a5b8ff5be104d5

crates/edr_solidity/src/nested_trace.rs
Wodann: Addressed by https://github.com/NomicFoundation/edr/pull/1425/changes/d8d10e2c0607d39ba124fd634fe9e69d80cbf995

crates/edr_solidity/src/trace_strategy.rs
Wodann: Addressed by https://github.com/NomicFoundation/edr/pull/1425/changes/d8d10e2c0607d39ba124fd634fe9e69d80cbf995

crates/edr_solidity/src/debug_info/dwarf.rs
Wodann: Addressed by https://github.com/NomicFoundation/edr/pull/1425/changes/4a179b4776c57c44cba47eb9201851c523e21c2a

crates/edr_solidity/src/compiler.rs
```diff
@@ -95,18 +98,36 @@ fn create_sources_model_from_ast(
...
+        .values()
+        .flat_map(HashMap::values)
+        .next()
+        .map(|contract| {
+            contract
+                .evm
+                .bytecode
+                .collect_ast_spans(&compiler_output.sources)
+        })
+        .unwrap_or_default();
```
Wodann: issue: While refactoring the code I ran into this piece of code, trying to access the polymorphic interface to collect AST spans of the `compiler_output` by accessing one of its artifacts. I changed the API to make the functionality directly accessible on the outer type.

crates/edr_solidity/src/debug_info.rs
Wodann: Addressed by https://github.com/NomicFoundation/edr/pull/1425/changes/8f508a60a2e4d51ecb55ec9d14915d80a6f77021

crates/edr_solidity/src/compiler.rs
Wodann: Addressed by https://github.com/NomicFoundation/edr/pull/1425/changes/8f508a60a2e4d51ecb55ec9d14915d80a6f77021

crates/edr_solidity/src/compiler.rs
Wodann: Addressed by https://github.com/NomicFoundation/edr/pull/1425/changes/8f508a60a2e4d51ecb55ec9d14915d80a6f77021

crates/edr_solidity/src/build_model.rs
Wodann: Addressed by https://github.com/NomicFoundation/edr/pull/1425/changes/8f508a60a2e4d51ecb55ec9d14915d80a6f77021

crates/edr_solidity/src/build_model.rs
```diff
@@ -89,12 +108,8 @@ impl SourceFile {
...
     /// Cached 1-based line number of the source location.
     line: OnceLock<u32>,
-    /// A weak reference to the source files mapping.
-    ///
-    /// Used to access the source file when needed.
-    pub(crate) sources: Weak<BuildModelSources>,
-    /// The file ID of the source file.
-    pub file_id: u32,
+    /// A weak reference to the source file.
+    source_file: Weak<RwLock<SourceFile>>,
```
Wodann, line 110: note: this change is not directly related to this PR but greatly simplifies the code and other changes.

Previously, we were passing in all `SourceFile` objects by weak reference and then doing a lookup using `file_id`. Instead we just pass in a weak reference to the specific `SourceFile`.

crates/edr_solidity/src/build_model.rs
```diff
@@ -226,7 +246,7 @@ pub struct ContractFunction {
     /// The type of the contract function.
     pub r#type: ContractFunctionType,
     /// The source location of the contract function.
-    pub location: Arc<SourceLocation>,
+    pub location: SourceLocation,
```
Wodann, line 247: note: `Arc` was not required, so switched to POD

crates/edr_solidity/src/error_inferrer.rs
```diff
@@ -167,8 +183,9 @@ pub(crate) fn filter_redundant_frames<HaltReasonT: HaltReasonTrait>(
                 return true;
             }
 
-            // this is probably a recursive call
-            if *idx > 0
+            // Same-location same-variant consecutive frames = recursion.
+            // The strategy dictates the minimum idx at which this filter applies.
+            if *idx >= recursion_start_idx
```
Wodann: @marianfe after looking into this deeper, I wonder whether this change is correct.

How can the first frame (i.e. `idx == 0`) ever be a recursive call? It would be the top of the stack, so if my understanding is correct, recursion cannot have occurred yet.

crates/edr_solidity/src/artifacts/solc.rs
```diff
@@ -0,0 +1,362 @@
...
+        });
+    }
+
+    extract_solc_contract_metadata(solc_version, input, output)
+        .map_err(SplitCompilerMetadataParseError::from)
+}
+
+/// Extracts [`IdentifiedContract`]s from the provided solc compiler input and
+/// output.
+pub fn extract_solc_contract_metadata(
```
popescuoctavian, line 175: Question: I believe that the initial implementation skipped unsupported solc versions (with `filter_on_solc_version`), while this is hard-failing for any build info that has an unsupported version.

Just checking that this is on purpose and complies with the intended design.

crates/edr_solidity/src/error_inferrer.rs
```diff
@@ -182,13 +181,14 @@ pub(crate) fn filter_redundant_frames<HaltReasonT: HaltReasonTrait>(
...
 
-            // Same-location same-variant consecutive frames = recursion.
-            // The strategy dictates the minimum idx at which this filter applies.
-            if *idx >= recursion_start_idx
-                && mem::discriminant(*frame) == mem::discriminant(next_frame)
-                && frame_source.range == next_frame_source.range
-                && frame_source.line == next_frame_source.line
-            {
+            let is_recursing = || {
+                *idx > 0
```
popescuoctavian, line 185: Question: Previously, this was determined using `recursion_start_idx`, which was removed. Solc and solx had different starting indexes under that implementation (1 vs 0). I believe that for solx the current implementation drops the idx = 0 frame when a recursive call is at that frame, which was not done before. Is that correct?

crates/edr_solidity/src/debug_info/dwarf.rs
```diff
@@ -124,9 +127,9 @@ fn is_stack_shuffle(opcode: OpCode) -> bool {
 /// [`Instruction`] vector that [`crate::source_map::decode_instructions`]
 /// produces for solc artifacts.
 pub fn decode_instructions(
```
popescuoctavian, line 131: Docs: There are a few `crate::source_map` broken links since `source_map` moved to `crate::artifacts::solc::source_map`.

popescuoctavian (approved):
Looks good to me overall! The refactoring makes sense and I like the move to a trait model.

I've left a couple of questions and a doc inconsistency. I don't think they are blocking, so I'm pre-approving this.

crates/edr_solidity/src/error_inferrer.rs
Wodann: That's correct. It wasn't clear to me why that change was introduced to me. I asked Marian about it in https://github.com/NomicFoundation/edr/pull/1425#discussion_r3583008432

It seems weird to me that the first frame of a call stack can be flagged as a recursive call, given that no recursion can have taken place yet (given the first frame should be the entry-point).

crates/edr_solidity/src/debug_info/dwarf.rs
Wodann: Thanks for catching that! Addressed by https://github.com/NomicFoundation/edr/pull/1425/changes/1e385671d8325f19604cc78c7c6729eaed82035f

I don't know why the docs CI check didn't catch that 🤔

crates/edr_solidity/src/artifacts/solc.rs
Wodann: Good catch! Addressed by https://github.com/NomicFoundation/edr/pull/1425/changes/e40cf92cfadfdabe78c6cf1fbb65e483f6f2bf84

We might want to change this in the future to turn it into an error instead of silently failing; but for now, I've left the behaviour as is and turn the error into an empty `Vec`.

Wodann (approved):
I've made some smaller changes and a larger refactor to get the code to a mergeable state.

The only thing I'm not certain about is the trait `TraceStrategy` but—to be honest—I don't fully understand where the variance comes from that's defined there.

One of the differences (the recursion index) seemed to not make a difference and I didn't understand the reason for it given that recursion can not happen at the first frame of a call stack anyway, as far as I understand.

I asked a question about it [here](https://github.com/NomicFoundation/edr/pull/1425#discussion_r3583008432) but for now I have returned the code to its pre-PR behaviour given that all tests succeed with the original logic.

For now, I believe this is good enough to merge. I can make further improvements to the code in follow-up PRs, if necessary.

crates/edr_solidity/src/debug_info/dwarf.rs
popescuoctavian: I think it should be caught. I took a note to look into why that isn't the case.

PR #1427, opened 2026-05-20
ci: pin benchmark jobs to edr-benchmark-runner

nebasuke:
The bare `self-hosted` label matched any runner in the org pool, including `solx-hardhat-shared-latitude-runner`, causing benchmark jobs to shuffle between two machines with materially different performance. The shuffling produced false-positive perf regressions on PRs whenever the PR landed on the slower runner against a baseline produced on the faster one.

Pin all three benchmark jobs to the new `edr-benchmark-runner` custom label (attached to `edr-gh-actions-runner` only), so every observation and comparison comes from the same physical machine.

### DIFF
--- a/.github/workflows/edr-benchmark.yml
+++ b/.github/workflows/edr-benchmark.yml
@@ -23,7 +23,7 @@ jobs:
   benchmarks-test:
     name: Benchmarks test
     environment: github-action-benchmark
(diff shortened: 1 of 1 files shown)


anaPerezGhiglia (approved):
Thanks for spotting this and fixing it! 🙌

PR #1429, opened 2026-05-21
chore(dev): simplify edr_napi `build:dev` script

anaPerezGhiglia:
While developing EDR, running `pnpm -C crates/edr_napi build:dev` is needed for generating/updating napi interface files.

Currently this script generates `TimeMock` related class and functions, that shouldn't be checked in the repository, so the developer needs to take care of ignoring those changes when committing.

This PR proposes removing `test-mock` feature flag from `build:dev` and instead create a new `build:test` script that enables it, and hook this new script in the `pretest` lifecycle script


crates/edr_napi/package.json
```diff
@@ -61,17 +61,18 @@
     "artifacts": "napi artifacts",
     "build": "pnpm run build:publish",
     "build:debug": "bash ../../scripts/build_edr_napi.sh --features op",
-    "build:dev": "bash ../../scripts/build_edr_napi.sh --release --features op,test-mock",
+    "build:dev": "bash ../../scripts/build_edr_napi.sh --release --features op",
```
Wodann, line 64: Thought: `build:dev` and `build:typingFile` now do the exact same thing. Consider removing `build:typingFile` to de-duplicate

Wodann (approved):

crates/edr_napi/package.json
anaPerezGhiglia: Not quite, since `build:typingFile` does not set `--release` flag. 

However, searching the usage the `build:typingFile`, is only used in `"Check that edr_napi typings file is up to date"` job, which I think it would be ok to run it with `--release`

Wodann: My bad.

Then there is value in the `build:typingFile`. It's a command meant for the "shortest build time" to generate an updated `index.d.ts` and `index.js` file. I use it all the time for that purpose. It's just a bit verbose. Once your PR is merged, I'll just rename it to `build:api` to make it shorter 😅

PR #1430, opened 2026-05-22
improvement: print full eyre::Report chain

Wodann:
When converting `eyre::Report`s to strings, we were only displaying the outermost error message for `eyre::Report` resulting in non-descriptive error messages.

Per the `eyre::Report` documentation:

```rs
/// To print causes as well using eyre's default formatting of causes, use the
/// alternate selector "{:#}".
```

This should allow end-users to better understand the reason for failures and debug issues with e.g. their `SolidityTestRunnerConfig`.

**Before**

```bash
  1) TransactionGasCapTest#setUp()
    Error: EVM error
```

**After**

```bash
  1) TransactionGasCapTest#setUp()
    Error: EVM error: transaction validation error: transaction gas limit (9223372036854775807) is greater than the cap (30000000): transaction gas limit (9223372036854775807) is greater than the cap (30000000)
```

### DIFF
--- a/.changeset/dry-melons-care.md
+++ b/.changeset/dry-melons-care.md
@@ -1,0 +1,5 @@
+---
+"@nomicfoundation/edr": patch
+---
+
+Print more detailed error descriptions for EVM, invariant fuzz, and cheatcode errors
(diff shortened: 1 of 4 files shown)


anaPerezGhiglia (approved):
💯

PR #1437, opened 2026-05-22
Introduce `estimate_gas_avoid_internal_oog` provider option.

### LINKED ISSUE #1455: Optional gas estimation mode that avoids internal OOG errors in non-failing calls
## Problem
The current gas estimation algorithm estimates the minimum gas required for the top-level call to succeed. For contracts that make non-failing calls to other contracts, this returns a value that lets the top-level call succeed while the internal call might fails with an out-of-gas (OOG) error, a technically valid estimate for the outer call, but broken behavior for the inner one.

## Goal
Offer an alternative gas estimation mode that avoids internal OOG failures in non-failing calls. The default estimation keeps its current behavior; users opt in to the new mode, ideally without relying on a separate second estimation pass.

alcuadrado:
This new option changes teh gas estimation algorithm to make it more friendly to contracts that use non-failing calls to other contracts.

Without this option, if a contract uses a non-failing call to another one, the gas estimation will always lead to the internal call failing due to an OOG error. The reason for this is that the gas estimation tries to estimate the minimum gas required for the top-level call to succeeed.

With this option, the gas estimation takes internal OOG errors into account, trying to improve on the original estimation if there's one. It does that by running a second estimation to look for the miimum gas that would produce no internal OOG error. When it can't find one, it default to the original estimation.

The reason to make this a second pass was simplicity instead of trying to combine everything into a single pass was keeping it simple, and the fact that most contract's don't use non-failing calls to others, so inner OOG errors tend to result in a topl-level call error.


anaPerezGhiglia:
Superseded by #1471

PR #1445, opened 2026-05-28
build: upgrade REVM to v38.0

popescuoctavian:
Upgrade REVM v34 → v38 (tag v107)

Upgrades REVM from v34 (`v103`) to v38 (`v107`), absorbing the API breaks across v104–v107 and the alloy 1.0 → 2.0 migration forced by `revm-inspectors 0.39` / `foundry-fork-db 0.26` / `op-revm 19+`.

## 1. Dependency version changes

All edits in the workspace root `Cargo.toml` (`[workspace.dependencies]`).

### REVM crates

| Crate | Before | After |
| --- | --- | --- |
| `revm` | 34.0.0 | 38.0.0 |
| `op-revm` | 15.0.0 | 19.0.0 |
| `revm-bytecode` | 8.0.0 | 10.0.0 |
| `revm-context` | 13.0.0 | 16.0.1 |
| `revm-context-interface` | 14.0.0 | 17.0.1 |
| `revm-database-interface` | 9.0.0 | 11.0.1 |
| `revm-handler` | 15.0.0 | 18.1.0 |
| `revm-inspector` | 15.0.0 | 19.0.0 |
| `revm-interpreter` | 32.0.0 | 35.0.1 |
| `revm-precompile` | 32.0.0 | 34.0.0 |
| `revm-primitives` | 22.0.0 | 23.0.0 |
| `revm-state` | 9.0.0 | 11.0.1 |

### Ecosystem crates (required by revm 38 / `revm-inspectors 0.39`)

| Crate | Before | After |
| --- | --- | --- |
| `revm-inspectors` | 0.34.2 | 0.39.0 |
| `foundry-fork-db` | 0.23 | 0.26 |
| `alloy-*` meta crates (consensus, eips, network, provider, rpc-client, rpc-types, signer family, transport family, …) | 1.0.22 | 2.0.5 |
| `alloy-hardforks` | 0.4.0 | 0.4.7 |
| `alloy-node-bindings` | 0.13.0 | 2.0.5 |
| `alloy-signer` | 1.6.1 | 2.0.5 |
| `alloy-dyn-abi`, `alloy-json-abi`, `alloy-primitives`, `alloy-sol-*` | 1.x | 1.5.2 (revm 38 requirement) |
| `tower` | 0.4 | 0.5 |
| `tower-http` | 0.5 | 0.6 |
| `reqwest` | 0.12 | 0.13 (with `rustls` feature) |

New direct deps: `alloy-rpc-types-engine` (`jwt` feature, for the foundry runtime transport) and `alloy-rpc-types-trace` (for `edr_provider`), both `2.0.5`.

## 2. Upstream breaks absorbed

- **v104 (revm 35)** — the bulk: `Bytecode` flattened enum → struct; `ExecutionResult` gas fields collapsed into `gas: ResultGas` + logs added to `Revert`/`Halt`; `EthPrecompiles`/`EthInstructions` lost `Default`; alloy-core fixed-byte maps (`AddressMap`/`B256Map`/`U256Map`); `BlockHashCache` ring buffer; `JournalInner::cfg.spec`; Amsterdam EIPs (7843/7708/8024); default hardfork → Osaka.
- **v105 (revm 36)** — version bumps only, no code changes.
- **v106 (revm 37)** — `EVMError::CustomAny` (`AnyError`); `CallInputs::known_bytecode` no longer `Option`; EIP-8037 state-gas reservoir (`initial_total_gas`, expanded `validate_initial_tx_gas`); `Precompile` inner `PrecompileFn` accessors removed.
- **v107 (revm 38)** — handler `reservoir` plumbing; `PrecompileOutput::gas_refunded` restored.
- **alloy 1.0 → 2.0** (forced transitively) — `TransactionBuilder` split, blob-tx API reshuffle, `AnyTransactionReceipt` struct, `ChainConfig`/`BlockHeader` additions, `Claims`/`JwtSecret` moved to `alloy-rpc-types-engine`.

## 3. Notable code changes

- Rewrote `OverriddenPrecompileProvider::run` for the new `PrecompileFn` signature + `precompile_output_to_interpreter_result`.
- Threaded `EvmState` / `U256Map` typed maps through `DatabaseCommit`, `StateDiff`, and the foundry backend.
- `ExecutionResult` → `ResultGas` (+ logs) across tracing, napi, foundry executors.
- Replaced the precompile-provider `Default` bound with a `new_precompile_provider(hardfork)` constructor; deleted the interim `DefaultEthPrecompiles` wrapper.
- alloy 2.0 transport fixes; added `rustls` to `foundry-compilers` and `aws-lc-rs` to `alloy-transport-ws` to restore TLS for solc downloads and `wss://` forks.
- Hardhat `assertEqualTraces`: normalize `refund: 0` and `0x`-prefixed `memory`/`storage` (geth v1.17.3+ / revm-inspectors 0.39 wire-format changes).
- Cooldown downgrades of `hyper`/`reqwest` in `Cargo.lock`.


crates/edr_solidity_tests/tests/it/helpers.rs
```diff
@@ -145,6 +145,11 @@ impl ForgeTestProfile {
             ffi: true,
             memory_limit: 1 << 26,
             spec: hardfork,
+            // Solidity tests intentionally use `gas_limit = u64::MAX`, so opt
+            // out of the EIP-7825 transaction gas cap (active by default for
+            // OSAKA and later).
+            transaction_gas_cap: None,
+            disable_transaction_gas_cap: true,
```
popescuoctavian: @Wodann this was flagged by Claude and I believe it is a valid point. Could you please take a look and tell me if I am missing something?

crates/foundry/evm/evm/src/inspectors/stack.rs
```diff
@@ -1346,7 +1350,15 @@ impl<
...
-                    call.known_bytecode = None;
+
+                    let target_account = ecx
+                        .journal_mut()
+                        .load_account_with_code(*target)
+                        .expect("failed to load account");
+                    call.known_bytecode = (
+                        target_account.info.code_hash,
+                        target_account.info.code.clone().unwrap_or_default(),
+                    );
```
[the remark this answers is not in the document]
popescuoctavian: This is on-par with [Foundry's inspector implementation](https://github.com/foundry-rs/foundry/blob/7a3ff694a97b87f2214aa79662fabc621bbf75a4/crates/evm/evm/src/inspectors/stack.rs#L1266).

crates/chain/spec/evm/src/error.rs
```diff
@@ -55,6 +55,7 @@ impl<DatabaseErrorT: DBErrorMarker + std::error::Error>
     fn from(value: EVMError<DatabaseErrorT, EvmTransactionValidationError>) -> Self {
         match value {
             EVMError::Custom(error) => Self::Custom(error),
+            EVMError::CustomAny(error) => Self::Custom(error.to_string()),
```
Wodann: suggestion: we are potentially throwing away valuable information that can help a caller debug this error. Ideally, we'd add a new `CustomAny(error)` to the `TransactionError` enum and pipe it through to call sites.

crates/edr_common/src/fmt/ui.rs
```diff
@@ -898,6 +898,7 @@ fn pretty_block_basics<T>(block: &Block<T, alloy_rpc_types::Header<AnyHeader>>)
                         excess_blob_gas,
                         parent_beacon_block_root,
                         requests_hash,
+                        ..
```
Wodann: issue: we're not printing the newly added values.

crates/chain/spec/evm/Cargo.toml
```diff
@@ -11,6 +11,7 @@ revm-context = { workspace = true, features = ["optional_balance_check", "option
 revm-handler.workspace = true
 revm-inspector.workspace = true
 revm-interpreter.workspace = true
+revm-primitives.workspace = true
```
Wodann: question: this dependency was added, but it doesn't appear to be used in any of the diffs. Is it a leftover from a prior edit?

crates/edr_napi/src/result.rs
```diff
@@ -161,17 +161,16 @@ impl From<&AfterMessage<EvmHaltReason>> for ExecutionResult {
...
                 output,
             } => {
                 let logs = logs.iter().map(ExecutionLog::from).collect();
 
                 Either3::A(SuccessResult {
                     reason: SuccessReason::from(*reason),
-                    gas_used: BigInt::from(*gas_used),
-                    gas_refunded: BigInt::from(*gas_refunded),
+                    gas_used: BigInt::from(gas.tx_gas_used()),
+                    gas_refunded: BigInt::from(gas.inner_refunded()),
```
Wodann: thought: It isn't immediately clear to me whether we should be using `final_refunded` or `inner_refunded`. I am not sure whether we have tests that tests the difference between these two occurrences to validate it.

It would be good to add those (at some point) and validate that this API is the correct one.

crates/edr_napi/src/result.rs
```diff
@@ -190,18 +189,18 @@ impl From<&AfterMessage<EvmHaltReason>> for ExecutionResult {
                     },
                 })
             }
-            edr_chain_spec_evm::result::ExecutionResult::Revert { gas_used, output } => {
+            edr_chain_spec_evm::result::ExecutionResult::Revert { gas, output, .. } => {
```
Wodann: suggestion(non-blocking): we can now return logs that occurred during reverts and halts. This would be a change in the N-API but it would be good to bubble up for usage by the end user.

crates/edr_op/src/spec.rs
```diff
@@ -63,6 +63,7 @@ fn cast_evm_error<DatabaseErrorT: Debug + std::error::Error>(
 ) -> TransactionError<DatabaseErrorT, InvalidTransaction> {
     match error {
         EVMError::Custom(error) => TransactionError::Custom(error),
+        EVMError::CustomAny(error) => TransactionError::Custom(error.to_string()),
```
Wodann: suggestion: avoid erasing the extra information of the `CustomAny` and bubble it up instead.

crates/chain/spec/evm/src/lib.rs
```diff
@@ -41,8 +41,15 @@ pub trait EvmChainSpec:
...
     /// Type representing a precompile provider.
-    type PrecompileProvider<BlockT: BlockEnvTrait, DatabaseT: Database>: Default
-        + PrecompileProvider<ContextForChainSpec<Self, BlockT, DatabaseT>, Output = InterpreterResult>;
+    type PrecompileProvider<BlockT: BlockEnvTrait, DatabaseT: Database>: PrecompileProvider<
+        ContextForChainSpec<Self, BlockT, DatabaseT>,
+        Output = InterpreterResult,
+    >;
+
+    /// Constructs the precompile provider for the given hardfork.
+    fn new_precompile_provider<BlockT: BlockEnvTrait, DatabaseT: Database>(
```
Wodann, line 50: praise: nice! I'm happy this now guarantees the hardfork upon construction

crates/edr_runtime/src/transaction.rs
```diff
@@ -35,12 +35,19 @@ pub fn validate<TransactionT: Transaction>(
     spec_id: EvmSpecId,
 ) -> Result<TransactionT, CreationError> {
     const EIP7623_DISABLED: bool = false;
+    const EIP8037_AMSTERDAM_ENABLED: bool = false;
```
Wodann: issue: this should be determined at runtime, depending on the `spec_id`

Wodann: Addressed by https://github.com/NomicFoundation/edr/pull/1445/changes/460b2fdf3cd37ebcd7a8d40ed33f57cc55a76ce8

crates/edr_solidity_tests/src/config.rs
```diff
@@ -141,7 +141,10 @@ impl<HardforkT: HardforkTr> SolidityTestRunnerConfig<HardforkT> {
             isolate: false,
             disable_block_gas_limit: false,
             transaction_gas_cap: None,
-            disable_transaction_gas_cap: false,
+            // Solidity tests intentionally run with very large gas limits
```
Wodann: thought: @kanej explicitly asked for this to have the same default as the JSON-RPC provider; hence why we changed this default. This would revert that request.

It would be good to be in alignment on whether we want to support setting of the EIP-7825 transaction gas limit for Solidity tests.

crates/edr_solidity_tests/tests/it/helpers.rs
```diff
@@ -145,6 +145,11 @@ impl ForgeTestProfile {
             ffi: true,
             memory_limit: 1 << 26,
             spec: hardfork,
+            // Solidity tests intentionally use `gas_limit = u64::MAX`, so opt
```
Wodann: thought: see https://github.com/NomicFoundation/edr/pull/1445/changes#r3351744087

crates/foundry/evm/core/src/backend/mod.rs
```diff
@@ -742,7 +742,7 @@ impl<
     pub fn replace_account_storage(
         &mut self,
         address: Address,
-        storage: Map<U256, U256>,
+        storage: alloy_primitives::map::U256Map<U256>,
```
Wodann: Nit: consider using a `use` statement

crates/foundry/evm/evm/src/executors/mod.rs
```diff
@@ -1712,28 +1708,32 @@ fn convert_executed_result<
...
-        } => (reason.into(), gas_refunded, gas_used, Some(output), logs),
-        ExecutionResult::Revert { gas_used, output } => {
+        } => (
+            reason.into(),
+            gas.inner_refunded(),
+            gas.tx_gas_used(),
+            Some(output),
+            logs,
+        ),
+        ExecutionResult::Revert { gas, output, .. } => {
```
Wodann: question: Do we want to bubble up the `logs` for reverts and halts?

crates/foundry/evm/evm/src/inspectors/stack.rs
```diff
@@ -1346,7 +1350,15 @@ impl<
                         .and_then(|selector| mocks.get(selector))
                 }) {
                     call.bytecode_address = *target;
-                    call.known_bytecode = None;
+
+                    let target_account = ecx
+                        .journal_mut()
+                        .load_account_with_code(*target)
+                        .expect("failed to load account");
```
Wodann, line 1357: question: Instead of `.expect`ing, can we bubble up the error? E.g. when using a remote provider that fails?

crates/tracing/src/lib.rs
```diff
@@ -10,16 +10,42 @@ use edr_chain_spec::HaltReasonTrait;
...
 
+/// Build a [`ResultGas`] from a [`Gas`] accumulator at end-of-execution. Used
+/// for Success/Revert outcomes that did not consume the entire gas limit.
+///
+/// The `floor_gas` argument is `0`: the EIP-7623 calldata floor is a
+/// transaction-level value computed during validation and is not available on
+/// the [`Gas`] accumulator the inspector observes. `total_gas_spent()` (the
+/// figure trace consumers read) is unaffected; only
+/// [`ResultGas::final_refunded`] would differ, and that is not surfaced through
+/// tracing.
```
Wodann: thought: I'm still unsure of how `final_refunded` works, so I can't be sure that this expressed invariants holds in reality.

crates/tracing/src/lib.rs
```diff
@@ -10,16 +10,42 @@ use edr_chain_spec::HaltReasonTrait;
...
+fn result_gas_from(gas: &Gas) -> ResultGas {
```
Wodann: nit: this name is a little confusing as it doesn't specify what the `result_gas` is constructed "from"

Wodann (changes requested):
Thank you for upgrading REVM! 🙏 

I noticed a couple of issues that I fixed manually:
- usage of new API for validating a transaction's initial gas
- there were a lot of duplicate crates being introduced. By upgrading `op-alloy-rpc-types` 0.17.2 → 2.0.0, I was able to collapse the entire duplicated alloy 1.7.3 family onto 2.0.5.
   - as part of this, I mapped the new `da_footprint_gas_scalar` field (Jovian) that op-alloy 2.0.0's L1BlockInfo requires.
   - pinned `reth-codecs`, `reth-codecs-derive`, `reth-zstd-compressors` to 0.3.0 (op-alloy's optional, never-compiled deps; ^0.3.0 allows it) to clear the 7-day cooldown.

We might want to update the REVM upgrade skill to keep an eye on the  `op-alloy-rpc-types` crate too.

I left questions and comments for some other issues/suggestions.

crates/chain/spec/evm/src/error.rs
popescuoctavian: Addressed by 15a1e731bf3b880149e386c47225d7f2f4362915.

crates/edr_common/src/fmt/ui.rs
popescuoctavian: Missed that, thanks! Addressed by def4609fdf763ee46f10789f4be342d54f573c1c.

crates/chain/spec/evm/Cargo.toml
popescuoctavian: That is correct, it was a leftover after I did some refactoring of the initial code. Addressed by 9568d214f82ae3ee3ccfbabb49cef00a643311c1.

crates/edr_napi/src/result.rs
popescuoctavian: Addressed by bb4d9e52dbe912801d5b2c6a409c4ab1f2713bb1.

Also added a changeset for this: 7a6a952f52e9c5fa029768c87e6de37f6cd06f49.

crates/edr_op/src/spec.rs
popescuoctavian: Addressed by 15a1e731bf3b880149e386c47225d7f2f4362915.

crates/foundry/evm/core/src/backend/mod.rs
popescuoctavian: Addressed by 5686b02858df7727fa307719fd9dbd3a8f77ecc7.

crates/foundry/evm/evm/src/executors/mod.rs
popescuoctavian: Addressed by 606a857e33a837eac1a426e9ffe383924315f352.

Just as a side note, execution logs are just a fallback option for Solidity tests, used only if inspector collected logs are empty.

crates/foundry/evm/evm/src/inspectors/stack.rs
popescuoctavian: This is not trivial. Since the function returns `Option<CallOutcome>`, one solution would be to set the context error and return `FatalExternalError`. I'm not sure this is ideal since it's difficult to track the cause.

As I have noted here https://github.com/NomicFoundation/edr/pull/1445#discussion_r3349435280, the current `expect` in on par with Foundry's implementation of this code.

Let me know if you think the first option is worth implementing.

crates/edr_napi/src/result.rs
popescuoctavian: IIUC, `inner_refunded` and `final_refunded` diverge only when EIP-7623 calldata floor gas is active and the statement `total_gas_spent - refunded < floor_gas` is true. In that case:
- `inner_refunded` = the `refunded` field, regardless of floor gas
- `final_refunded` = the effective refund: 0 when floor gas absorbs the refund, otherwise `refunded` (same as `inner_refunded`).

Considering this, I believe the code should be changed to use `final_refunded` instead, which means that the tracing `floor_gas = 0` [initialization](https://github.com/NomicFoundation/edr/pull/1445#discussion_r3351820058) doesn't hold up. If that's the case, I can look into our options to make that accurate. What do you think?

crates/tracing/src/lib.rs
popescuoctavian: See https://github.com/NomicFoundation/edr/pull/1445#discussion_r3357257487.

crates/tracing/src/lib.rs
popescuoctavian: Addressed by b1c7f09518c2d966f276df048dd0f52843b23b62.

crates/edr_solidity_tests/src/config.rs
popescuoctavian: Addressed by 04b128154243b0a8ebd94ef1f6324f0e42bd3b0a. Following an offline discussion, I have reverted the change since it is not directly related to the REVM upgrade. This will be investigated on the side and, if necessary, updated in a follow up PR.

crates/edr_solidity_tests/tests/it/helpers.rs
popescuoctavian: See: https://github.com/NomicFoundation/edr/pull/1445#discussion_r3357655780

crates/edr_solidity_tests/tests/it/helpers.rs
popescuoctavian: See: https://github.com/NomicFoundation/edr/pull/1445#discussion_r3357655780

crates/edr_solidity_tests/src/config.rs
popescuoctavian: 04b128154243b0a8ebd94ef1f6324f0e42bd3b0a caused most of the Solidity tests we have to fail since the `gas_limit = i64::MAX` was bigger than the Osaka-enforced cap. I've added df86c575ef37054ed95ec2b88639608c67ded6ad to address that. 

The following table shows how the transaction gas cap -> default gas limit are enforced in Solidity tests after that fix.   Assuming no explicit `gasLimit` and no `blockGasLimit`:

  | Hardfork  | `transactionGasCap` | `disableTransactionGasCap` | Resolved gas limit             |
  | --------- | ------------------- | -------------------------- | ------------------------------ |
  | pre-Osaka | unset               | `false`                    | `i64::MAX` (no cap exists)      |
  | pre-Osaka | unset               | `true`                     | `i64::MAX`                      |
  | pre-Osaka | `C`                 | `false`                    | `C`                             |
  | pre-Osaka | `C`                 | `true`                     | `i64::MAX`                      |
  | Osaka+    | unset               | `false`                    | **`16_777_216`** (hardfork default cap) |
  | Osaka+    | unset               | `true`                     | `i64::MAX`                      |
  | Osaka+    | `C`                 | `false`                    | `C`                             |
  | Osaka+    | `C`                 | `true`                     | `i64::MAX`                      |

crates/edr_napi/src/result.rs
Wodann: That's my understanding as well. I wasn't sure which of the two values should be used in the execution receipt, though.

After investigating Foundry, I think we should change this to `final_refunded` in the `InspectorStackRefMut` and `convert_executed_result` for sure. Foundry does the same. See:
1. https://github.com/foundry-rs/foundry/blob/master/crates/evm/evm/src/inspectors/stack.rs#L934
2. https://github.com/foundry-rs/foundry/blob/master/crates/evm/evm/src/executors/mod.rs#L1268

For the N-API entry, we don't have a reference to decide but I think we should also use `final_refunded`. My motivation is:

The code in question reports the transaction's execution result, not an intermediate state. As such, we should always report the post-floor gas refund:

```rs
/// The result when the EVM terminates successfully.
#[napi(object)]
pub struct SuccessResult {
    /// The reason for termination
    pub reason: SuccessReason,
    /// The amount of gas used
    pub gas_used: BigInt,
    /// The amount of gas refunded
    pub gas_refunded: BigInt,
    /// The logs
    pub logs: Vec<ExecutionLog>,
    /// The transaction output
    pub output: Either<CallOutput, CreateOutput>,
}
```

IFF we were also reporting internals of the gas tracking state in steps, we should return the state of the actual refund counter there, but we don't atm.

crates/foundry/evm/evm/src/inspectors/stack.rs
Wodann: Based on our error reporting strategy, ideally we'd do better than Foundry here. IIRC, we do exactly what you propose in the provider.

If it's a lengthy implementation, feel free to do it in a follow-up PR.

Wodann (approved):
Nice work on the improvements!

Approving this pending the change of `gas.inner_refunded` to `gas.final_refunded` in 3 call sites.

crates/edr_napi/src/result.rs
popescuoctavian: Addressed by 6af7c2913a95d68018fd51449608bc994b1bf319.

popescuoctavian:
Socket warning is triggered by test code. Their analysis:

>No malicious behavior detected. This fragment is a test module for hyper legacy client behavior using a mock connector to simulate I/O, connection errors, and HTTP upgrade flows. It demonstrates data flows through mocked inputs to the client and back as responses for assertion, with verbose logging for debugging. The structure is safe for its intended testing purpose, though the extensive debug output could be noisy in production builds. Security risk is low to moderate due to test scaffolding, not production code paths.

Ignoring the warning.

popescuoctavian:
@SocketSecurity ignore cargo/hyper-util@0.1.20

PR #1447, opened 2026-05-30
refactor(trie): migrate off abandoned triehash to alloy-trie

nebasuke:
Noticed when tackling https://github.com/NomicFoundation/edr/pull/1433. The `hash-db` upgrade is impossible due to `triehash` being unmaintained. Did some Claude assisted research and found we can quite cleanly replace triehash with `alloy-trie` which then also allows us to cut `edr_trie`. 

Feel free to shout at me for Rust style.

# Claude Summary
## Motivation

`edr_trie` was a thin wrapper around `triehash` 0.8.4 — last released in **2021**, on the now-archived `paritytech/trie`. It pins `hash-db` 0.15.2, which permanently blocks any `hash-db` upgrade (see #1433, where Renovate's `hash-db` 0.16 bump can't compile because `KeccakHasher` would implement a different `hash_db::Hasher` than the one `triehash` requires).

All of `edr_trie`'s functionality has direct equivalents in **alloy-trie**, which is already a dependency (used by `edr_state_persistent_trie`). This removes the dead dependency stack and unblocks `hash-db`.

## Changes

- `ordered_trie_root` → `alloy_trie::root::ordered_trie_root_with_encoder` at the 4 call sites (tx / receipts / withdrawals roots). The encoder writes into a reused buffer, dropping a per-item `Vec<u8>` allocation vs the prior `.map(alloy_rlp::encode)`.
- `sec_trie_root` (state/storage root in `edr_state_api`) → `alloy_trie::root::{state_root_ref_unhashed, storage_root_unhashed}` (the `ethereum` feature — reth's own helpers).
- `trie_root` — unused, dropped.
- `crates/trie` removed; `triehash`, `hash-db`, and `hash256-std-hasher` drop out of the tree.

`BasicAccount` is field-identical to `alloy_trie::TrieAccount` (same fields, same order, same `RlpEncodable` derive), so the state-root encoding is byte-for-byte unchanged — verified by the existing `precompiles_state_root` golden test.

## Verification

- `cargo check --workspace --all-targets`, `cargo +nightly fmt --check`, `cargo clippy --all-targets` — clean.
- `cargo test -p edr_state_api` — `precompiles_state_root` (hardcoded expected hash), `empty_state_root`, `empty_storage_root` pass.
- `cargo test -p edr_block_local -p edr_block_header -p edr_chain_l1` — pass (tx/withdrawals/receipts root paths).
- `triehash` / `hash-db` / `hash256-std-hasher` confirmed gone from `Cargo.lock`.

## Note

This obsoletes #1433 — once this lands, `hash-db` is no longer a dependency, so neither the 0.16 bump nor the pin are needed.


crates/state/api/src/account.rs
```diff
@@ -50,3 +50,18 @@ impl From<(&AccountInfo, B256)> for BasicAccount {
         }
     }
 }
+
+// `storage_root` and `code_hash` are both `B256`, so transposing them here
+// compiles but silently changes every state root (`state_root` encodes accounts
+// through this conversion). The `precompiles_state_root` test is the only
+// guard.
+impl From<BasicAccount> for alloy_trie::TrieAccount {
```
Wodann, line 58: Future work: https://github.com/NomicFoundation/edr/issues/1460

Wodann (approved):
Thanks for making these changes.

There was some unnecessary boilerplate code, which I removed in this commit: https://github.com/NomicFoundation/edr/pull/1447/changes/ae6c6f3902bdbe3bc9f33526d82fbae822da62db

PR #1453, opened 2026-06-01
chore: remove unused tower-http workspace dependency

nebasuke:
Similar to https://github.com/NomicFoundation/edr/pull/1446, `tower-http` is an unused dependency (used by Anvil).


anaPerezGhiglia (approved):
LGTM

PR #1454, opened 2026-06-01
chore: remove unused dependencies

anaPerezGhiglia:
used `cargo machete` as input and Claude to identify false positives


Wodann (approved):
LGTM; thanks for doing this!

PR #1461, opened 2026-06-05
refactor(provider): introduce ObservedExecution to encapsulate EVM observer lifecycle

anaPerezGhiglia:
## Context
While implementing gas estimation improvements (#1455), several structural issues in the existing code became apparent: the EVM observer lifecycle was repeated manually at every call site, and `estimate_gas` logic was spread across `data.rs` rather than living in `gas.rs` alongside its helpers. This PR addresses those improvements independently to avoid mixing concerns and enable focused review

## Changes
Introduces `ObservedExecution<T>` and `observe_execution` in `observability.rs` to encapsulate the observer lifecycle (create → run → collect_and_report), replacing the manual pattern repeated at every EVM call site.

### Key changes:
- `observe_execution` wraps observer creation and collection into a single call
- `ObservedExecution::into_result_and_traces` provides a consuming accessor for the execution result and filtered call traces
- `EvmObservedData::into_call_traces` / `CallResultWithMetadata::into_call_traces` are consuming filter methods that replace inline if/else blocks
- `estimate_gas` is moved from `data.rs` into `gas.rs`, with `check_gas_limit` and `binary_search_estimation` refactored to take a shared `Context` struct; this also eliminates the one `call_trace_arena.clone()` that existed in the previous implementation


crates/edr_provider/src/data/gas.rs
```diff
@@ -1,121 +1,94 @@
...
+            self.header,
+            self.cfg_env.spec,
+            self.scheduled_blob_params,
+        )
+    }
 }
 
 /// Test if the transaction successfully executes with the given gas limit.
 /// Returns true on success and return false if the transaction runs out of gas
 /// or funds or reverts. Returns an error for any other halt reason.
```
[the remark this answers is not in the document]
anaPerezGhiglia: addressed by [cfe3428](https://github.com/NomicFoundation/edr/pull/1461/commits/cfe34288aaa1f937fe68baf854c85bf1631f2dfd)

crates/edr_provider/src/data/gas.rs
```diff
@@ -1,121 +1,93 @@
...
-    ProviderError,
+    error::{EstimateGasFailure, ProviderErrorForChainSpec, TransactionFailure},
+    observability::{observe_execution, EvmObserver, EvmObserverConfig, ObservedExecution},
+    time::TimeSinceEpoch,
+    ProviderError, ProviderSpec,
 };
 
-pub(super) struct CheckGasLimitArgs<'a, HardforkT, SignedTransactionT> {
+/// Shared EVM execution context passed to all gas estimation functions.
+pub(super) struct Context<'a, ChainSpecT: ChainSpec + HardforkChainSpec> {
```
Wodann: suggestion: I recommend using a more descriptive type name like `GasEstimationContext` to distinguish it from other context types. With this name doing lookups using the type symbol will be hard.

crates/edr_provider/src/data/gas.rs
```diff
@@ -171,6 +121,116 @@ pub(super) fn binary_search_estimation<
...
+pub(super) fn estimate_gas<
+    TimerT: Clone + TimeSinceEpoch,
+    ChainSpecT: ProviderSpec<TimerT, SignedTransaction: TransactionMut>,
+>(
+    context: &Context<'_, ChainSpecT>,
+    contract_decoder: Arc<RwLock<ContractDecoder>>,
+    minimum_cost: u64,
+    observer_config: &EvmObserverConfig,
+) -> Result<EstimateGasResult, ProviderErrorForChainSpec<ChainSpecT>> {
+    // Measure the gas used by the transaction with optional limit from call request
```
Wodann, line 136: issue: I believe the HH2 benchmarks we use to track performance don't track this specific case, but I expect a performance regression for a scenario that:
- enables tracing
- calls `eth_estimateGas` often

The simplest way to reproduce this would be to add a criterion benchmark that calls `estimate_gas` on the `ProviderData` repeatedly.

The reason for this is that your code reconstructs the `EvmObserver` for every executed call and calling `collect_and_report`. Previously we were reusing the same `EvmObserver` and avoiding reallocation of memory by calling `flush_and_report`.

This is the difference:

```rs
    /// Collects the [`TracingInspector`]'s traces and ABI decodes them.
    pub fn collect(
        self,
        address_to_executed_code: &HashMap<Address, Bytes>,
        precompile_addresses: &HashSet<Address>,
    ) -> Result<CallTraceArena, serde_json::Error> {
        let mut arena = self.inspector.into_traces();

        let mut decoder = self.decoder.write();
        decoder.populate_call_trace_arena(
            &mut arena,
            address_to_executed_code,
            precompile_addresses,
        )?;

        Ok(arena)
    }

    /// Takes the [`TracingInspector`]'s traces and ABI decodes them, replacing
    /// the current traces with an empty arena.
    pub fn take(
        &mut self,
        address_to_executed_code: &HashMap<Address, Bytes>,
        precompile_addresses: &HashSet<Address>,
    ) -> Result<CallTraceArena, serde_json::Error> {
        let mut arena = std::mem::take(self.inspector.traces_mut());

        // Reset the inspector
        self.inspector.fuse();

        let mut decoder = self.decoder.write();
        decoder.populate_call_trace_arena(
            &mut arena,
            address_to_executed_code,
            precompile_addresses,
        )?;

        Ok(arena)
    }
```

`self.inspector.fuse()` keeps the `TracingInspector`'s memory.

This can be avoided by constructing the `EvmObserver` once and passing that same instance as an `&mut EvmObserver` to a secondary function `observe_execution_mut`—which doesn't construct the `EvmObserver` every time.

crates/edr_provider/src/observability.rs
```diff
@@ -311,3 +325,72 @@ impl<
         self.tracing_inspector.step(interp, context);
     }
 }
+
+pub(crate) trait WithExecutionResult {
```
Wodann, line 332: suggestion: Could you add documentation explaining what this trait it for?

Wodann (changes requested):
Thanks for simplifying the code. I have one major concern, re. performance impact of the change.

crates/edr_provider/src/data/gas.rs
Wodann: After discussing this on a call, I realised that this is in fact not a regression. This is mimicking the same behaviour as the existing code. As such, it's a non-concern for this PR.

I did create an issue to investigate this as a potential performance optimisation in the future: https://github.com/NomicFoundation/edr/issues/1467

Wodann (approved):

crates/edr_provider/src/observability.rs
anaPerezGhiglia: Addressed by [cd28569](https://github.com/NomicFoundation/edr/pull/1461/commits/cd2856956b95cdcdf9a845d990a1670f82865d53)

crates/edr_provider/src/data/gas.rs
anaPerezGhiglia: renamed to `GasCallContext` in [6fadf70](https://github.com/NomicFoundation/edr/pull/1461/commits/6fadf70e4669ba53fcd9e62beae2d36a299975ff)

PR #1471, opened 2026-06-12
feat(provider): add `gasEstimationMode` config to avoid internal out-of-gas in `eth_estimateGas`

### LINKED ISSUE #1455: Optional gas estimation mode that avoids internal OOG errors in non-failing calls

anaPerezGhiglia:
## Context
Currently `eth_estimateGas` searches for the lowest gas limit at which the top-level call succeeds. For contracts that catch internal call failures, that estimate can be too low: the transaction "succeeds" while an internal call silently runs out of gas.

## Changes

> Design doc: [Gas estimation mode that avoids internal OOG errors](https://app.notion.com/p/nomicfoundation/Gas-estimation-mode-that-avoids-internal-OOG-errors-37c578cdeaf580158477f6b611b31297#51c578cdeaf582538a0181142318dd64)

Add an optional `gasEstimationMode` provider configuration:

  - `TopLevelSuccess` (default): the existing behavior, unchanged.
  - `NoInternalOutOfGas`: an estimation is only accepted if, additionally, no internal sub-call ends in an out-of-gas error (detected from the recorded call traces).

When no OOG-free estimation exists, that is the transaction hits an internal OOG even at its full gas limit, estimation fails with a new `InternalCallOutOfGas` transaction failure.

Mode-specific success criteria and failure construction are encapsulated in `GasEstimationMode`; the estimation flow is mode-agnostic.

## Note to reviewers 👀 

  I tried to keep the commit history clean, but the implementation went through several refactors and renames along the way, so reviewing commit by commit may feel like re-reviewing the same code in different shapes. Reviewing the full diff in one pass is probably more efficient. However the commit history is still useful as a reference for tracing how a particular piece ended up the way it did.

closes #1455


crates/edr_provider/src/data/gas.rs
```diff
@@ -19,17 +23,26 @@ use edr_receipt::ReceiptTrait as _;
...
     time::TimeSinceEpoch,
-    ProviderError, ProviderSpec,
+    ProviderError, ProviderSpec, TransactionFailureReason,
 };
 
+pub struct EstimateGasResult {
+    pub call_trace_arenas: Vec<CallTraceArena>,
+    pub estimation: u64,
+}
+
```
anaPerezGhiglia, line 45: The struct existed already, I moved it from `crates/edr_provider/src/config.rs` since it belong in this module.

crates/edr_provider/src/error.rs
```diff
@@ -709,6 +709,9 @@ impl<HaltReasonT: HaltReasonTrait> std::fmt::Display for TransactionFailure<Halt
                 )
             }
             TransactionFailureReason::Inner(halt) => write!(f, "{halt:?}"),
+            TransactionFailureReason::InternalCallOutOfGas => {
+                write!(f, "Transaction internal call ran out of gas")
+            }
```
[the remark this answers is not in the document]
anaPerezGhiglia: fixed grammar in [a1569b1](https://github.com/NomicFoundation/edr/pull/1471/commits/a1569b166831386a0908e4ec9442126d0f5f5903)

crates/edr_napi/src/config.rs
```diff
@@ -149,6 +149,27 @@ pub struct HardforkActivationByTimestamp {
     pub timestamp: BigInt,
 }
 
+/// Controls the gas estimation strategy used by `eth_estimateGas`.
+#[napi]
+#[derive(Default)]
```
Wodann: nit: I don't think the N-API type requires a default.

crates/edr_provider/src/data/gas.rs
```diff
@@ -183,54 +324,41 @@ pub(super) fn estimate_gas<
...
-    call_trace_arenas.extend(estimation_call_trace_arenas);
+/// Executes the transaction with the given gas limit and returns whether the
+/// estimation mode accepts the execution, along with its call trace (when the
+/// trace collection criteria are met).
+fn probe_gas_limit<ChainSpecT: ProviderChainSpec<SignedTransaction: TransactionMut>>(
+    context: &GasCallContext<'_, ChainSpecT>,
+    observer_config: &EvmObserverConfig,
+    gas_limit: u64,
+    estimation_mode: GasEstimationMode,
+) -> Result<(bool, Option<CallTraceArena>), ProviderErrorForChainSpec<ChainSpecT>> {
```
Wodann: suggestion: consider wrapping the `SuccessT` in a struct to self-document returned values; e.g.

```rs
pub struct ProbeGasEstimationResult {
  success: bool,
  call_trace: Option<CallTraceArena>,
}
```

crates/edr_provider/src/data/gas.rs
```diff
@@ -132,10 +210,71 @@ pub(super) fn estimate_gas<
...
+    // This can happen if the execution behavior depends on the available gas.
+    // Search upwards for the lowest gas limit the mode accepts.
+    let EstimateGasResult {
+        call_trace_arenas: estimation_call_trace_arenas,
+        estimation,
+    } = binary_search_estimation::<ChainSpecT>(
+        context,
+        lowest_estimation,
+        context.transaction.gas_limit(), /* transaction original gas_limit was validated on
+                                          * `measure_gas_with_full_limit` */
```
Wodann: nit + typo: odd comment styling. Consider moving it above this line as a `//`

```suggestion
        // transaction's `gas_limit` was validated by `measure_gas_with_full_limit`
        context.transaction.gas_limit(),
```

crates/edr_scenarios/src/lib.rs
```diff
@@ -152,6 +152,9 @@ impl From<ScenarioProviderConfig> for ProviderConfig {
             // Overriding precompiles is not supported in scenarios
             precompile_overrides: HashMap::default(),
             transaction_gas_cap: value.transaction_gas_cap,
+            // TODO: consider supporting `gas_estimation_mode` in scenarios to be able to compare
```
Wodann: question: Could you please create an issue to track this in GitHub and link that here?

crates/edr_provider/src/observability.rs
```diff
@@ -366,13 +373,18 @@ impl<ExecutionResultT: WithExecutionResult> ObservedExecution<ExecutionResultT>
...
+    pub fn into_result_and_filtered_traces(self) -> (ExecutionResultT, Option<CallTraceArena>) {
+        let should_include_traces = self
+            .include_call_traces
+            .should_include(|| !self.execution_result.result().is_success());
+
+        let execution_result = self.execution_result;
+        let call_trace_arena =
+            should_include_traces.then_some(self.evm_observed_data.call_trace_arena);
+
+        (execution_result, call_trace_arena)
```
Wodann: question: Why was the internal logic moved from `into_call_traces` internal implementation into the `ObservedExecution` type?

crates/edr_provider/src/data/gas.rs
```diff
@@ -41,6 +54,70 @@ pub(super) struct GasCallContext<'a, ChainSpecT: ChainSpec + HardforkChainSpec>
...
+                        &evm_observed_data.call_trace_arena,
+                        contract_decoder,
+                    ))
+                } else {
+                    Ok(gas.tx_gas_used())
+                }
+            }
+        }
+    }
+}
```
Wodann, line 98: question: Did you consider implementing this as a trait to make it extensible using runtime polymorphism?

```rs
trait GasEstimator {
  type HaltReason;

  fn accepts(..)

  fn accepted_gas_used(...)
}
```

Wodann (approved):
Overall LGTM! Just some questions, minor suggestions, and nits.

Feel free to apply them at will and merge afterwards.

crates/edr_provider/src/data/gas.rs
anaPerezGhiglia: I'd considered this, but wanting `probe_gas_limit` and `measure_gas_with_full_limit` to stay symmetric pushed me toward a single generic `ProbeResult<T>`, and since its `value: T` field documented nothing, I dropped it instead of stalling the PR.

Your feedback made me realize symmetry only needs consistent treatment, not a shared type. So I ended up adding two structs: `GasProbe { accepted, call_trace }` and `GasMeasurement { gas_used, call_trace }` ([387fe72](https://github.com/NomicFoundation/edr/pull/1471/commits/387fe72f11dfcb5ac1fbd6184eabe995f8ea0bc4)). 
Named fields self-document, and `accepted` matches the `accepts`/`accepted_gas_used` vocabulary on `GasEstimationMode` .

Thanks for the nudge! 🙇

crates/edr_scenarios/src/lib.rs
anaPerezGhiglia: #1480 created and comment updated by [0e5f729](https://github.com/NomicFoundation/edr/pull/1471/commits/0e5f7298bf22040fe3f55919045d8aac91376082)

crates/edr_napi/src/config.rs
anaPerezGhiglia: addressed by [2dfead0](https://github.com/NomicFoundation/edr/pull/1471/commits/2dfead0957b7bfbdc0ff23911024feda09463644)

crates/edr_provider/src/observability.rs
anaPerezGhiglia: Good catch! No good reason. It's the leftover of some back-and-forth refactoring. 
I rolled it back to use the EvmObservedData::into_call_traces instead in [1195bcd](https://github.com/NomicFoundation/edr/pull/1471/commits/1195bcddc4e3517f35e73188dd707852a3d53c5f)

crates/edr_provider/src/data/gas.rs
anaPerezGhiglia: No, I originally did not consider implementing this as a trait. Considering it now, I don't see what we'd gain from it to justify the extra complexity.

IIUC there are two ways the trait could play out, and IMHO both add complexity without a present benefit:

1. **Replace the provider-layer enum with a `Box<dyn GasEstimator>`** by mapping the NAPI enum directly to the estimator. Besides the dynamic dispatch (heap + vtable), this doesn't play nicely with `ProviderConfig` being `Clone + Debug` and round-tripping through the serde-able `ScenarioProviderConfig`.

2. **Keep the enum for config/serde and only build the `dyn GasEstimator` at `estimate_gas` time.** This avoids the serde problem, but then we keep the enum *and* add the trait + per-mode impls on top: strictly more layers, still with runtime dispatch, for flexibility that we don't need today.

The enum, on the other hand, gives us compile-time exhaustiveness checks, a free `Copy` implementation, and simpler config machinery (`Clone`/`Debug`/serde all derived).

So by the KISS principle I'd just leave it as an enum for now. Please let me know if there's something I'm missing, happy to read other points of view.

PR #1484, opened 2026-06-17
refactor: reusable cancellable thread synchronisation component

Wodann:
Turns the synchronisation pattern used by the `AsyncDeallocator` into a reusable building block.

### DIFF
--- a/crates/edr_napi/src/async_deallocator.rs
+++ b/crates/edr_napi/src/async_deallocator.rs
@@ -1,6 +1,7 @@
 use std::io;
 
 use crossbeam_channel::{select_biased, unbounded, SendError, Sender};
+use derive_where::derive_where;
 use edr_utils_sync::CancellableThread;
 use napi::tokio::runtime;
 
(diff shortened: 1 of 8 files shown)


crates/edr_napi/src/async_deallocator.rs
```diff
@@ -74,38 +60,16 @@ impl<T: Send + 'static> AsyncDeallocator<T> {
...
-            tracing::error!("AsyncDeallocator thread panicked: {error:?}");
-        }
-    }
-}
-
 /// A cheap, cloneable handle for sending values to an [`AsyncDeallocator`].
 ///
 /// [`Self::deallocate`] is infallible from the caller's perspective: if the
 /// dedicated thread is unavailable, it falls back to the tokio blocking pool.
+#[derive(Clone)]
```
popescuoctavian: I believe the `derive` adds an unnecessary `Clone` constraint on `T`. In the previous manual implementation of `Clone` it is not present and I don't think it should be required: `Sender<T>` implements `Clone` for any `T`.

In the current use cases this is not an issue, since `T = Arc<dyn SyncProvider>` is always `Clone`, but since this is meant to be a reusable component we should keep it as non-restrictive as possible.

popescuoctavian (approved):
LGTM overall, aside from the `derive` note. Pre-approving this to make merging easier. Thanks!

crates/edr_napi/src/async_deallocator.rs
Wodann: Good catch; thanks!

Addressed by https://github.com/NomicFoundation/edr/pull/1484/changes/623f11d67fb665a0769cb10ebd8e0ded55876492

PR #1487, opened 2026-06-18
fix: broken CI check for generated OP chain types

Wodann:
https://github.com/rust-lang/git2-rs/pull/1168 changed the `default-features`. When we adopted a new version of `git2` in https://github.com/NomicFoundation/edr/pull/1483, we didn't update the included features, breaking [this CI workflow](https://github.com/NomicFoundation/edr/actions/runs/27761633928).

By adding the `"https"` feature, the issue should be resolved.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4728,6 +4728,9 @@ dependencies = [
  "libc",
  "libgit2-sys",
  "log",
+ "openssl-probe",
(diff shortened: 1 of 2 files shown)


anaPerezGhiglia (approved):
Thanks for fixing this!

PR #1488, opened 2026-06-18
[DO NOT MERGE] ci: adopt Hardhat 3 E2E regression benchmarks

Wodann:
Test PR to validate new CI workflow:

- [x] A new workflow added from a fork does not run in [CI checks](https://github.com/NomicFoundation/edr/pull/1488/checks)
   - [x] As a result, `/bench` comments don't work
- [ ]


Wodann:
Testing for the `/bench` command is complete. This PR is no longer necessary

PR #1503, opened 2026-06-23
fix: use temporary path per workflow run

Wodann:
The HH3 regression benchmark now runs to completion but fails to update EDR versions for scenarios that have a stale checkout with conflicting semver version numbers.

This PR fixes that by always using a unique temporary path per workflow run.

Having a deterministic temp directory — in the past — allowed us to investigate the intermediate `regression-output.json` file on disk if there was a failure. Using a unique temp path prohibits this, so on failure we write it to the CI log instead now.

See failing CI job: https://github.com/NomicFoundation/edr/actions/runs/27950141616/job/82769063058

### DIFF
--- a/.github/workflows/hh3-regression-benchmark.yml
+++ b/.github/workflows/hh3-regression-benchmark.yml
@@ -85,8 +85,6 @@ jobs:
     env:
       # Hardhat is checked out into ./hardhat; the EDR repo is the workspace root.
       HH: ${{ github.workspace }}/hardhat
-      # E2E_CLONE_DIR is derived from the runner's temp dir in the "Configure E2E
-      # clone directory" step below (runner.temp is unavailable in job-level env).
     steps:
       - name: Checkout EDR
         uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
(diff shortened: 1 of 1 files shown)


Wodann:
This PR doesn't alter the actual code, so the JS Solidity test runner benchmark regression is a false positive: https://github.com/NomicFoundation/edr/actions/runs/27996265392/job/82862688696?pr=1503

.github/workflows/hh3-regression-benchmark.yml
```diff
@@ -85,12 +85,8 @@ jobs:
...
       # Hardhat is checked out into ./hardhat; the EDR repo is the workspace root.
       HH: ${{ github.workspace }}/hardhat
-      # We use a clone dir that's different than Hardhat's CI to avoid overwriting
-      # their CI files and allow inspection before the next workflow run.
-      #
-      # The clone dir is deliberately outside of the workspace to avoid package
-      # managers of the scenarios from picking up the workspace's configuration.
-      E2E_CLONE_DIR: /tmp/edr-e2e-clones
+      # E2E_CLONE_DIR is derived from the runner's temp dir in the "Configure E2E
+      # clone directory" step below (runner.temp is unavailable in job-level env).
```
anaPerezGhiglia: I think this comment in unnecessary

anaPerezGhiglia (approved):

.github/workflows/hh3-regression-benchmark.yml
Wodann: Addressed by https://github.com/NomicFoundation/edr/pull/1503/changes/98cbde55338d71a129f984046050fc143449bc72

PR #1506, opened 2026-06-23
ci: fix cancelled queue for HH3 regression benchmark

Wodann:
A queued workflow run for the HH3 regression benchmark was cancelled because a new commit was made to `main` with the same concurrency group.

This change ensures each commit on `main` has its own concurrency group, ensuring that we get one benchmark entry per commit.

[Failed workflow](https://github.com/NomicFoundation/edr/actions/runs/28033850402)

As an aside, I fixed the `Cargo.lock` file as a prior merged PR broke the `Cargo.lock` file. To prevent this from happening in the future, I asked Bas to make the check for `edr_napi` typings being up-to-date required.

### DIFF
--- a/.github/workflows/hh3-regression-benchmark.yml
+++ b/.github/workflows/hh3-regression-benchmark.yml
@@ -25,14 +25,20 @@ on:
     types: [created]
 
 concurrency:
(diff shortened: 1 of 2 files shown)


anaPerezGhiglia (approved):
LGTM

PR #1509, opened 2026-06-24
feat: add support for negative test filtering in Solidity tests

### LINKED ISSUE #1491: Support negative test name filtering
Currently EDR's Solidity Test runner exposes `testPattern` as a config option, which takes a regex.

To support negative filtering of test names across Hardhat's test plugins we should add a `excludeTestPattern` (or some other name). The semantics would allow both `testPattern` and `excludeTestPattern` to be combined, tests matching `testPattern` minus those excluded by `excludeTestPattern`.

This will allow Hardhat to implement: `pnpm hardhat test --grep token --grep-exclude badcoin` (https://github.com/NomicFoundation/hardhat/issues/8184)

popescuoctavian:
This PR adds a new `excludeTestPattern` option to the Solidity test runner, which allows negative test name filtering. The exclusion pattern works both separately and in combination with the positive filtering from `testPattern`. In cases where both are specified, they are applied in `testPattern` -> `excludeTestPattern` order.

### DIFF
--- a/.changeset/sparkly-donkeys-worry.md
+++ b/.changeset/sparkly-donkeys-worry.md
@@ -1,0 +1,5 @@
+---
+"@nomicfoundation/edr": minor
+---
+
+Added a new `excludeTestPattern` option to the Solidity test runner config for negative test name filtering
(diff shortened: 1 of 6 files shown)


Wodann (approved):
LGTM!

PR #1510, opened 2026-06-24
ci: comment on completion of `/bench` regression run

Wodann:
When a HH3 regression benchmark run is triggered by `/bench`, there was no easy way to find the results. Instead you'd have to go to the "Actions" tab and find it manually.

This PR makes that easier for `/bench`-triggered runs by:
1. Adding a link to the "🚀 Starting regression benchmark for" message
2. Posting a comment upon completion of the regression benchmark, notifying about success/failure incl. a link.

### DIFF
--- a/.github/workflows/hh3-regression-benchmark.yml
+++ b/.github/workflows/hh3-regression-benchmark.yml
@@ -355,7 +355,7 @@ jobs:
     if: >-
       !cancelled()
         && github.event_name == 'issue_comment'
-        && needs.setup.outputs.should_run == 'true'
+        && needs.regression-benchmark.result != 'skipped'
     runs-on: ubuntu-latest
     permissions:
       issues: write # post the result comment on the PR
--- a/.github/workflows/hh3-regression-benchmark.yml
+++ b/.github/workflows/hh3-regression-benchmark.yml
@@ -344,3 +344,46 @@ jobs:
         # this manually to avoid polluting future runs.
         if: always()
         run: rm -rf "$RUNNER_TEMP/edr-e2e-clones" "$RUNNER_TEMP/pnpm-metadata-cache"
+
+  report:
+    name: Report result to PR
+    needs: [setup, regression-benchmark]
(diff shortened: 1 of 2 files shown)


.github/workflows/hh3-regression-benchmark.yml
```diff
@@ -336,3 +336,46 @@ jobs:
...
+    name: Report result to PR
+    needs: [setup, regression-benchmark]
+    # 1. run on benchmark failure but skip when the run is superseded (concurrency)
+    # 2. Only report back to the PR for `/bench` comment triggers
+    # 3. Only report back if the benchmark actually started
+    # prettier-ignore
+    if: >-
+      !cancelled()
+        && github.event_name == 'issue_comment'
+        && needs.setup.outputs.should_run == 'true'
```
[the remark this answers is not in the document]
Wodann: Hallucination:

> A default status check of success() is applied unless you include one of these functions.
> 
> [..]
> 
> if: ${{ cancelled() }}

source: "A default status check of success() is applied unless you include one of these functions."

anaPerezGhiglia: `needs.setup.outputs.should_run == 'true'` is a fragile way of making sure this job  doesn't execute when the regression-benchmark job is skipped, because it only works because it to duplicates that job's own if. It'd be really easy to forget updating it here and have the two drift apart.

I think gating on the benchmark's actual result would represent better what you're claiming in the comments, and would make it more robust to maintain:
`needs.regression-benchmark.result != 'skipped'`

GHA [docs](https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#needs-context)

anaPerezGhiglia (approved):
I left a suggestion for improvement, but I'm approving it since it's not blocking

.github/workflows/hh3-regression-benchmark.yml
Wodann: Good idea! Addressed by https://github.com/NomicFoundation/edr/pull/1510/changes/f6c418d9887a004bdea44f9b0004baa799f4727b

PR #1526, opened 2026-07-01
build: upgrade to pnpm 11

nebasuke:
Builds on #1393 (now merged). With Node 20 dropped and CI on Node 24, pnpm 11's `Node >= 22.13` requirement is satisfied, so we can move off the pinned pnpm 10.

I've left the `pnpm/action-setup` bump, as I will need access to the `edr-benchmark-runner` to debug/fix that.

# Claude summary

## Changes
- `packageManager` + the `pnpm@X.Y.Z` workflow pins: `10.34.3` → `11.9.0`
- `engines`: `node >=22.13` (pnpm 11's real floor), `pnpm >=11`
- Moved `overrides` + `patchedDependencies` from `package.json`'s `pnpm` field into `pnpm-workspace.yaml` — pnpm 11 no longer reads that field (the one genuine breaking change).
- Declared `allowBuilds: false` for `esbuild`/`keccak`/`secp256k1`, matching our `ignore-scripts=true` posture.
- Release workflow: `pnpm artifacts` and `pnpm exec tsc` run right after `prepublish.sh` *intentionally* desyncs `edr_napi/package.json` from the lockfile. pnpm 11 defaults `verifyDepsBeforeRun` to `install`, which would trigger a frozen install and fail there — so those two commands pass `--config.verify-deps-before-run=false`.
- Dropped the now-obsolete pnpm-major Renovate gate.

## Follow-up: `pnpm/action-setup` v5 → v6 (deferred)

Intentionally **not** bumped here — kept the Renovate `<6` cap. action-setup reads the pnpm version from `packageManager` regardless of the action version, so v5 installs pnpm 11.9.0 fine; the bump isn't required for this upgrade.

v6's self-installer bootstraps pnpm via the runner's system `npm`, which is **corrupted on the self-hosted `edr-benchmark-runner`** (`externals/node24/bin/npm` → `Cannot find module '../lib/cli.js'`; node itself is fine). v5's install path doesn't touch that npm, so it's unaffected. v6 is confirmed working everywhere the npm is healthy — all GitHub-hosted runners, and the self-hosted regression runner (`hardhat-linux-amd64-self-hosted`).

To adopt v6 later: repair the `edr-benchmark-runner`'s npm (delete the stray `node_modules` / re-extract the node24 externals), then move the three `action-setup` pins to v6 and lift the Renovate cap. That broken npm will bite any node24 action that shells to npm on that runner, so it's worth fixing regardless.

### DIFF
--- a/.github/workflows/edr-npm-release.yml
+++ b/.github/workflows/edr-npm-release.yml
@@ -164,7 +164,7 @@ jobs:
             curl -fsSL https://sh.rustup.rs | sh -s -- -y --profile minimal --default-toolchain stable
             export PATH="/root/.cargo/bin:$PATH"
 
-            npm i -g pnpm@10.34.3 --ignore-scripts
+            npm i -g pnpm@11.9.0 --ignore-scripts
             pnpm -v
             npm i -g sfw
             sfw pnpm install --frozen-lockfile --prefer-offline
@@ -191,7 +191,7 @@ jobs:
             curl -fsSL https://sh.rustup.rs | sh -s -- -y --profile minimal --default-toolchain stable
             export PATH="/root/.cargo/bin:$PATH"
 
-            npm i -g pnpm@10.34.3 --ignore-scripts
+            npm i -g pnpm@11.9.0 --ignore-scripts
             pnpm -v
             pnpm install --frozen-lockfile --prefer-offline
             pnpm run build --strip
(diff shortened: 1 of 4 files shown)


anaPerezGhiglia (approved):
Thanks for tackling this! 🙇

PR #1527, opened 2026-07-01
build: extend typescript-eslint recommended-type-checked preset

nebasuke:
Updated the eslint rules to be based off the `recommended-type-checked` preset, as that seems to be the new standard. I disabled whichever didn't make sense, and performed some automatic/manual fixes to adhere to the new lints. 

# Claude Summary
## What this does

Replaces the hand-maintained rule allowlist with the upstream **`recommended-type-checked` preset** as the baseline, so future typescript-eslint releases surface new high-value rules automatically instead of requiring a manual allowlist edit each release. The config now carries only our deltas on top of the preset.

*(Previously scoped as a manual cherry-pick of individual rules on top of #1524; pivoted to extending the preset directly. #1393 and #1524 have since merged, so this is now based directly on `main`.)*

## The `any` decision

Extending `strict-type-checked` outright produces **~2,500 violations**, ~2,237 from the `no-unsafe-*` / `no-explicit-any` family alone — this repo's test/RPC/benchmark code is pervasively `any` (dynamic JSON-RPC payloads, fixtures). So the preset is extended but that whole family is turned **off**; enforcing it would mean thousands of low-value assertions.

## Deltas kept on top of the preset

- **Reconfigured for real false-positives:**
  - `ban-ts-comment` → allow `@ts-ignore` *with description* (napi suites need it for feature-gated `op`/`mock` exports absent only in the `testNoBuild` build, where `@ts-expect-error` would report an unused directive).
  - `only-throw-error` → allow chai's `AssertionError` (its type doesn't extend `Error`).
- **Kept explicit:** the stricter/stylistic rules the preset omits, and preset rules we run with custom options.
- **Dropped as redundant:** explicit rules the preset now provides at the same severity (`await-thenable`, `no-empty-object-type`, `no-floating-promises`, `no-misused-new`, `no-namespace`, `no-this-alias`, `no-unused-expressions`, `prefer-namespace-keyword`, `restrict-plus-operands`) plus core `no-throw-literal` (superseded by the type-aware `only-throw-error`).

## Findings the preset surfaced (all fixed)

| Rule | count | Fix |
|---|---:|---|
| `no-unnecessary-type-assertion` | 23 | pure `eslint --fix` |
| `require-await` | 20 | drop `async` where no async work; napi coverage/gasReport callbacks keep `async` (generated `=> Promise<void>` signature) with an inline-disable |
| `no-misused-promises` | 3 | ethers v6 `.on()` returns a Promise — wrap the `new Promise` executor body and `void` the call |
| `no-require-imports` | 3 | genuine runtime requires — inline-disable each with reason |
| `no-base-to-string` | 1 | `JSON.stringify` the `ForgeTestKind` before interpolating |
| `no-unsafe-enum-comparison` | 1 | compare against `TestStatus.Success`, not the raw `"Success"` |
| `prefer-promise-reject-errors` | 1 | reject with an `Error` |
| `no-unused-vars` | 2 | fallout from `--fix` removing sole type-import uses |

## Notes

No consumer-facing change (dev tooling + benchmark/test tweaks) → no changeset. `pnpm run lint` clean, packages typecheck.

### DIFF
--- a/config/eslint/eslintrc.js
+++ b/config/eslint/eslintrc.js
@@ -4,10 +4,27 @@ module.exports = {
     es6: true,
     node: true,
   },
-  extends: ["plugin:prettier/recommended"],
+  extends: [
+    "plugin:@typescript-eslint/recommended-type-checked",
+    "plugin:prettier/recommended",
+  ],
   parser: "@typescript-eslint/parser",
   plugins: ["import", "mocha", "@typescript-eslint"],
   rules: {
+    // The `recommended-type-checked` preset (extended above) is the baseline.
+    // Everything below is our delta on top of it:
+    //   1. the `no-unsafe-*` / `no-explicit-any` family is turned OFF — this
+    //      repo's test/RPC code is pervasively `any` (dynamic JSON payloads),
+    //      so enforcing it would mean thousands of low-value assertions;
+    //   2. the remaining entries are stricter or stylistic rules the preset
+    //      does not include, or preset rules we run with non-default options.
+    "@typescript-eslint/no-explicit-any": "off",
+    "@typescript-eslint/no-unsafe-argument": "off",
+    "@typescript-eslint/no-unsafe-assignment": "off",
+    "@typescript-eslint/no-unsafe-call": "off",
+    "@typescript-eslint/no-unsafe-member-access": "off",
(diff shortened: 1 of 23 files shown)


popescuoctavian (approved):
LGTM, thanks for updating this!

PR #1530, opened 2026-07-02
feat(solidity-tests): inline config parsing using Slang v2

popescuoctavian:
This PR adds native inline config parsing to EDR using Slang v2 API.

**Note:** Contract-level inline config support will be covered by a subsequent PR.

**Note for reviewer:** I believe the code should be reviewed as a whole, rather than per-commit, since the last 2 commits introduce a lot of changes.

# Claude Summary

## Summary

Add Solidity test inline-configuration parsing (`forge-config:` / `hardhat-config:`
NatSpec directives, e.g. `/// forge-config: default.fuzz.runs = 100`), parsing test sources 
directly with Slang v2. EDR now derives the per-test overrides itself instead of receiving 
them pre-built over NAPI.

## Behavior

- Inline config is collected once, up front, before any test runs.
- Ill-formed inline configuration **fails the whole run** up front (matching
  Hardhat/Foundry), rather than silently dropping config or failing a single suite.
  One problem is reported per affected test function.
- The rejected `runSolidityTests` promise carries the structured, located problems as
  an `inlineConfigErrors` array on the thrown error (source name, contract, function,
  line, message).

## NAPI API changes (breaking)

- **Removed** the `testFunctionOverrides` argument of `SolidityTestRunnerConfigArgs`
  and its associated types: `TestFunctionOverride`, `TestFunctionIdentifier`,
  `TestFunctionConfigOverride`, `FuzzConfigOverride`, `InvariantConfigOverride`,
  `TimeoutConfig`.
- **Added** `testSourcePaths` (solc source name → absolute path) and `importMappings`
  (non-relative import path → absolute path) so EDR can read each test source and
  resolve its imports from disk.

## Implementation

- New `edr_solidity_tests::inline_config` module: Slang v2 `CompilationBuilder` parses
  each test source with imports resolved from disk; per-function overrides are extracted
  from the leading NatSpec and cached by source name.
- The import resolver is built in `TryFrom<TestRunnerConfig>` and carried on
  `SolidityTestRunnerConfig`; `MultiContractRunner::new` runs collection (off the async
  runtime) and aborts the run on any malformed directive.

## Testing

- Rust unit + `it` integration tests (override behavior + malformed-config abort).
- JS integration tests and the NAPI mocha suite.


crates/edr_napi/src/solidity_tests/config.rs
```diff
@@ -332,22 +358,9 @@ impl SolidityTestRunnerConfigArgs<'_> {
...
-                                .unwrap_or_default()
-                        })
-                        .map(|override_item| override_item.identifier.clone().try_into())
-                        .collect::<Result<_, _>>()
-                })
-                .transpose()?
-                .unwrap_or_default(),
+            // Per-function `allowInternalExpectRevert` comes from inline config, applied per-suite
+            // by the test runner.
+            functions_internal_expect_revert: std::collections::HashSet::default(),
```
Wodann: Question: Is there a reason this member field cannot be removed from the struct?

crates/edr_napi/src/solidity_tests/inline_config.rs
```diff
@@ -0,0 +1,231 @@
...
+    /// `project/test/Foo.t.sol`).
+    pub source_name: String,
+    /// The contract the offending directive belongs to, if known.
+    pub contract: Option<String>,
+    /// The test function the offending directive belongs to, if known.
+    pub function: Option<String>,
+    /// The 1-based line of the offending directive within the source, if known.
+    pub line: Option<u32>,
+    /// The problem itself; discriminate on its `kind` tag.
+    pub problem: Either8<
```
Wodann: Question: I thought this problem had been resolved in N-API v3. Did you still run into issues using the type export `InlineConfigProblem`?

crates/edr_napi/src/solidity_tests/inline_config.rs
```diff
@@ -0,0 +1,231 @@
...
+            contract: item.contract.clone(),
+            function: item.function.clone(),
+            line: item.line,
+            problem: to_problem(&item.error),
+        }
+    }
+}
+
+/// Builds the error to reject `runSolidityTests` with when inline-config
+/// validation fails, carrying the structured, located problems on the JS error
```
Wodann: typo: "with when" seems to be missing part of the sentence.

crates/edr_napi/src/solidity_tests/inline_config.rs
```diff
@@ -0,0 +1,231 @@
...
+/// as its `inlineConfigErrors` property.
+///
+/// Must be called on the JS thread (it builds JS values); the reject path in
+/// [`crate::context`] does so from the deferred's resolver. Falls back to a
+/// plain message-only error if building the structured object fails.
+pub(crate) fn to_napi_error(env: &Env, errors: &InlineConfigErrors) -> napi::Error {
```
Wodann, line 280: Thought: There is a trait `ToNapiValue` that you can use to manually implement conversions from a Rust type to an N-API type. It receive exactly this interface. Would that potentially be a more idiomatic way of returning a custom error object?

crates/edr_napi/src/context.rs
```diff
@@ -309,42 +318,67 @@ impl EdrContext {
...
+
+            let outcome = match create_result {
+                // An inline-config failure carries structured, located problems
+                // that we surface on the rejected promise's error as
+                // `inlineConfigErrors`. Building that JS object requires the JS
+                // thread, so route it through the deferred's resolver (which
+                // runs there) rather than `deferred.reject`, which only carries
+                // a message.
+                Err(solidity::CreateTestRunnerError::InvalidInlineConfig(errors)) => {
+                    RunOutcome::InvalidInlineConfig(errors)
```
Wodann, line 331: suggestion: Instead of introducing a `RunOutcome` type, can you call `deferred.resolve(|env| ..)` here and `return`?

I think it keeps the code a little simple to parse.

crates/edr_napi/index.d.ts
```diff
@@ -1636,6 +1693,25 @@ export interface SolidityTestRunnerConfigArgs {
...
   eip712CanonicalTypes?: Array<string>
+  /**
+   * Maps the solc source names of the test-suite sources (e.g.
+   * `project/test/Foo.t.sol`) to their absolute paths on disk.
+   *
+   * Used to parse inline test configuration (`forge-config:`/
+   * `hardhat-config:` NatSpec directives) directly from the sources. A test
+   * source without an entry has no inline configuration collected.
+   */
+  testSourcePaths?: Record<string, string>
```
Wodann, line 1762: Thought: The test suites are received as `ArtifactId`s. Did you explore receiving absolute paths in that struct instead of using this mapping? Were there any blockers?

crates/edr_solidity_tests/src/inline_config/directives.rs
```diff
@@ -0,0 +1,664 @@
...
+    "fuzz",
+    "invariant",
+    "allowInternalExpectRevert",
+    "isolate",
+    "evmVersion",
+];
+
+/// Inline-config profiles the parser accepts as a leading dot-segment prefix.
+/// Only `default` is supported today; add new profiles here to extend support.
+const SUPPORTED_PROFILES: [&str; 1] = ["default"];
```
Wodann, line 54: Question: In Foundry, can users create profiles with custom names?

crates/edr_solidity_tests/src/inline_config/directives.rs
```diff
@@ -0,0 +1,664 @@
...
+/// (`max-test-rejects` and `max_test_rejects` both become `maxTestRejects`).
+/// A trailing delimiter with no following character is left untouched to avoid
+/// silently fixing malformed keys.
+fn delimiter_to_camel(input: &str, delim: char) -> String {
+    let mut out = String::with_capacity(input.len());
+    let mut chars = input.chars();
+    while let Some(c) = chars.next() {
+        if c == delim {
+            match chars.next() {
+                Some(next) => out.extend(next.to_uppercase()),
```
Wodann, line 234: nit: If we are only `extend`ing by one character, I'd prefer using `.push`. `push` exclusively receives a single character, whereas `extend` can receive multiple characters. The narrower constrained thus more clearly expresses the intent to the reader, imo.

crates/edr_solidity_tests/src/inline_config/directives.rs
```diff
@@ -0,0 +1,664 @@
...
+    let strip_decoration = |line: &str| -> String {
+        let line = line.trim_start();
+        let line = line.strip_prefix("///").unwrap_or(line);
+        let line = line.strip_prefix("/**").unwrap_or(line);
+        let line = line.strip_suffix("*/").unwrap_or(line);
+        let line = line.trim_start();
+        let line = line.strip_prefix('*').unwrap_or(line);
+        // A block comment's closing `*/` can trail a directive on its own line.
+        let line = line.strip_suffix("*/").unwrap_or(line);
+        line.trim().to_owned()
```
Wodann: Question: Is there a reason you need an owned `String` rather than continuing to use the `str` slice?

crates/edr_solidity_tests/src/inline_config/directives.rs
```diff
@@ -0,0 +1,664 @@
...
+    #[test]
+    fn timeout_parsed_as_seconds() {
+        let cfg = parse("/// forge-config: fuzz.timeout = 30", "testFoo")
+            .unwrap()
+            .unwrap();
+        assert_eq!(
+            cfg.fuzz.unwrap().timeout,
+            Some(TimeoutConfig { time: Some(30) })
+        );
+    }
```
Wodann, line 663: praise: nice parsing implementation and test suite!

crates/edr_solidity_tests/src/inline_config/error.rs
```diff
@@ -0,0 +1,167 @@
...
+}
+
+/// A single inline-config problem together with enough location to point the
+/// user at it — modeled on the stack-trace `SourceReference` surfaced to
+/// consumers.
+///
+/// [`contract`](Self::contract), [`function`](Self::function) and
+/// [`line`](Self::line) are absent for a source-level problem (e.g. an
+/// unsupported solc version or an unreadable source file), where there is no
+/// single directive to point at.
```
Wodann: nit: It feels as though this could be more efficiently represented if the source-level problems don't carry the `None` for `contract`, `function`, and `line` at all and those fields are moved one level down to be more specific.

crates/edr_solidity_tests/src/inline_config.rs
```diff
@@ -0,0 +1,71 @@
+//! Parses inline configuration for Solidity tests from NatSpec comments.
```
Wodann, line 1: nit: we don't use `mod.rs` in EDR crates, instead opting for the `src/inline_config.rs` style of defining new modules. This makes it easier to lookup files by name in IDEs (e.g. Ctrl+P, "inline_config" in VSCode)

(The only exception to this is in `crates/foundry` if I'm not mistaken.)

crates/edr_solidity_tests/src/inline_config/natspec.rs
```diff
@@ -0,0 +1,203 @@
...
+    loop {
+        // Skip whitespace between the node/previous block and the next comment.
+        let head = region.get(..end).unwrap_or("");
+        let trimmed = head.trim_end();
+        end = trimmed.len();
+        if end == 0 {
+            break;
+        }
+
+        if trimmed.ends_with("*/") {
```
Wodann, line 69: Question: Is it valid natspec to have consecutive `/** */` blocks or mix `/** */` blocks with `///` blocks?

The current code seems to allow this.

crates/edr_solidity_tests/src/inline_config/overrides.rs
```diff
@@ -0,0 +1,180 @@
...
+/// The 1-based line number of `offset` within `source`.
+fn line_of(source: &str, offset: usize) -> u32 {
+    let end = offset.min(source.len());
+    let newlines = source
+        .as_bytes()
+        .iter()
+        .take(end)
+        .filter(|&&byte| byte == b'\n')
+        .count();
+    u32::try_from(newlines + 1).unwrap_or(u32::MAX)
```
Wodann: suggestion: Instead of silently failing, bubble up an error (if possible)

js/integration-tests/solidity-tests/test-contracts/FuzzConfigOverride.t.sol
```diff
@@ -6,6 +6,7 @@ import {Vm} from "forge-std/src/Vm.sol";
 
 contract FuzzConfigOverrideTest is Test {
```
Wodann, line 7: Thought: Now that we handle inline configs internally in EDR, do we still need the JS tests that previously tested the inline configs passed in via the `SolidityTestRunnerConfig`; other than to test the structured errors that are generated?

popescuoctavian:
/bench hardhat-ref=popescuoctavian/remove-inline-config scenarios=1inch* benchmarks="test solidity"

Wodann (commented):
This is looking good. I like the way the code is structured in the `inline_config` directory. I do have some questions to clarify my understanding, primarily about the API layer.

popescuoctavian:
/bench hardhat-ref=popescuoctavian/remove-inline-config scenarios=1inch* benchmarks="test solidity"

crates/edr_napi/src/context.rs
popescuoctavian: This would result in two different closure types at two resolve sites, which is not allowed. `JsDeferred<Data, Resolver>` bakes the concrete closure type into the deferred's type, so a single deferred value accepts exactly one closure type. This was the reason I went with a separate `RunOutcome` type.

crates/edr_napi/src/solidity_tests/inline_config.rs
popescuoctavian: I don't think that trait helps here since the reject path requires an `napi::Error`, not a value, so it would still need to be wrapped in an error, similar to the `Ok(napi::Error::from(error_object.to_unknown()))` line below.

Additionally, to preserve the same error format, it would require introducing a new type (something like `InlineConfigErrorReport<'a>(&'a InlineConfigErrors)` and implementing `to_napi_value` which is an `unsafe fn`.

crates/edr_solidity_tests/src/inline_config/directives.rs
popescuoctavian: Yes, but Hardhat only supports `default` right now. There is an issue that I've created in Hardhat which is tracking this https://github.com/NomicFoundation/hardhat/issues/8145.

crates/edr_solidity_tests/src/inline_config/directives.rs
popescuoctavian: Using `.extend()` here avoids the need to handle panics: some non-ASCII characters can produce more than one character when uppercased (for example 'ß'.to_uppercase() => "SS"). Of course this isn't likely to happen here, but I believe that using `extend` keeps the code cleaner. Not a strong opinion though, so let me know if you want me to change it.

js/integration-tests/solidity-tests/test-contracts/FuzzConfigOverride.t.sol
popescuoctavian: I agree, most of the JS tests can be removed - we can just keep the error checks and one of each kind to make sure the NAPI wiring still applies inline configs correctly. However, before that, I need to have a look to make sure we have full coverage for existing test cases on the Rust side and move some tests there (at a quick glance `allowInternalExpectRevert` doesn't). So I'll attempt to refactor this and clean it up in a follow-up PR.

crates/edr_napi/src/solidity_tests/config.rs
popescuoctavian: I've refactored this and moved it directly to `CheatsConfig` in 03119c4d6cc690a829077feb19dd75d9ee4b0525. The data still needs to be kept there since `allowInternalExpectRevert` is applied at the inspector level.

crates/edr_napi/src/solidity_tests/inline_config.rs
popescuoctavian: I believe this code was written before NAPI v3 support was merged 😄. Addressed by c77754afd1a0ffe1fb360abe470fa5e22c3fd930.

crates/edr_napi/src/solidity_tests/inline_config.rs
popescuoctavian: Not a typo, but I agree it was odd wording: "with" refers to the error which `runSolidityTests` is rejected "with" when ... 😅. Addressed by 4b6a22bacc8cc652e1606b77047805170359b600.

crates/edr_solidity_tests/src/inline_config/directives.rs
popescuoctavian: Addressed by e588011e2df7a6a16baf3a94dda4367afd6adffc.

crates/edr_solidity_tests/src/inline_config/error.rs
popescuoctavian: Good idea, addressed by 722716e38be2b679a0e822d685003dbdaa07b0d9.

crates/edr_solidity_tests/src/inline_config.rs
popescuoctavian: Addressed by 416a837978b0e8cb4e04f83e7f6bf6694c82f8b1.

crates/edr_solidity_tests/src/inline_config/natspec.rs
popescuoctavian: Hardhat (solc build info) doesn't allow this right now (it picks the block/line closest to the function definition when any regular comment or blank line splits doc comments). However, Foundry now uses Solar to parse inline configs, which allows this. I went with the largest coverage (Foundry) here to maximize compatibility, but failed to document it. I've added documentation and tests for these cases in 013839518b304f65c2fa7186cad76cc4a349afb9.

crates/edr_solidity_tests/src/inline_config/overrides.rs
popescuoctavian: Addressed by adf6c0ec1f3af6946d1eb81dc09d3d2c89ae8b39.

crates/edr_napi/index.d.ts
popescuoctavian: I did consider this, but ended up not implementing it. The main reason was that this caused changes in a lot of places: [ `context.rs`](https://github.com/NomicFoundation/edr/blob/2b5d1178de97b7c2af0092997fbf9fdfc0e4cf66/crates/edr_napi/src/context.rs) looks up JS-provided test-suite ids in `known_contracts` maps built from the artifacts, and ids would be constructed both with and without the path (unless Hardhat also provides absolute paths for all non-test-contracts). It also implied changes in the linker and Foundry type conversion to adopt it, or excluding the absolute path when identifying artifacts and manually implementing Eq/Hash/Ord.

I'm also not sure if it's worth the time investment, considering the changes we're planning on artifact loading. Let me know what you think.

popescuoctavian:
/bench hardhat-ref=popescuoctavian/remove-inline-config benchmarks="test solidity"

crates/edr_napi/index.d.ts
Wodann: Thanks for explaining the reasoning. I'll follow the same approach as your for EIP-712 types too then.

crates/edr_solidity_tests/src/inline_config/directives.rs
Wodann: Thanks for explaining. I didn't realise that a single char could result in multiple chars when uppercased! 😮

Wodann (approved):
Thanks for addressing the feedback.

Looks good to merge when the Hardhat team is also ready to merge their part of this change.

PR #1543, opened 2026-07-06
feat: simulate EIP-7928 blockAccessListHash header field (Amsterdam)

anaPerezGhiglia:
## Summary
Adds minimal, experimental EIP-7928 support: from the Amsterdam+ hardfork, block headers carry a `blockAccessListHash` field.

The value is simulated: EDR does not build the real block access list, so it is not the actual keccak256(rlp(blockAccessList)) from the spec. Two guarantees hold:
 - a block that introduces no state changes gets the empty-RLP-list hash, `keccak256(rlp([]))`, as the EIP specifies;
 - no two blocks in the same blockchain share a hash — a state-changing block's hash is derived deterministically from its `parent_hash`, which is unique per block within a chain.

## Changes
- `PartialHeader::new` owns the hardfork gate: the field exists only on Amsterdam+ (a pre-Amsterdam override is ignored), defaulting to the empty-list hash. This covers every locally-built header, genesis, reserved, and mined, at a single construction point.
- Block builder finalize replaces that default with the parent-hash-derived value when the block changed state, and leaves an externally-supplied value untouched.
- Added new `block_access_list_hash` to `HeaderOverrides`.
- Added a new `amsterdam_header_overrides` helper so block replay can supply the on-chain value. Currently unused, waiting for Amsterdam testnets to include a new CI check.


crates/edr_chain_l1/src/block.rs
```diff
@@ -809,3 +820,118 @@ impl<
...
+/// that produced different state.
+fn block_access_list_hash(
+    current: Option<B256>,
+    state_diff: &StateDiff,
+    state_root: B256,
+) -> Option<B256> {
+    if current == Some(KECCAK_RLP_EMPTY_ARRAY) && !state_diff.as_inner().is_empty() {
+        Some(keccak256(
+            format!("blockAccessListHash{state_root}").as_bytes(),
+        ))
```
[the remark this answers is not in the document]
anaPerezGhiglia: I don't think either of these claims justifies the change, given how small the allocated string is and considerint the value is a simulated placeholder anyway.

crates/block/header/src/lib.rs
```diff
@@ -76,6 +79,9 @@ pub struct BlockHeader {
     ///
     /// [EIP-7685]: https://eips.ethereum.org/EIPS/eip-7685
     pub requests_hash: Option<B256>,
+    /// The Keccak-256 hash of the RLP-encoded access list.
+    /// Was added by EIP-7928 and is ignored in legacy headers
```
Wodann: suggestion(non-blocking): consider adding a link to EIP
```suggestion
    /// Was added by [EIP-7928] and is ignored in legacy headers
    ///
    /// [EIP-7928](https://eips.ethereum.org/EIPS/eip-7928)
```

crates/block/header/src/lib.rs
```diff
@@ -310,6 +317,8 @@ pub struct PartialHeader {
     ///
     /// [EIP-7685]: https://eips.ethereum.org/EIPS/eip-7685
     pub requests_hash: Option<B256>,
+    /// The Keccak-256 hash of the RLP-encoded access list (EIP-7928)
+    pub block_access_list_hash: Option<B256>,
```
Wodann:
```suggestion
    /// The Keccak-256 hash of the RLP-encoded access list ([EIP-7928])
    ///
    /// [EIP-7928](https://eips.ethereum.org/EIPS/eip-7928)
    pub block_access_list_hash: Option<B256>,
```

crates/edr_chain_l1/src/rpc/block.rs
```diff
@@ -102,6 +102,11 @@ pub struct L1RpcBlock<TransactionT> {
     /// [EIP-7685]: https://eips.ethereum.org/EIPS/eip-7685
     #[serde(skip_serializing_if = "Option::is_none")]
     pub requests_hash: Option<B256>,
+    /// Keccak-256 hash of the RLP-encoded block access list
+    ///
+    /// [EIP-7928](https://eips.ethereum.org/EIPS/eip-7928)
+    #[serde(skip_serializing_if = "Option::is_none")]
```
Wodann: issue: consider adding `#[serde(default)]` to ensure that an `undefined` (i.e missing) `blockAccessListHash` field results in `None`.

This is similar to other hardfork-dependent RPC values like `excessBlobGas`

crates/edr_chain_l1/src/block.rs
```diff
@@ -623,6 +626,14 @@ impl<
                 .as_secs();
         }
 
+        // Must run after the reward loop and state-root computation above, so both
+        // `state_diff` and `state_root` are final.
```
Wodann: issue: in fork-mode `stateRoot` is pseudorandomly generated. As a result, blockchain's with different state can get the same pseudorandomly generated `stateRoot`.

For local mode the state root suffices, but for fork mode we thus need to manually account for all possible forms of variance. It might be hard (to impossible) to encode them all in a deterministic hash calculation.

After thinking about this, even more causes of variance came to mind:
- transactions; variance is encoded in the transaction root hash
- prior state; 
- hardfork; the user might use a custom hardfork activation for the current block, resulting in different execution flow.
- irregular state changes (`hardhat_setCode`, etc.) are applied on top of a block's state and are not included in the prior block's header.
- state overrides (e.g. `eth_call`); although we don't support this for all JSON-RPC methods listed in the JSON-RPC spec yet.

The parent block's block hash might suffice to encode the prior state variance, but it cannot capture the hardfork variance, nor irregular state changes, nor state overrides. And then there might be even more causes of variance that I missed.

We could add all of these sources of variance to the hash computation, but it might grow out of hand.

With the increasing list of sources of variance, it's starting to feel like we might continue to find bugs over time; and despite me liking the desired outcome of block-content-derived hashes, a more realistic solution might be to just pass a pseudorandom hash generator in during `fn finalize` and keep it simple until we decide to do a refactor that computes the actual block-level access list.

Separate from the above issue, I do agree with your design choice to use the RLP encoding of an empty array if `state_diff` is empty.

crates/edr_generic/src/rpc/block.rs
```diff
@@ -98,6 +98,11 @@ pub struct GenericRpcBlock<TransactionT> {
     /// [EIP-7685]: https://eips.ethereum.org/EIPS/eip-7685
     #[serde(skip_serializing_if = "Option::is_none")]
     pub requests_hash: Option<B256>,
+    /// Keccak-256 hash of the RLP-encoded block access list
+    ///
+    /// [EIP-7928](https://eips.ethereum.org/EIPS/eip-7928)
+    #[serde(skip_serializing_if = "Option::is_none")]
```
Wodann: issue: add `#[serde(default)]` to ensure this is `None` if missing

crates/edr_provider/tests/integration/eip7928.rs
```diff
@@ -0,0 +1,139 @@
...
+fn mine_block(provider: &Provider<L1ChainSpec>) {
+    provider
+        .handle_request(ProviderRequest::with_single(MethodInvocation::EvmMine(
+            None,
+        )))
+        .expect("evm_mine should succeed");
+}
+
+/// Sends a value-transfer transaction, mined into a block.
+fn transfer_funds(provider: &Provider<L1ChainSpec>) {
```
Wodann: thought: If I'm not mistaken, you (or Bas) added a similar function to `edr_provider::test_utils` last week.

Wodann (changes requested):

crates/edr_generic/src/rpc/block.rs
anaPerezGhiglia: addressed by [dc68bcd](https://github.com/NomicFoundation/edr/pull/1543/commits/dc68bcdf66ee9f68774bc9103f39a87ee266798f)

crates/edr_chain_l1/src/rpc/block.rs
anaPerezGhiglia: Addressed by [dc68bcd](https://github.com/NomicFoundation/edr/pull/1543/commits/dc68bcdf66ee9f68774bc9103f39a87ee266798f)

crates/edr_chain_l1/src/block.rs
anaPerezGhiglia: After thinking about this more, I went with a simpler approach that guarantees:
- two blocks in the same blockchain can't share a `block_access_list_hash`;
- a block with no state changes gets the empty-RLP-list hash, `keccak256(rlp([]))`, as the EIP specifies.

We get both by seeding the keccak with the block's `parent_hash`: within a chain the parent hash is unique per block, so sibling blocks can't collide — without having to enumerate every source of variance (transactions, prior state, hardfork, irregular state, overrides). The trade-off is that the value isn't consistent across different provider configurations (e.g. a different hardfork), which I think is fine for a simulated field.

I discarded the alternative of injecting the randomly generated `block_access_list_hash` through the `finalize_block` signature: it threads an Amsterdam-specific field through several chain-agnostic signatures (the shared BlockBuilder trait, the miner, block replay) for a value most of them don't care about. It feels out of place and hurts maintainability, without a clear needed functional gain over the parent-hash approach.

So I chose the current solution for its simplicity and small footprint. I did prototype the injection version on a separate branch, you can check it out here https://github.com/NomicFoundation/edr/compare/feat/amsterdam-eip-7928-BALs...alt/eip-7928-BALs

Wodann: suggestion: update the comment to no longer mention `state_root`.

crates/edr_chain_l1/src/block.rs
```diff
@@ -809,3 +820,114 @@ impl<
         self.finalize(rewards)
     }
 }
+
+/// Resolves a block's simulated block access list hash (EIP-7928), upholding
+/// two guarantees:
+/// - a block that introduces no state changes keeps the empty-RLP-list hash, as
+///   the EIP specifies;
+/// - no two blocks in the same chain share a hash: a state-changing block's
+///   hash is derived from its parent hash, which is unique per block.
```
Wodann, line 829: Suggestion: Maybe add a note that: "reverting to a snapshot would be considered forking of the chain, so the same block-level access list can reoccur in that fork"

What would happen is:
- mines a new block (generates EIP-7928 hash from prior block)
- reverts to previous block
- mines a new block (generates EIP-7928 hash from same prior block)

Wodann (approved):
Some minor suggestions for improving docs but otherwise looks good to me.

Feel free to merge after addressing them.

crates/edr_chain_l1/src/block.rs
anaPerezGhiglia: addressed by [180953a](https://github.com/NomicFoundation/edr/pull/1543/commits/180953ad4b83660285c9584a609dfa843dcc8ace)

crates/edr_chain_l1/src/block.rs
anaPerezGhiglia: addressed by [180953a](https://github.com/NomicFoundation/edr/pull/1543/commits/180953ad4b83660285c9584a609dfa843dcc8ace)

PR #1545, opened 2026-07-06
ci: add actionlint + zizmor GitHub Actions linters

nebasuke:
Based on successful usage for Slang, I've added `actionlint` and `zizmor` as they respectively catch syntax and security issues in GitHub actions.

# Claude summary

Adds two static-analysis gates for the repo's own GitHub Actions, which had none before:

- **actionlint** — workflow correctness (syntax, expression validity, `runs-on` labels, and shellcheck over `run:` blocks).
- **zizmor** — GitHub Actions security auditing (excessive permissions, template injection, cache poisoning, etc.).

Both run as blocking jobs in `edr-ci.yml` and fail the check on any finding.

## How they run

| Tool | Mechanism | Notes |
|---|---|---|
| actionlint | `uses: docker://rhysd/actionlint@<digest>` | rhysd's official image, digest-pinned; bundles shellcheck + pyflakes, so no runner deps. |
| zizmor | `taiki-e/install-action` with `tool: zizmor@1.26.1` | Same installer already used for `cargo-hack`. `GH_TOKEN` enables the online audits. |

Both are separate jobs with `permissions: contents: read` and no `needs:`, so they run in parallel with the rest of CI.

`.github/actionlint.yaml` declares the two self-hosted runner labels (`edr-benchmark-runner`, `hardhat-linux-amd64-self-hosted`) so `runs-on:` isn't flagged as unknown.

## Findings resolved

The first run surfaced 43 zizmor findings (plus 4 actionlint). All are resolved so both gates land green:

- **`excessive-permissions` (26)** — added explicit least-privilege `permissions:` blocks across 9 workflows. Scoped per workflow: `contents: read` for the read-only ones (benchmark push and Slack notify use PAT/webhook secrets, not `GITHUB_TOKEN`); `pull-requests: read` for `check-changeset-added`; and mdbook's `pages`/`id-token: write` moved down to the deploy job.
- **`template-injection` (7 of 9)** — routed `needs.*` outputs (`filename`, publish `tag`) through `env:` in `edr-npm-release.yml`; no behaviour change.
- **`cache-poisoning` (2 of 3)** — disabled caching on the publish path: `package-manager-cache: false` on the validation job's setup-node, and removed the cooldown-check cache step (also closes a small gap where a poisoned cache could hide a too-fresh dependency at release).
- **`github-env` (2)** — suppressed as verified-safe (a static `+crt-static` flag and `rustc --print host-tuple` output — neither is external input).

## Suppressions kept (with justifications)

- **`cache-poisoning`** on the build job's cargo cache — releases already build cold (guarded by `if: tag == ''`), so the published artifact never comes from cache; zizmor just can't statically evaluate the `needs.*`-based guard. Fixing it "for real" would drop PR build caching for zero security gain.
- **`template-injection`** on `strategy.job-total` — a controlled integer, not external input.
- **`superfluous-actions`** (×2) — `addnab/docker-run-action` is a deliberate cross-arch (arm64 / arm64-musl) test runner.
- **`adhoc-packages`** — pinned `npm@11.6.2`, required for Trusted Publishing.

## Verification

Ran both tools locally against `.github` — actionlint and zizmor both report **no findings**.

---

**Note:** CI-only change, so apply the **"no changeset needed"** label (otherwise `check-changeset-added` fails).

### DIFF
--- a/.github/actionlint.yaml
+++ b/.github/actionlint.yaml
@@ -1,6 +1,4 @@
-# Self-hosted runner labels actionlint can't know about, so it doesn't flag
-# `runs-on:` as an unknown label. Keep in sync with the runners registered for
-# this repo/org.
+# Self-hosted runner labels, so actionlint doesn't flag `runs-on:` as unknown.
 self-hosted-runner:
   labels:
     - edr-benchmark-runner
--- a/.github/actionlint.yaml
+++ b/.github/actionlint.yaml
@@ -1,0 +1,7 @@
+# Self-hosted runner labels actionlint can't know about, so it doesn't flag
+# `runs-on:` as an unknown label. Keep in sync with the runners registered for
+# this repo/org.
+self-hosted-runner:
+  labels:
+    - edr-benchmark-runner
+    - hardhat-linux-amd64-self-hosted
--- a/.github/actions/setup-llvm-cov/action.yml
+++ b/.github/actions/setup-llvm-cov/action.yml
@@ -15,4 +15,4 @@ runs:
     - name: Pin CARGO_BUILD_TARGET to the host triple
       if: inputs.builds-napi-crate == 'true'
       shell: bash
-      run: echo "CARGO_BUILD_TARGET=$(rustc --print host-tuple)" >> "$GITHUB_ENV"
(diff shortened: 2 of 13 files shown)


popescuoctavian (approved):
LGTM, thanks for adding this! The benchmark failure looks like flakiness, but I've started a re-run anyway. Pre-approving this.

PR #1546, opened 2026-07-07
fix(ci): floor the HH3 regression benchmark sentinel version at the last npm release

popescuoctavian:
This PR provides a fix for the scenario where a new HH version has been released after the initial `/bench` run. In this scenario, subsequent runs would fail because the sentinel version would collide with the newly published HH version.

### The failure chain

1. The sentinel prediction. At job start, compute-sentinel-versions.cjs computes the version the local Hardhat will be republished as: repo version + 1 patch. The checked-out Hardhat repo's packages/hardhat/package.json is at 3.9.0, so it predicted HH_VER=3.9.1. The workflow's repoint step then sets version=3.9.1 on the checkout — per its own comment, that pre-bump exists so Hardhat's harness sees the package as "already bumped" and republishes it at exactly that version.
2. Upstream moved. Hardhat released 3.9.1 to npm on 2026-07-02 (I verified: it's npm's current latest), four days before this run — while the benchmarked Hardhat checkout still says 3.9.0.
3. The harness bumped again. Hardhat's e2e publisher (regression.ts → "Detecting packages changed since release") compares against the last npm release. The other packages log "already bumped to X" (version ahead of release → keep it), but hardhat's repointed 3.9.1 now equals the real npm release 3.9.1, so it logged hardhat (changed since hardhat@3.9.1) → "Bumping patch versions: hardhat 3.9.1 → 3.9.2" and published 3.9.2.
4. Check A compares against the stale prediction: Published hardhat latest (3.9.2) != 3.9.1 → red.

### DIFF
--- a/.github/scripts/compute-sentinel-versions.cjs
+++ b/.github/scripts/compute-sentinel-versions.cjs
@@ -19,20 +19,68 @@ function edrVersion(edrBaseVersion, shortSha) {
 // harness pins each scenario's `hardhat` dependency to it, and scenarios pull
 // Hardhat plugins whose `peerDependencies` use ranges like `hardhat@^3.8.0`.
 // node-semver excludes prereleases from such ranges.
-function hardhatVersion(hardhatBaseVersion) {
-  const core = hardhatBaseVersion.split("+")[0].split("-")[0];
-  const [major, minor, patch] = core.split(".").map(Number);
-  if (![major, minor, patch].every(Number.isInteger)) {
-    throw new Error(`Unparseable Hardhat version: ${hardhatBaseVersion}`);
+//
+// It must also be strictly ahead of the *last npm release*, not just of the
+// checked-out repo's version: the harness's publish step patch-bumps any
+// workspace package whose version doesn't exceed its last release, which would
+// desync the published version from this prediction and fail the workflow's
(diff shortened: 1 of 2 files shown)


.github/scripts/compute-sentinel-versions.cjs
```diff
@@ -19,20 +19,68 @@ function edrVersion(edrBaseVersion, shortSha) {
...
+// checked-out repo's version: the harness's publish step patch-bumps any
+// workspace package whose version doesn't exceed its last release, which would
+// desync the published version from this prediction and fail the workflow's
+// "Validate scenarios used the local EDR build" step. The Hardhat checkout can
+// lag npm (its package.json only catches up when a release lands on the
+// benchmarked ref), so floor the sentinel at the published version before
+// bumping.
+function hardhatVersion(hardhatBaseVersion, lastPublishedVersion) {
+  let core = parseCore(hardhatBaseVersion);
+  if (lastPublishedVersion !== undefined) {
```
Wodann, line 33: nit: based on the control-flow, it looks as though `lastPublishedVersion` would always be defined. If so, you can drop this if statement.

Wodann (approved):
Thanks for fixing this!

PR #1547, opened 2026-07-07
fix: JSON-RPC provider call traces omitting event logs

### LINKED ISSUE #1542: Provider/RPC call traces omit event logs (`EvmObserver` never forwards `log`/`log_full`)
## Summary

On the provider/RPC path (traces produced for `eth_sendTransaction` / `eth_call` / `eth_estimateGas`, i.e. Hardhat 3 mocha/JS tests run with `-vvvv`), verbose call traces **never show emitted events**. The Solidity-test path (`hardhat test solidity`), which uses a sibling inspector, renders events correctly. Same EDR, same renderer, same trace model, events are just never collected on the provider path.

## Steps to reproduce (see code below)

1. Take any contract with a function that emits an event.
2. **Provider path:** run a mocha/JS test that calls it, with `hardhat test <test> -vvvv`. Call traces print, but there is **no `[event]` line** for the emitted event.
3. **Solidity-test path (control):** `hardhat test solidity <t.sol> -vvvv` → `[event] Name(args)` lines **do** appear.

## Likely root cause (Claude investigation in EDR)

Both paths render events the same way: `CallTrace::from_arena_node` (`crates/edr_napi/src/solidity_tests/test_results.rs:648`) turns each arena node's `logs` into `[event]` children (`:684`). Events show up only if `node.logs` is populated — and on the provider path it appears to stay empty because of two separate gaps:

1. **`EvmObserver` doesn't capture logs.** Its `Inspector` impl (`crates/edr_provider/src/observability.rs:275-334`) handles `call`, `call_end`, `create`, `create_end`, and `step`, but not `log` / `log_full`. revm's default `log` hook is a no-op, so emitted events never reach the wrapped `SolidityTracingInspector` (which does implement them at `crates/edr_solidity/src/tracing.rs:94,98`).
2. **The provider tracer isn't told to record logs.** It's built with `TracingInspectorConfig::default_parity().set_steps(true)` (`crates/edr_provider/src/observability.rs:163-166`), which leaves `record_logs` at `false`. The Solidity-test path sets `record_logs: true` (`crates/foundry/evm/traces/src/lib.rs:209`).

Both need fixing — one alone isn't enough. Without the `log` hook (gap 1), the event never reaches the inspector, so enabling `record_logs` (gap 2) on its own records nothing.

This points to an EDR-only fix; no Hardhat changes needed. Hardhat already sets `includeCallTraces: All` at `-vvvv` and already prints `[event]` for `LogKind.Log` children — that branch is just never hit today because EDR sends empty `children`.

## Expected vs actual

- **Expected:** provider-path call traces include decoded event logs as `[event]` children (as the Solidity-test path already does
- **Actual:** events are silently dropped.

Difference in output:
- Left: has event, correct
- Right: missing event, wrong

Image

---

Solidity file: EventExample.sol
```
// SPDX-License-Identifier: MIT
pragma solidity ^0.8.0;

contract EventExample {
  event ValueSet(address indexed setter, uint256 newValue);

  uint256 public value;

  function setValue(uint256 newValue) public {
    value = newValue;
    emit ValueSet(msg.sender, newValue);
  }
}
```

Solidity test: EventExample.t.sol
```
// SPDX-License-Identifier: MIT
pragma solidity ^0.8.0;

import "../../contracts/EventExample.sol";
import "forge-std/Test.sol";

contract EventExampleTest is Test {
  EventExample eventExample;

  function setUp() public {
    eventExample = new EventExample();
  }

  function testSetValueEmitsEvent() public {
    eventExample.setValue(42);
    require(eventExample.value() == 42, "value should be 42");
  }
}
```

TS test: event-trace-example.ts
```
import { describe, it } from "mocha";

import { expect } from "chai";

import hre from "hardhat";

describe("EventExample (provider path) - event trace repro", () => {
  it("emits ValueSet; its [event] line should appear in -vvvv traces but does not", async () => {
    const { ethers } = await hre.network.create();

    const EventExample = await ethers.getContractFactory("EventExample");
    const eventExample = await EventExample.deploy();
    await eventExample.waitForDeployment();

    const tx = await eventExample.setValue(42n);
    await tx.wait();

    // Proves the event really is emitted on-chain; the bug is purely that the
    // verbose call trace printed above never includes the decoded `[event]`.
    await expect(tx).to.emit(eventExample, "ValueSet");
  });
});
```

popescuoctavian:
This PR fixes an issue where the JSON-RPC provider verbose call traces never show emitted events when running `-vvvv` tests. It adds event log forwarding to the `EvmObserver` and enables logs recording in the tracing config.

closes #1542

### DIFF
--- a/.changeset/fruity-crabs-listen.md
+++ b/.changeset/fruity-crabs-listen.md
@@ -1,0 +1,5 @@
+---
+"@nomicfoundation/edr": patch
+---
+
+Fixed JSON-RPC provider call traces omitting event logs
--- a/crates/edr_provider/src/observability.rs
+++ b/crates/edr_provider/src/observability.rs
@@ -16,7 +16,7 @@ use edr_evm::{ExecutionResultAndStateWithMetadata, ExecutionResultWithMetadata};
 use edr_gas_report::SyncOnCollectedGasReportCallback;
 use edr_inspector_bytecode::ExecutedBytecodeCollector;
 use edr_primitives::{Address, Bytes, HashMap, HashSet};
-use edr_receipt::ExecutionResult;
+use edr_receipt::{log::ExecutionLog, ExecutionResult};
 use edr_solidity::{
     config::IncludeTraces, contract_decoder::ContractDecoder, tracing::SolidityTracingInspector,
 };
@@ -163,7 +163,9 @@ impl EvmObserver {
         let tracing_config = if config.verbose_raw_tracing {
             TracingInspectorConfig::all()
         } else {
-            TracingInspectorConfig::default_parity().set_steps(true)
+            TracingInspectorConfig::default_parity()
+                .set_steps(true)
+                .set_record_logs(true)
         };
 
         Self {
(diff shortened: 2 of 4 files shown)


Wodann (approved):
Thanks for the quick fix!

PR #1563, opened 2026-07-16
ci: add filters for HH3 regression benchmark

popescuoctavian:
This PR integrates into EDR the ability to filter which commands/steps (benchmarks) and which scenarios run as part of the regression benchmark via glob patterns.

Example:

`/bench hardhat-ref=ci/regression-benchmark-filters scenarios=1inch* benchmarks="test solidity"`

Runs the `test solidity` command on all the `1inch` scenarios (1inch-aqua, 1inch-cross-chain-swap, 1inch-swap-vm).

### DIFF
--- a/.github/workflows/hh3-regression-benchmark.yml
+++ b/.github/workflows/hh3-regression-benchmark.yml
@@ -19,8 +19,20 @@ on:
         required: false
         type: string
         default: "main"
-  # ChatOps: comment `/bench` (optionally `/bench hardhat-ref=<branch|sha|PR>`)
-  # on a same-repo PR. Authorization + gating happens in the `setup` job.
+      scenario-filter:
+        description: "Glob(s) selecting which projects/scenarios to run, by directory name (e.g. `1inch*`). Comma-separated. Empty = all. Forwarded to `bench:regression --scenarios`."
+        required: false
+        type: string
+        default: ""
+      benchmark-filter:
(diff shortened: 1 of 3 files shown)


popescuoctavian:
/bench hardhat-ref=ci/regression-benchmark-filters scenarios=1inch* benchmarks="test solidity"

.github/workflows/hh3-regression-benchmark.yml
```diff
@@ -212,17 +228,34 @@ jobs:
...
+        #   otherwise the whole benchmark runs.
         run: |
+          set -euo pipefail
+          extra_args=()
+          if [ -n "${SCENARIO_FILTER:-}" ]; then
+            extra_args+=(--scenarios "$SCENARIO_FILTER")
+          fi
+          if [ -n "${BENCHMARK_FILTER:-}" ]; then
+            extra_args+=(--benchmarks "$BENCHMARK_FILTER")
+          fi
```
Wodann: Could this be replaced by always setting the `$BENCHMARK_FILTER` and `$SCENARIO_FILTER` env vars, but e.g. with `*`—to match all scenarios/benchmarks—if absent?

Wodann (changes requested):
For EDR, the `main` branch should only have to with a `test*` filter. Similarly, I think we should default the `workflow_dispatch` and `/bench` comment to run with a filter that only selects those benchmark commands/steps.

The reason for this is that:
1. it saves a lot of CI time (maybe as much as 90% of time)
2. compilation is not affected by EDR in any way

.github/workflows/hh3-regression-benchmark.yml
popescuoctavian: Addressed by 06476f6755e551c20d74a65dbfcb31e75a28bf25.

Wodann (approved):
Thanks for addressing the feedback. LGTM!

PR #1564, opened 2026-07-16
Bump Rust toolchain to 1.97.1

nebasuke:
Bumps the pinned Rust toolchain from 1.96.1 to 1.97.1 to pick up the LLVM fix for the enum-niche miscompilation ([rust-lang/rust#159035](https://github.com/rust-lang/rust/issues/159035)) that crashed macOS x86-64 builds in solx.

With the bump to 1.97 I had to apply a few more formatting fixes.

# Claude summary

The only pin in the repo is the `rust-toolchain` file; CI's `setup-rust` action reads it by default, so all jobs pick up the new version automatically. The explicit `stable`/`nightly` references in other workflows are intentional and unaffected. No `rust-version` MSRV field exists to update.

See NomicFoundation/solx#547 for the full reasoning.

### DIFF
--- a/crates/foundry/evm/evm/src/executors/invariant/mod.rs
+++ b/crates/foundry/evm/evm/src/executors/invariant/mod.rs
@@ -633,7 +633,7 @@ impl<
         // Throw an error to abort test run if the invariant function accepts input
         // params
         if !invariant_contract.invariant_function.inputs.is_empty() {
-            return Err(eyre!("Invariant test function should have no inputs"))?;
+            Err(eyre!("Invariant test function should have no inputs"))?;
         }
 
         let (invariant_test, mut corpus_manager) =
@@ -673,7 +673,7 @@ impl<
 
             // We stop the run immediately if we have reverted, and `fail_on_revert` is set.
             if self.config.fail_on_revert && invariant_test.reverts() > 0 {
-                return Err(eyre!("call reverted"))?;
+                Err(eyre!("call reverted"))?;
             }
 
             while current_run.depth < self.config.depth {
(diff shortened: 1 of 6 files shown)


crates/foundry/evm/evm/src/executors/invariant/mod.rs
```diff
@@ -633,7 +633,7 @@ impl<
         // Throw an error to abort test run if the invariant function accepts input
         // params
         if !invariant_contract.invariant_function.inputs.is_empty() {
-            return Err(eyre!("Invariant test function should have no inputs"))?;
+            Err(eyre!("Invariant test function should have no inputs"))?;
```
anaPerezGhiglia: I think the following is more idiomatic for this case. It states clearly that the call flow should be interrupted. My instinct is that `?` is preferred for _propagating_ received errors rather than _producing_ them

```rust
 return Err(eyre!("...").into());
```

crates/foundry/evm/evm/src/executors/invariant/mod.rs
```diff
@@ -673,7 +673,7 @@ impl<
 
             // We stop the run immediately if we have reverted, and `fail_on_revert` is set.
             if self.config.fail_on_revert && invariant_test.reverts() > 0 {
-                return Err(eyre!("call reverted"))?;
+                Err(eyre!("call reverted"))?;
```
anaPerezGhiglia: ditto

anaPerezGhiglia (approved):
Thanks for taking care of this!
I left a code style suggestion, but I'll leave the final decision to you

crates/foundry/evm/evm/src/executors/invariant/mod.rs
nebasuke: Good suggestion, thank you! Applied to both.

nebasuke:
Previous benchmark runs did not indicate this test, merging.

PR #1569, opened 2026-07-20
fix: restore per-compiler `recursion_start_idx` in frame filtering

marianfe:
Follow-up to #1425. 

Issue: There was a regression in the javascript integration test sweep that compares solc and solx stack traces, specifically related to whether the idx recursion start strategy in error_inferrer is needed or not. 

The root cause has to do with the fact that the recursion is checked per message, not per top of the call stack, so a recursion at idx == 0 is possible within a given message. 

## Options verified

Both unified configurations were tested against the solx-parity-sweep, changing only the index condition:

1) Unified `idx > 0` (default state after #1425): `DeepRecursionTest#testDeepRecursion` fails — solc renders the correct 5 frames (three `this.recurse(depth - 1)` call sites at `:111`), solx renders only 3: two recursion frames are dropped.

2) Exemption from `idx 0` for both compilers: `DeepRecursionTest` passes, but `ModifierRevertTest` and `NestedModifierRevertTest` regress on the **solc** side (3 → 4 frames): the duplicate modifier-entry frame that solc's `idx > 0` check was removing is incorrectly kept.

No single index works with both compilers.

## Rationale 

`filter_redundant_frames` runs per message: `get_stack_trace` filters every nested CALL's sub-stack, and the parent embeds the result after pushing the call-site frame. When the parent contributed no frames of its own — common under solx, whose DWARF jump classification does not push dispatcher-entry frames the way solc's source maps do — the merged list starts with the call-site frame at `idx 0`. For external self-recursion that frame and the child's first frame resolve to the same statement, and without the exemption the enclosing-range rule drops one recursion level per nesting depth.

## Change

Restores `TraceStrategy::recursion_start_idx` (solc = 1, solx = 0) that was originally added in the draft of #1425 prior to refactoring, and thread the strategy into `filter_redundant_frames`. 

Solc behaviour is identical to upstream Hardhat.
solx now correctly passes the javascript sweep scenario tests.

### DIFF
--- a/.changeset/solx-recursion-frame-filtering.md
+++ b/.changeset/solx-recursion-frame-filtering.md
@@ -1,0 +1,5 @@
+---
+"@nomicfoundation/edr": patch
+---
+
+Fixed missing recursion frames in solx stack traces for external self-recursive calls.
--- a/crates/edr_solidity/src/error_inferrer.rs
+++ b/crates/edr_solidity/src/error_inferrer.rs
@@ -22,7 +22,7 @@ use crate::{
     },
     trace_strategy::{
         function_start_source_reference, source_location_to_source_reference, PanicHelperContext,
-        TraceStrategyError,
+        TraceStrategy, TraceStrategyError,
     },
 };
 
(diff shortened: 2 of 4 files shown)


popescuoctavian (approved):
LGTM. Thanks for verifying this and providing context on the differences between solc and solx here.

Wodann (approved):

PR #1572, opened 2026-07-22
test(solx-parity-sweep): add bare-modifier-revert and linked-library scenarios

nebasuke:
Deliberate split from #1552. First add the test cases as sweep scenarios with a solx <-> solc diff, then fix it in #1552. Also makes #1552 a tiny bit more digestable. Main test case is a bare revert, that is not handled well for solx (follow up in #1552).

Fable fumbled completely, and tripped over itself 3 times trying to regenerate the fixtures. I've added documentation, a regeneration command, a reproducible test, and a readme that hopefully will make it easier for future humans and AI.

Note that I've changed the fixtures to include only the needed compiled forge-std code. While it was omitted from the committed source code, the complete library was added to the compiled code. The regen-fixtures I've added does this in a reproducible manner. 

I'm not fully happy with `regen-fixtures,js` as It's still a bit complex/awkward, but it will be iterated on in #1552 (likely moved to `gen-solx-fixtures`) and after hardhat-solx gets published we can drop more of the logic (as already mentioned in TODOs by marianfe). 

Also note I had to do a regen for solx to incorporate the change to in hardhat-solx to `-O1` (see  https://github.com/NomicFoundation/hardhat/pull/8419) as it was previously compiled using `-O3`. 

# Claude summary

Two additions to the parity sweep's scenario corpus — one pinning a known solx↔solc divergence for #1552 to flip, one at parity as a regression net — plus the fixture work described above, detailed below: the regeneration and forge-std trim, the `pnpm regen-fixtures` flow, and the guard test enforcing the source/fixture coupling.

## `BareModifierRevertTest` — pinned divergence ("red" vs solc)

A modifier's bare `revert()` compiles to a shared revert helper that solx leaves unmapped in the DWARF, and the revert carries no return data — so every location-anchored revert heuristic misses. Today's shape, pinned in `scenariosDivergingFromSolc`:

- **solc:** 3 frames, bottom at the `revert()` statement inside the modifier (`BareRevertGuardTarget.guarded`, line 442).
- **solx:** 1 frame — the target contract's frames vanish entirely and the trace collapses to the test's own call site (`BareModifierRevertTest.testBareModifierRevert`, line 454) with the raw `EvmError: Revert`.

When #1552 lands (selector-based function recovery + revert walk-back), this golden breaks in the improvement direction: the trace regains the target frames and the residual shrinks to a one-line delta (solx reports the guard condition, solc the `revert()` statement — the statement's line is absent from solx's line table because LLVM merges the identical revert blocks; WONTFIX solx-side). The follow-up updates the pin to that shape.

## `ExternalLibraryRevertTest` — parity regression net

`ExternalRevertingLib.fail` is `public`, so the call is a DELEGATECALL into a separately deployed, link-referenced library, and the frame must resolve inside the **library's own** debug info. The existing `RevertingLib` scenario doesn't exercise this — its `internal` function inlines into the caller, so no linking and no DELEGATECALL. This scenario is at parity with solc today and joins the strict check; it exists so a regression in link-reference handling or cross-artifact frame resolution breaks loudly. (The provider-path twin, with manual `linkReferences` substitution, is in #1552's `StackTraceScenarios` fixture.)

## Scenarios fixture: regenerated via the standard flow

The fixture pair (`crates/edr_solidity/fixtures/solx_compiler_{input,output}_scenarios.json`) is this project's own build output: the build-info pair from a solx-profile hardhat compile, which `pnpm test` runs. What that compile produces depends on the toolchain of the day — hardhat-solx's settings (the explicit `-O1` default since hardhat#8419 fixed the plugin) and the solx release selected by `SOLIDITY_TO_SOLX_VERSION_MAP` (`0.8.34` → `0.1.4` today, `0.1.6` bump planned). The committed snapshot predated #8419 — captured at solx's `-O3` default, per the note above — so the standard flow no longer reproduced it; the committed input, moreover, had never been a capture at all (#1425 hand-wrote it as scaffolding and only captured the output — which is why its settings never matched the output beside it).

This PR regenerates **both** files through the standard flow — fixed plugin, solx 0.1.4 via the map, now including this PR's scenario contracts — and the forge-std trim noted above is worth quantifying: the fixture drops from 42.6 MB to 7.6 MB, and fixture parsing for the 78-test suite from 6.5 s to ~1 s. The cut keeps `Scenarios.t.sol` **plus its inheritance closure** (the 7 forge-std bases): EDR's build model resolves inherited functions (`StdInvariant`'s `targetSelectors()` etc.) through the base contracts' ASTs via `linearizedBaseContracts`, so `sources` and `contracts` must cover the same closure — an earlier leaf-only cut broke `edr_provider`'s solx stack-trace tests in CI on exactly that. What's dropped (`console`/`safeconsole`/`Vm` ASTs, ~35 MB) is read by nothing; the regen script computes the closure so it stays correct as scenarios gain new bases. The filter is only a size optimization, though — the fixture's sufficiency is enforced where the knowledge lives: `scenarios_fixture_satisfies_contract_metadata_extraction` in `edr_solidity` runs contract metadata extraction (the exact function that failed on CI) over the committed fixture, red-proven against the over-filtered version, so a bad regen now fails `cargo test -p edr_solidity` directly rather than two hops away in provider integration tests. The committed fixtures are byte-for-byte the output of `pnpm regen-fixtures`, so input and output are consistent for the first time. Re-pinned in the same change: the four PC-anchored dwarf tests (the INVALID-opcode one now finds its instruction by opcode rather than PC; the moved PCs are the `-O3` → `-O1` shift). Everything else holds, including the three provider goldens.

`pnpm regen-fixtures` (pure Node, no jq) rebuilds the gitignored `contracts/` corpus copy from scratch — a stray leftover `.sol` would otherwise be silently compiled into the committed input — wipes `artifacts/build-info` so exactly one fresh pair exists, compiles the solx profile, writes the extracted pair, and reports whether the committed files actually changed; run on a clean tree it doubles as a staleness check. It fails loudly if the compile output lacks the scenarios section. It is deliberately a separate command from `pnpm test`: a test run that rewrites the fixtures would absorb toolchain drift instead of detecting it.

Investigating the provenance also surfaced [solx#594](https://github.com/NomicFoundation/solx/issues/594): solx embeds the build directory in the DWARF, so `debugInfo` bytes differ across checkouts even on identical toolchains — expected noise for regenerators until it's fixed.

## Append-only guard for `Scenarios.t.sol`

`Scenarios.t.sol` is spliced at test time into the committed scenarios fixture, whose AST and DWARF reference byte offsets into the source as it existed at generation time. Between regenerations the file must only be appended to, but nothing enforced that: a rename or formatter run inside the generation-time prefix desyncs every line assertion on the fixture, in the worst case to plausible-but-wrong locations instead of loud failures.

New test `scenarios_source_is_append_only` pins the generation-time prefix with its keccak hash. The prefix length is read from the fixture's own `SourceUnit` span, so a regeneration moves the hashed prefix automatically and forces the hash to be re-pinned — the failing assertion prints the new value, and its message states the rule: append, or regenerate and re-pin.

## Docs

- Sweep README: the regeneration command and re-pin instruction up front, with the provenance details (toolchain dependence, how to re-derive the PC anchors, output filter, input splice, solx#594) as a follow-on list; plus an adding-scenarios walkthrough (append-only rule, dynamic key discovery, divergence-table pinning etiquette).
- New fixtures index at `crates/edr_solidity/fixtures/README.md`, written for contributors: each committed compiler artifact mapped to what generates it and how to regenerate it, plus the conventions previously only discoverable by archaeology (source splicing, the CBOR `solcx` version stamp for identifying which solx produced an output, solx#594).

## Notes

- Verified: sweep 4/4 green against solx 0.1.4 (hardhat-solx's current pin), with both new scenarios present in both profiles; `edr_solidity` 79/79 and `edr_provider` `solx_stack_trace` 3/3 against the regenerated fixture; `pnpm regen-fixtures` reproduces both committed files byte-for-byte (clean tree after a fresh run); provider suite at its known baseline (the 12 fork tests need `ALCHEMY_URL`).
- Rebased onto main after #1571 merged; the new scenario's row is added to the README's reworked pinned-divergences table.
- Test-only: labeled `no changeset needed`.


marianfe (approved):
great addition with the instructions to regenerate fixtures. i think directionally the PR makes sense, didn't have time to review in detail, i'll defer to @popescuoctavian . however, one thing that i'm not sure about, is how come the output of the compiler went down by 867k lines ? is that intentional? did the formatting change that much between solx versions? @nebasuke

popescuoctavian:
@marianfe IIUC this is due to the [size optimizations ](https://github.com/NomicFoundation/edr/blob/f3ce111c95405e649b7ff3bcebf6ce71e9b1c8a0/js/integration-tests/solx-parity-sweep/scripts/regen-fixtures.js#L74-L80) added in `regen-fixtures.js` - it drops contracts from `forge-std` which are not referenced, which drastically reduces the overall size

popescuoctavian (approved):
LGTM, thanks @nebasuke ! The output size optimizations are really nice and I appreciate the detailed documentation.

nebasuke:
> @marianfe IIUC this is due to the [size optimizations ](https://github.com/NomicFoundation/edr/blob/f3ce111c95405e649b7ff3bcebf6ce71e9b1c8a0/js/integration-tests/solx-parity-sweep/scripts/regen-fixtures.js#L74-L80) added in `regen-fixtures.js` - it drops contracts from `forge-std` which are not referenced, which drastically reduces the overall size

Exactly, this was a conscious choice (see also my PR description). Note that while it is not the most elegant (but importantly, reproducible), I thought it would make sense to do medium term until we can drop more of the infrastructure when hardhat-solx is published and we move this to be CI tested.

PR #1574, opened 2026-07-23
test: consolidate the solx fixtures and run the parity sweep in CI

nebasuke:
Now with EDR patched to Hardhat 3.15.0+, we can import the hardhat-slang-solx plugin, and run the solx fixture via that. This allows it to run in CI, rather than depending on a local install. Additionally I migrated over the fixtures to contracts, rather than the JS toolchain that was there before. This gets rid of our last remaining JS fixture, and a good chunk of hardcoded code. 

It adds some extended CI time to the integration test job . Had some fun in #1755 but decided to cut it as it was too finicky. In the end it will not make a difference to CI runs as the integration test job will normally finish much earlier than the Test runs. 

# Claude summary

Every Rust test that pinned solx DWARF/trace behavior against the committed sweep fixture now pins it against the regenerable `stack_trace_scenarios` fixture instead, and the sweep fixture — with its entire regeneration apparatus — is deleted. One regen flow remains for all committed solx fixtures: `cargo run -p edr_tool_cli -- gen-solx-fixtures <solx>`. The committed fixtures are generated with solx 0.1.8, and the sweep installs the published `hardhat-slang-solx` and runs for real — in CI too.

## Why

The repo carried two fixture regeneration flows. The `gen-solx-fixtures` tool regenerates its fixtures from committed inputs with any solx binary — no JS toolchain involved. The sweep `scenarios` fixture needed the hardhat-solx pipeline, a `pnpm regen-fixtures` script, and a frozen-prefix regime: because the live `Scenarios.t.sol` was spliced into a fixture compiled from an older version of itself, the source was append-only, enforced by a keccak prefix-hash guard, documented in four places, with a planned CI sync test to keep source and fixture from drifting.

All of that apparatus existed to serve the fixture's Rust consumers. Moving the consumers dissolves it: the sweep itself compiles its corpus live on every run and never read the committed fixture.

## What moved

- **Corpus** (`StackTraceScenarios.sol`): ~20 new plain contracts mirroring the sweep scenarios' shapes — the require/assert/panic family (0x01/0x11/0x12/0x21/0x31/0x32), custom error, constructor reverts (direct + via internal helper), the recursion trio (external deep, internal, cross-contract mutual), internal helper chain, internal library, modifiers, fallback/receive reverts, inline-assembly `revert`, `invalid()`. Forge-std is not needed: every pinned DWARF shape is a plain contract construct — the forge harness framing (invariant/fuzz/expectRevert) never reached the DWARF and stays sweep-only.
- **`edr_provider` breadth tests**: every test that deployed from the sweep fixture now deploys the plain equivalent from the regenerable fixture, through the existing `stack_trace_scenarios_provider` helpers. Assertions unchanged in kind — entry variant + source line (+ panic statement line), full `assert_trace_shape` pins where they existed.
- **`edr_solidity` dwarf pins**: all line- and PC-anchored tests re-anchored against the new fixture via the documented probe. Every predicted line and trace-shape pin held on the first run; only three PC anchors needed re-deriving (`0x26` divide REVERT, `0x36` assembly REVERT, `0x87` modifier require — the last one landed on the same PC as the old corpus).

## Coverage preserved deliberately

- **Cross-file inheritance**: new `StackTraceScenariosBase.sol` with a base contract, inherited by a scenario contract in the main file. The build model resolves inherited functions through base contracts' ASTs (`linearizedBaseContracts`) — the exact shape that broke CI during #1572's fixture cut — and a new provider test asserts an inherited function's revert resolves into the base's own source file. Both `gen-solx-fixtures` stack-trace entries (optimizer mode 1 and the mode 3 twin from #1577) are multi-source now, and the test helpers splice both sources.
- **Dense multi-contract compilation unit**: the cross-contract line-leak guard (an INVALID opcode once picked up a different contract's line row) keeps its neighbors — the corpus is one file with ~30 contracts.

## Deleted, and why that's safe

| Deleted | Why |
| --- | --- |
| `solx_compiler_{input,output}_scenarios.json` (7.5 MB) | zero consumers left |
| `scripts/regen-fixtures.js` + `pnpm regen-fixtures` | nothing left to regenerate |
| `scenarios_source_is_append_only` (keccak guard) | the corpus is no longer spliced into a frozen fixture |
| `scenarios_fixture_satisfies_contract_metadata_extraction` | it policed the regen script's closure filter; `gen-solx-fixtures` writes solx's verbatim output — no filter exists |
| pretest copy step in `maybe-build.js` | `Scenarios.t.sol` now lives in the sweep project directly |
| `pin_invariant_failure` | DWARF-identical to `pin_direct_require` once the forge invariant framing (which never reached the DWARF) is gone |

This also retires the planned compare-only CI sync test from the sweep-in-CI follow-up list: there is no committed sweep artifact left to drift. The remaining Rust↔sweep coupling is intentional duplication — the same shapes exist in two independently compiled corpora, one pinned in Rust against a committed fixture, one checked live against solc for parity.

## solx 0.1.8

The last commit regenerates all three committed pairs (Counter, stack-trace scenarios at optimizer mode 1, and the mode 3 twin) with the 0.1.8 release — the version `hardhat-slang-solx`'s map selects for 0.8.34. For this corpus, 0.1.8 codegen is byte-identical to 0.1.6 apart from the CBOR `solcx` stamp (plus the two creation bytecodes that embed a child contract's stamp). Every line, trace-shape and PC pin held as is, so there is no re-pin commit. The version-specific prose (fixtures README stamp example, sweep README version-map note) now names 0.1.8.

## Sweep side

`Scenarios.t.sol` moves into the sweep as an ordinary committed source (`contracts/` is no longer gitignored). Adding a sweep scenario is now: edit the file, run `pnpm test` — no fixture regen, no frozen prefix, no cross-crate re-pin. The sweep's own behavior is untouched: same corpus, same live compile, same divergence pins.

The sweep now installs `@nomicfoundation/hardhat-slang-solx` 3.0.3 as an ordinary devDependency (its version map selects solx 0.1.8 for 0.8.34). That needs the workspace's hardhat past the plugin's `^3.15.0` peer range, which `main` satisfies at 3.16.0. With it, the self-skip apparatus goes: the type shim, `maybe-build.js` (pretest is the same `pnpm build:dev` as the other integration suites) and the dynamic-import guard. The plugin requires the build profile to be named `slang-solx`, so the `solx` profile is renamed.

First real run, against solx 0.1.8 through the published plugin: 4/4, with the `InternalRecurseTest` divergence pin unchanged. Because `edr-integration-tests` runs `pnpm run --recursive --filter './js/integration-tests/*' test`, the sweep now runs in CI alongside `solidity-tests` — the sweep-in-CI item from the follow-up list, with no extra job. Each CI run downloads the ~60 MiB solx binary into Hardhat's global compiler cache (nothing caches that directory today, same as solc); caching it is a small follow-up if the download turns out to be flaky or slow.

## Notes

- Validation on the tip: `cargo test -p edr_solidity` 81/81; `cargo test -p edr_provider --features test-utils --test main solx_stack_trace` 37/37 (36 on `main` + the inherited-function test); a fresh `gen-solx-fixtures` run with 0.1.8 is a zero-diff no-op; the parity sweep passes 4/4 against the published plugin; `pnpm lint` clean.
- Test-only: `no changeset needed`.


popescuoctavian (approved):
LGTM, thanks!

PR #1577, opened 2026-07-25
fix(edr_solidity): solx stack-trace attribution for declaration-level and line-0 DWARF

nebasuke:
Adding test coverage for stack trace tests, along with a test fixture regen within edr_tool_cli. I will follow-up after hardhat-solx is published as this gives multiple opportunities to simplify/merge the infrastructure and add it to CI. 

This PR notable bumps the used version of solx from 0.1.4 to 0.1.6.

Fixes for solx -> stack trace mappings (as part of 0.1.4 -> 0.1.6):
* bare reverts
* handling of generated code (that now has line-0 since https://github.com/NomicFoundation/solx/pull/582)

Further test coverage in https://github.com/NomicFoundation/edr/pull/1552.  Note that due to the nature of the fixtures I pre-added all scenarios to `StackTraceScenarios.sol` including the tests I saved for #1552. 

Added test cases for `-O3` in specific, as DWARF mapping are slightly worse there. This likely due to underlying solx optimisations, but I thought it would be good to have as a red-green test, as the original tests (by Marian) were using `-O3`.

# Claude summary

Fixes the solx (DWARF) stack-trace attribution gaps found while broadening provider-path coverage, and adapts the DWARF consumption to solx 0.1.6's line-0 emission — including a frame-loss regression the parity sweep caught on 0.1.6 that would otherwise have shipped. The fixtures move to solx 0.1.6 and become regenerable in-repo, with a mode-3 variant keeping every compat path red-provable.

Split out of #1552, which now stacks the broad provider-path test coverage on top of this.

## Root cause

solx attributes compiler-generated helper code — calldata decoding, shared revert builders — to the **declaration** of the enclosing function or contract in its DWARF line table, where solc leaves the same code **unmapped** in source maps. Since 0.1.6 solx emits **line 0** for that code instead ([solx#582](https://github.com/NomicFoundation/solx/pull/582)), which the decoder's Pass-3 fallback turns back into the enclosing function's declaration — so the declaration-attributed paths below are the permanent consumer architecture, not a pre-0.1.6 shim. At hardhat-solx's default `-O1`, solx 0.1.6 statement-attributes most of this code directly; other optimizer modes and older releases still produce the declaration-attributed and unmapped shapes.

Three location-anchored inference paths broke on that difference:

1. **`fails_right_after_call` / `is_call_failed_error`** compare every post-CALL step location to the call site by strict equality. Under solc, unmapped helper instructions are skipped; under solx, the same instructions carry declaration-level locations (whose range merely *contains* the statement), so the heuristics bailed and the failure misclassified as a generic revert at the call site.
2. **Shared revert helpers flattened out of a modifier** are attributed to the modified function's declaration line, so a modifier's failing `require` was reported at the function signature. The statement's real line *is* present in the line table — on the message-building instructions executed just before the revert — so this is recoverable EDR-side.
3. **A modifier's bare `revert()`** compiles to a shared helper that is **entirely unmapped**, and its return data is empty, so every revert heuristic missed and inference degraded to `OtherExecutionError` at the contract declaration.

Both fixes for (1) and (2) live in the solx strategy; `SolcTraceStrategy` is behavior-identical (strict equality kept, new parameters ignored):

- New `TraceStrategy::step_still_at_statement(step, statement)`: solx treats a step location that **contains** the statement (declaration-level padding) as "still at the statement"; solc keeps `==`. Used by `is_last_location`.
- `TraceStrategy::revert_source_reference` receives a lazy `step_pcs` thunk (same pattern as `PanicHelperContext`). When the reverting instruction sits on the function's declaration line, `SolxTraceStrategy` walks the executed steps back to the last statement-level location.

For (3), the revert branch falls back to a strategy hook when the reverting instruction has no resolvable containing function: `failing_function_from_calldata` resolves the called function from the calldata selector under solx (solc returns `None`, behavior-identical), gated to `REVERT` opcodes and to calldata that actually decodes for the resolved function, so dispatch-level reverts still classify as `InvalidParamsError`. `revert_source_reference` takes the location as an `Option` — unmapped gets the same walk-back recovery as declaration padding, with the function start as last resort.

### Review hardening of the walk-back

A review round produced three constraints (no existing golden changed): only statements of the failing function or of a modifier executed in the frame qualify — a statement of any other function ends the walk (previously the first resolvable location won, even the dispatcher's or an earlier internal call's); declaration matching requires source-file identity instead of bare line numbers; and the contract-declaration line counts as declaration-level padding alongside the function's. The `step_still_at_statement` containment proxy is deliberately kept as-is: it can only over-match when every instruction of a later statement is declaration-attributed, and the parity sweep guards that residual. A later pass tried to also unify the constructor/fallback/receive display names between the two source-reference builders; the hardhat-tests corpus rejected it — function-start frames pin the raw (empty) AST name — so the asymmetry stays, documented at the site.

## solx 0.1.6 adaptation: the line-0 frame loss

solx 0.1.6 also emits `DW_AT_call_line 0` for inlined subroutines whose call site is compiler-generated — most importantly user functions inlined into the `__entry` dispatch. That cost solx traces the `Contract.function` callstack frame solc renders between the caller frame and the revert (sweep scenarios `ModifierRevertTest`/`NestedModifierRevertTest` collapsed from 3 frames to 2). Two-part fix:

1. `range_call_site_to_location` falls back to the inlined function's own `decl_file`/`decl_line` when `call_file`/`call_line` are absent or zero — the same declaration line solc renders for that frame. (solx's DWARF carries no `decl_column` — LLVM subprograms track only file and line — so the resolved offset skips the line's indentation to land inside the function's AST span.)
2. `intermediate_frames` previously deduped against the raw failing function — under Pass-3 synthesis that names the flattened-into function while the bottom frame actually renders under the modifier (via the walk-back), so the just-recovered frame was dropped as a "duplicate". It now dedups against the bottom frame's already-resolved source reference, computed before the intermediate frames at each call site.

## Effect on the traces

- `ReturndataSizeError` surfaces **at the call site** for both a returndata-size mismatch and a typed call to a codeless account, replacing a bare revert frame at the call site. This is parity with the solc route: solc ≥ 0.8.10 emits no EXTCODESIZE probe for returndata-expecting calls either, so both routes classify the EOA case as a returndata failure. (A true `NoncontractAccountCalledError` would need a void-returning call scenario — future coverage.)
- A multi-statement modifier revert reports the failing `require`'s line instead of the declaration line — removing the provider-level twin of the sweep's pinned `NestedModifierRevertTest` divergence.
- A modifier's bare `revert()` reports the `revert()` statement inside the modifier, matching solc. At optimizer mode 3 — where the entry path carries no statement line — the walk-back lands one line earlier, on the guard condition, instead of collapsing to `OtherExecutionError` at the contract declaration; the remaining one-line gap is the solx-side residual.
- The called function's frame survives a cross-contract modifier revert.

The `solx_stack_trace.rs` diff is deliberately additive — the existing tests and helpers are untouched (the one deletion is an import line), so what shows up there is what changed semantically. Deduplicating those helpers happens in the stacked coverage PR, where the breadth of tests motivates it.

Includes a changeset (patch), following #1569's precedent for user-visible stack-trace fixes.

## Test provenance

All eight new provider goldens (plus the red-green decoder unit test for the line-0 call site) fail on `main`. The `-O1` fixtures can no longer reproduce the declaration-attribution symptoms — solx 0.1.6 statement-attributes the default pipeline — so those tests pin the parity shape, while the **mode-3 fixture variant** pins the compat inference paths. Each mechanism was verified load-bearing by disabling it individually:

| Mechanism | Red without it (and on `main`) |
| --- | --- |
| calldata-selector recovery | `mode3_bare_modifier_revert_recovers_the_failing_function` (`OtherExecutionError` at the contract declaration) |
| declaration-attribution detection | `mode3_nested…` / `mode3_cross_contract…` (revert at the declaration line instead of the failing `require`) |
| revert walk-back | all three mode-3 goldens |
| `step_still_at_statement` containment | both returndata goldens (misclassified as a generic revert) |
| dwarf line-0 → declaration fallback | `cross_contract…` (dropped callee frame) + the decoder unit test |

Trace-shape assertions go through `assert_trace_shape`, which prints one line per entry: a gained frame may be an improvement (update the pin), a lost frame is a regression.

## Fixtures: solx 0.1.6 + `gen-solx-fixtures`

`cargo run -p edr_tool_cli -- gen-solx-fixtures <solx-binary>` regenerates the self-contained fixture outputs (`counter`, `stack_trace_scenarios`, `stack_trace_scenarios_mode3`): it splices the `fixtures/sources/` contents into the committed input JSON (whose `content` fields are deliberately empty), runs `solx --standard-json`, and writes solx's verbatim output. It lives next to `gen-execution-api` — the repo's existing home for generators of committed artifacts. That surfaced two latent bugs fixed here: `update::project_root()` still assumed the crate's pre-#1139 `crates/tools/` depth, so it resolved to `crates/` instead of the repo root; and `update()` failed with a pathless ENOENT on a not-yet-committed output file instead of creating it.

The inputs pin `settings.optimizer.mode` explicitly (trace shapes differ per mode) at hardhat-solx's `-O1`, the pipeline these fixtures stand in for — not solx's own default. The mode-3 variant of `stack_trace_scenarios` exists because `-O1` DWARF is statement-attributed since 0.1.6: without it, the walk-back, declaration-attribution, and selector-recovery paths would be live code with no reachable input in the repo — they guard non-default modes and pre-0.1.6 artifacts. Re-running the subcommand against the committed tree yields zero diff, so the fixtures round-trip deterministically.

**Everything is regenerated with solx 0.1.6** (released 2026-07-22): the debug info changed for every contract, deployed bytecode is identical modulo the metadata hash, and every pre-existing golden holds. Two PC-anchored decoder tests moved: [solx#583](https://github.com/NomicFoundation/solx/pull/583) resolves inline-assembly source refs to Yul AST nodes, so the assembly `revert` and `invalid()` now report solc's statement lines instead of the function-decl fall-back.

The `scenarios` fixture is outside `gen-solx-fixtures`' reach: its input also names forge-std sources whose contents are scrubbed from the committed JSON, so it regenerates from the sweep project instead — `pnpm regen-fixtures`, the flow #1572 added — which is why it moves in its own step here. That also makes it the one fixture built by the real hardhat-solx pipeline rather than a bare `solx --standard-json`.

New `StackTraceScenarios.sol` fixture: plain contracts, no forge-std, so `gen-solx-fixtures` can rebuild it from the repo alone. It is committed whole — `GuardedBareRevert`, `ValidatedCounter(Caller)` and `ExpectsWord` pin the fixes here, the dispatch-level contracts are consumed by the stacked coverage PR. Splitting the file would move the line numbers every golden pins.

## Sweep pins

`BareModifierRevertTest` reaches parity with the bare-revert fix; `InlineAssemblyRevertTest` and `InvalidOpcodeTest` with solx#583. Only `InternalRecurseTest` (optimizer unrolls 3-deep self-recursion) stays pinned. The pin set now assumes hardhat-solx maps `0.8.34` to solx 0.1.6.

## Verification

- `solx_stack_trace.rs`: 11/11 (3 pre-existing goldens unchanged across the regeneration, 5 `-O1` + 3 mode-3 new; all 8 new ones proven red on `main`, plus the per-mechanism disable checks above).
- `edr_solidity` unit tests: 80/80; `edr_provider` harness: 54 lib + 191 integration tests green at the tip (fork tests skip without `ALCHEMY_URL`).
- JS parity sweep run locally against solx 0.1.4, 0.1.5 and 0.1.6 (hardhat-solx at `-O1`, local plugin build): 0.1.6 + these fixes → strict parity green; 0.1.4 and 0.1.5 + these fixes → still green, so the changes are inert for pre-line-0 artifacts.
- solc behavior is guarded code-structurally (solc strategy unchanged) and by the hardhat-tests stack-trace corpus in CI.


crates/tool/cli/src/update.rs
```diff
@@ -16,7 +16,15 @@ pub enum Mode {
...
 #[allow(unused)]
 pub fn update(path: &Path, contents: &str, mode: Mode) -> anyhow::Result<()> {
-    let old_contents = fs::read_to_string(path)?;
+    let old_contents = match fs::read_to_string(path) {
+        Ok(old_contents) => old_contents,
+        Err(error) if error.kind() == std::io::ErrorKind::NotFound && mode == Mode::Overwrite => {
+            eprintln!("creating {}", path.display());
+            fs::write(path, contents)?;
+            return Ok(());
+        }
```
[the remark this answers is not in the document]
nebasuke: Fixed in 2919162ea — the normalization is hoisted above the read, so the create-on-missing branch writes the same normalized bytes as the update path. Verified `gen-solx-fixtures` still round-trips with zero diff against the committed fixtures.

crates/tool/cli/src/solx_fixtures.rs
```diff
@@ -0,0 +1,192 @@
...
+        child.wait_with_output()
+    })?;
+    if !output.status.success() {
+        bail!(
+            "`{} {}` failed: {}",
+            solx.display(),
+            args.join(" "),
+            output.status
+        );
+    }
```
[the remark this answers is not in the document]
nebasuke: stderr is not discarded — it is deliberately left inherited (only stdout is piped), so solx's diagnostics stream to the terminal live while the tool runs; the warnings solx prints during a regen are visible exactly this way. Capturing stderr to embed it in the failure message would trade that live streaming for a duplicated dump after the fact, for a local dev tool whose output is already on screen.

crates/tool/cli/src/update.rs
```diff
@@ -16,9 +16,16 @@ pub enum Mode {
...
-    let old_contents = old_contents.replace("\r\n", "\n");
     let contents = contents.replace("\r\n", "\n");
+    let old_contents = match fs::read_to_string(path) {
+        Ok(old_contents) => old_contents.replace("\r\n", "\n"),
+        Err(error) if error.kind() == std::io::ErrorKind::NotFound && mode == Mode::Overwrite => {
+            eprintln!("creating {}", path.display());
+            fs::write(path, &contents)?;
+            return Ok(());
+        }
+        Err(error) => bail!("reading {}: {error}", path.display()),
```
popescuoctavian: Suggestion: use typed error instead of formatting it to a message (i.e. add message to context - `Err(error) => Err(error).context(format!("reading {}", path.display()))?`)

crates/edr_solidity/src/error_inferrer.rs
```diff
@@ -483,23 +512,26 @@ fn check_custom_errors<HaltReasonT: HaltReasonTrait>(
...
+        error_message,
+    )?;
+
     if let Some(loc) = &last_instruction.location
-        && let Some(failing_function) = loc.get_containing_function()?
+        && loc.get_containing_function()?.is_some()
     {
+        let bottom_source_reference = bottom_entry
+            .source_reference()
+            .expect("CustomError always carries a source reference");
```
popescuoctavian: Suggestion: the ErrorInferrer's convention is to use `InferrerError:InvariantViolation` for errors that that are are not locally-provable. I believe the same can be applied here for consistency.

crates/edr_solidity/src/trace_strategy.rs
```diff
@@ -25,12 +28,15 @@ pub enum TraceStrategyError {
     ContractMetadata(#[from] ContractMetadataError),
 }
 
+/// Lazily yields the trace's EVM step PCs in execution order. A thunk
+/// because trace steps are generic over the halt-reason type, which can't
+/// cross the object-safe trait boundary.
+pub type StepPcs<'a> = &'a dyn Fn() -> Vec<u32>;
```
popescuoctavian: Suggestion: the type alias hides the reference. Including `&` in the alias means that use sites no longer read as borrows. I'd avoid this by removing the reference from the alias and adding it to the use sites to make the borrow visible (`&StepPcs`).

popescuoctavian (approved):
LGTM overall! Left a few minor suggestions which are not blocking. Feel free to merge without requesting another review. Thanks!

crates/edr_solidity/src/trace_strategy.rs
nebasuke: Cool, thanks for the suggestion. Applied!

crates/edr_solidity/src/error_inferrer.rs
nebasuke: Ah thanks for the heads up, applied!

crates/tool/cli/src/update.rs
nebasuke: Applied directly

PR #1586, opened 2026-07-28
fix(edr_solidity_tests): include failing call's trace in invariant setup failures

popescuoctavian:
This PR is a follow-up to #1575.

When an invariant is already broken in the initial state, `invariant_fuzz` fails during its initial `assert_invariants` check and returns an error before any fuzzed calls run. There is no counterexample sequence to replay in that case, and the only copy of the failing call's trace was inside the `RawCallResult` that `FailedInvariantCaseData::new` consumed and discarded. In `Always` mode the stack trace was then built from the setup traces alone — whose last entry succeeded — producing an empty stack trace.

Fixed by carrying the trace through the error:

- `FailedInvariantCaseData` keeps the offending call's traces, exposed via `InvariantFuzzError::take_traces()`.
- `TestResult::invariant_setup_fail` merges those traces into `execution_traces`, mirroring `single_result`/`fuzz_result`, so the subsequent `collect_stack_trace` sees the failing call as the last trace.

The other invariant failure scenarios were already covered: invariants broken during fuzzing (including `fail_on_revert` and `afterInvariant` failures) replay the counterexample with step tracing via `replay_error`/`replay_run`, which regenerates the failing trace itself.

### DIFF
--- a/.changeset/shiny-planets-behave.md
+++ b/.changeset/shiny-planets-behave.md
@@ -1,0 +1,5 @@
+---
+"@nomicfoundation/edr": patch
+---
+
+Fixed missing Solidity test stack trace when an invariant is already broken at the campaign's initial check and stack traces are collected with `CollectStackTraces::Always`.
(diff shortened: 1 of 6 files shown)


Wodann (approved):
Thanks for the quick turnaround!

PR #1587, opened 2026-07-28
ci: pull node images from a GHCR mirror instead of Docker Hub

nebasuke:
I've seen the docker image flake for a while, particularly when creating more PRs. I've now taken the step of mirroring the images into ghcr.io, for EDR, privately. This is possible by using the GITHUB_TOKEN.

Mirrored docker images are already working in this PR! 

Documented the process of adding further node images. Did a couple of small quick fixes to the book to fix up a chapter link and an extension. 

Note: the Docker login warning is not a problem as it's a short-lived GITHUB_TOKEN (not a PAT) and is aimed at local dev. 

# Claude summary
## Problem

`edr-npm-release.yml` runs on every PR and makes ~16 Docker Hub pulls per run (`node:24-bullseye-slim` / `node:{22,24,26}[-alpine]` across the build matrix and the four Linux test-binding jobs). Docker Hub rate-limits pulls, which is the source of the intermittent pull failures we see in CI.

## Change

- New `mirror-docker-images.yml`: copies the node tags CI needs into `ghcr.io/nomicfoundation/edr/mirror/node:<tag>` using the runner's preinstalled `docker buildx imagetools create` (no new third-party actions). Runs weekly (upstream refreshes these tags in place with security patches), on `workflow_dispatch`, on pushes to `main` that touch the workflow itself, and on same-repo PRs that touch it — so a PR adding a tag (e.g. a new node version) mirrors and tests it before merge. Each copy retries with backoff and verifies the mirrored manifest digest matches upstream, so an unfaithful copy fails loudly.
- `edr-npm-release.yml`: all six image references now point at the mirror. GHCR sits on the same infrastructure as the runners, so these pulls are both faster and not subject to Docker Hub's limits.
- New book chapter (`Development → CI Docker image mirror`) documenting the add-a-node-version flow — add the tag to the mirror list in the same PR that changes the release matrix, and the PR trigger mirrors it before merge — plus the `GITHUB_TOKEN` access model. Someone editing the release matrix has no reason to open the mirror workflow and see its comments, hence a findable chapter.
- Drive-by (separate commit): two pre-existing book `SUMMARY.md` gaps — `10_dynamic_base_fee_params.md` was never listed, and the `08_cheatcodes` entry was missing its `.md` extension.

The mirror packages stay **private** (making org packages public needs an org admin), so every pull authenticates with `GITHUB_TOKEN`: `docker/login-action` before the raw `docker run` steps, and `registry`/`username`/`password` inputs on the `docker-run-action` legs (it's a container action, so a host `docker login` wouldn't reach it). Access works because publishing via `GITHUB_TOKEN` from the mirror workflow links the packages to this repo automatically — including for fork PRs, whose read-only token still carries `packages: read`. For that reason the packages must never be seeded from a personal token.

Deliberately excluded: the `docker://rhysd/actionlint` image in `edr-ci.yml` also comes from Docker Hub, but it's a single digest-pinned pull per run and Renovate tracks its tag+digest — mirroring it would break that bump flow for marginal gain.

## Rollout

No manual steps. This PR touches `mirror-docker-images.yml`, so the mirror job runs on the PR itself and creates the packages; the release-workflow jobs that raced ahead of it on the first run will fail their pulls — re-run failed jobs once the mirror job is green.

## Notes

- Private GHCR storage counts against the org Packages quota (a few GB for these multi-arch images; Actions-download bandwidth is free). If an org admin later flips the packages public, storage becomes free — nothing in this PR depends on visibility either way.
- Adding a node version to the release matrix requires adding the tag to the mirror list first; a missing tag fails loudly with "manifest unknown".

## Verification

- `actionlint` (with `SHELLCHECK_OPTS=--severity=warning`, matching CI): clean on both files.
- `zizmor` 1.26.1 (CI's version, online audits enabled): no findings.

### DIFF
--- a/.github/workflows/edr-npm-release.yml
+++ b/.github/workflows/edr-npm-release.yml
@@ -55,9 +55,10 @@ jobs:
             cache_extra_paths: ~/.cargo/registry/src/
 
           # Container images come from our GHCR mirror (mirror-docker-images.yml),
-          # not Docker Hub, whose pull rate limits made these jobs flaky.
-          # The mirror packages are private, so every pull authenticates with
-          # GITHUB_TOKEN (registry inputs / docker login below).
+          # not Docker Hub, whose pull rate limits made these jobs flaky. New or
+          # changed tags must be added to the mirror workflow's tag list in the
+          # same PR. Release runs pull the official Docker Hub image directly
+          # (see the "Select image source" steps).
           - host: ubuntu-24.04
             target: x86_64-unknown-linux-gnu
             docker: ghcr.io/nomicfoundation/edr/mirror/node:24-bullseye-slim
@@ -152,11 +153,28 @@ jobs:
         run: ${{ matrix.settings.build }}
         shell: bash
 
+      # Release runs build on the official Docker Hub image so the mirror is
+      # never in the supply chain of published binaries; at 1-2 releases a week
+      # its rate limits don't bite there. Everything else uses the mirror.
+      - name: Select image source
+        id: select-image
+        if: ${{ matrix.settings.docker }}
+        shell: bash
+        env:
+          MIRROR_REF: ${{ matrix.settings.docker }}
+          RELEASE_TAG: ${{ needs.check_commit.outputs.tag }}
+        run: |
+          if [ -n "$RELEASE_TAG" ]; then
+            echo "ref=node:${MIRROR_REF##*:}" >> "$GITHUB_OUTPUT"
+          else
+            echo "ref=$MIRROR_REF" >> "$GITHUB_OUTPUT"
+          fi
+
       - name: Build in docker (gnu)
         uses: NomicFoundation/docker-run-action@63f044457cfb71a5c63fa589218c89a418565d9c # Fork of v3 with updated Docker (https://github.com/addnab/docker-run-action/issues/62)
         if: ${{ matrix.settings.docker && matrix.settings.flavor == 'gnu'}}
         with:
-          image: ${{ matrix.settings.docker }}
+          image: ${{ steps.select-image.outputs.ref }}
           registry: ghcr.io
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
@@ -185,7 +203,7 @@ jobs:
         uses: NomicFoundation/docker-run-action@63f044457cfb71a5c63fa589218c89a418565d9c # Fork of v3 with updated Docker (https://github.com/addnab/docker-run-action/issues/62)
         if: ${{ matrix.settings.docker && matrix.settings.flavor == 'musl' }}
         with:
-          image: ${{ matrix.settings.docker }}
+          image: ${{ steps.select-image.outputs.ref }}
           registry: ghcr.io
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
@@ -265,6 +283,7 @@ jobs:
     name: Test bindings on Linux-x64-gnu - node@${{ matrix.node }}
     needs:
       - build
+      - check_commit
     permissions:
       contents: read
       packages: read
@@ -303,11 +322,18 @@ jobs:
       - name: Test bindings
         # Setting CI=1 is important to make PNPM install non-interactive
         # https://github.com/pnpm/pnpm/issues/6615#issuecomment-1656945689
-        run: docker run --rm  -e CI=1 -v "$(pwd)":/build -w /build/crates/edr_napi ghcr.io/nomicfoundation/edr/mirror/node:${{ matrix.node }} bash -c "npm install -g pnpm@11.15.1; pnpm testNoBuild"
+        env:
+          RELEASE_TAG: ${{ needs.check_commit.outputs.tag }}
+        run: |
+          # Releases pull Docker Hub directly; see the build job's "Select image source".
+          IMAGE="ghcr.io/nomicfoundation/edr/mirror/node:${{ matrix.node }}"
+          if [ -n "$RELEASE_TAG" ]; then IMAGE="node:${{ matrix.node }}"; fi
+          docker run --rm -e CI=1 -v "$(pwd)":/build -w /build/crates/edr_napi "$IMAGE" bash -c "npm install -g pnpm@11.15.1; pnpm testNoBuild"
   test-linux-x64-musl-binding:
     name: Test bindings on x86_64-unknown-linux-musl - node@${{ matrix.node }}
     needs:
       - build
+      - check_commit
     permissions:
       contents: read
       packages: read
@@ -344,11 +370,18 @@ jobs:
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
       - name: Test bindings
-        run: docker run --rm  -e CI=1 -v "$(pwd)":/build -w /build/crates/edr_napi ghcr.io/nomicfoundation/edr/mirror/node:${{ matrix.node }}-alpine sh -c "npm install -g pnpm@11.15.1; pnpm testNoBuild"
+        env:
+          RELEASE_TAG: ${{ needs.check_commit.outputs.tag }}
+        run: |
+          # Releases pull Docker Hub directly; see the build job's "Select image source".
+          IMAGE="ghcr.io/nomicfoundation/edr/mirror/node:${{ matrix.node }}-alpine"
+          if [ -n "$RELEASE_TAG" ]; then IMAGE="node:${{ matrix.node }}-alpine"; fi
+          docker run --rm -e CI=1 -v "$(pwd)":/build -w /build/crates/edr_napi "$IMAGE" sh -c "npm install -g pnpm@11.15.1; pnpm testNoBuild"
   test-linux-aarch64-gnu-binding:
     name: Test bindings on aarch64-unknown-linux-gnu - node@${{ matrix.node }}
     needs:
       - build
+      - check_commit
     permissions:
       contents: read
       packages: read
@@ -379,10 +412,21 @@ jobs:
       - name: Install dependencies
         run: |
           pnpm install --frozen-lockfile --prefer-offline --cpu=arm64 --libc=glibc
+      - name: Select image source # releases pull Docker Hub directly; see the build job
+        id: select-image
+        env:
+          MIRROR_REF: ghcr.io/nomicfoundation/edr/mirror/node:${{ matrix.node }}
+          RELEASE_TAG: ${{ needs.check_commit.outputs.tag }}
+        run: |
+          if [ -n "$RELEASE_TAG" ]; then
+            echo "ref=node:${MIRROR_REF##*:}" >> "$GITHUB_OUTPUT"
+          else
+            echo "ref=$MIRROR_REF" >> "$GITHUB_OUTPUT"
+          fi
       - name: Setup and run tests # zizmor: ignore[superfluous-actions] deliberate cross-arch (arm64) test runner
         uses: addnab/docker-run-action@4f65fabd2431ebc8d299f8e5a018d79a769ae185 #v3
         with:
-          image: ghcr.io/nomicfoundation/edr/mirror/node:${{ matrix.node }}
+          image: ${{ steps.select-image.outputs.ref }}
           registry: ghcr.io
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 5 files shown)


.github/workflows/mirror-docker-images.yml
```diff
@@ -0,0 +1,76 @@
...
+              if [ "$attempt" = 3 ]; then
+                echo "Failed to mirror $SOURCE after 3 attempts" >&2
+                exit 1
+              fi
+              sleep $((attempt * 30))
+            done
+            # A faithful copy preserves the manifest digest; a mismatch means
+            # imagetools rewrote the manifest (or upstream moved the tag mid-copy).
+            SRC_DIGEST=$(docker buildx imagetools inspect --format '{{.Manifest.Digest}}' "$SOURCE")
+            DEST_DIGEST=$(docker buildx imagetools inspect --format '{{.Manifest.Digest}}' "$DEST")
```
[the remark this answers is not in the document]
nebasuke: Actions runs `run:` steps with `bash -e {0}` by default, and under `-e` a plain assignment like `SRC_DIGEST=$(...)` exits with the substitution's status — so a failed `inspect` already aborted the step, and the both-digests-empty case wasn't reachable that way. (The silent-continue behavior needs a `local`/`export` prefix or no `-e`.) That said, relying on the implicit default was subtle, so 7f5250abe adds an explicit `set -euo pipefail`.

.github/workflows/edr-npm-release.yml
```diff
@@ -53,24 +54,28 @@ jobs:
             # that caching it isn't worth the size hit.
             cache_extra_paths: ~/.cargo/registry/src/
 
+          # Container images come from our GHCR mirror (mirror-docker-images.yml),
+          # not Docker Hub, whose pull rate limits made these jobs flaky.
+          # The mirror packages are private, so every pull authenticates with
+          # GITHUB_TOKEN (registry inputs / docker login below).
```
anaPerezGhiglia: I would add a comment note here to explicitly remind the developer that if they need to change the image versions in this workflow they need to do it in the mirror workflow together with this file. 
I would prefer to skip the Github_token authentication part from this comment but add the update mirror note to try to keep the comment succint.

book/src/02_development/11_ci_docker_mirror.md
```diff
@@ -0,0 +1,13 @@
+# CI Docker image mirror
+
+CI pulls its Docker images from a GHCR mirror (`ghcr.io/nomicfoundation/edr/mirror/*`) instead of Docker Hub: Docker Hub rate-limits pulls, which fail intermittently on GitHub-hosted runners as a result. The mirror is maintained by `.github/workflows/mirror-docker-images.yml`, which re-copies the images weekly and can be run on demand via `workflow_dispatch`.
+
+## Adding a Node.js version (or any new tag)
+
+Add the tag to the `TAGS` list in `mirror-docker-images.yml` in the same PR that changes the matrix in `edr-npm-release.yml`. The mirror workflow runs on same-repo PRs that touch it, so the new tag is mirrored — and the release matrix testable against it — before merge. A tag referenced in CI but missing from the mirror fails loudly with `manifest unknown`.
```
anaPerezGhiglia, line 9: Let's add a similar note as you did on the PR description
> the mirror job runs on the PR itself and creates the packages; the release-workflow jobs that raced ahead of it on the first run will fail their pulls — re-run failed jobs once the mirror job is green.

book/src/02_development/11_ci_docker_mirror.md
```diff
@@ -0,0 +1,13 @@
+# CI Docker image mirror
+
+CI pulls its Docker images from a GHCR mirror (`ghcr.io/nomicfoundation/edr/mirror/*`) instead of Docker Hub: Docker Hub rate-limits pulls, which fail intermittently on GitHub-hosted runners as a result. The mirror is maintained by `.github/workflows/mirror-docker-images.yml`, which re-copies the images weekly and can be run on demand via `workflow_dispatch`.
```
Wodann: nit: is missing the case for: runs on PRs that modify the `mirror-docker-images.yml`

Wodann (approved):
LGTM. Thanks for the fix!

One quality of life improvement would be for the `edr-npm-release.yml` job to add a step that blocks execution until the `mirror-docker-images.yml` workflow has completed successfully. I employ this technique for the HH3 regression benchmark workflow—it waits for the edr-ci.yml job to succeed before starting.

This would avoid the user from having to re-run CI after modifying the `mirror-docker-images.yml`.

anaPerezGhiglia (changes requested):
I like the intent of decoupling CI from Docker Hub's rate limits, but this change also removes a safety property we currently have: at release time, we build directly on an official Docker Hub image.
The release workflow is the one we most care about protecting, since it produces the binaries that end up executed on users' machines.

With this approach, if a colleague's credentials get stolen (any token with write:packages, or any workflow with packages: write), an attacker could push a compromised image to the mirror under the same tag we build on — and the release flow would happily run on it, potentially ending with us publishing compromised binaries to the npm registry.

The digest check added in `mirror-docker-images.yml` isn't enough for this: it only validates the copies that the mirror job itself publishes. It says nothing about what's in the mirror at the moment the release workflow pulls from it.

I think we should validate the digest at release time, against Docker Hub. Resolving a tag's digest doesn't count as a pull for rate-limiting purposes (the Hub API is separate infrastructure), so this shouldn't reintroduce the flakiness the mirror is solving.

Concretely: each docker job resolves the upstream digest from the Docker Hub API, then pulls the mirror by digest instead of by tag. Pull-by-digest is content-addresse: the runtime verifies every byte against the digest, so a tampered mirror tag simply can't be substituted, and there's no check-then-pull race. The matrix stays as is; each docker job gains one step:

```yaml
  - name: Resolve image digest from Docker Hub
    id: resolve
    if: ${{ matrix.settings.docker }}
    shell: bash
    run: |
      set -euo pipefail
      TAG="${IMAGE##*:}"
      DIGEST=$(curl -sf --retry 3 "https://hub.docker.com/v2/repositories/library/node/tags/${TAG}" | jq -re .digest)
      echo "image=${IMAGE%:*}@${DIGEST}" >> "$GITHUB_OUTPUT"
    env:
      IMAGE: ${{ matrix.settings.docker }}
```

and the image references change to the resolved output:

```diff
  - image: ${{ matrix.settings.docker }}
  + image: ${{ steps.resolve.outputs.image }}
```

The four test-binding jobs follow the same pattern with the tag built from matrix.node. The GHCR login stays as is, pulling a private package by digest still needs auth.

One behavior to document in `11_ci_docker_mirror.md`: if upstream refreshes a tag in place before the weekly mirror re-sync, the resolved digest won't exist in the mirror yet and the pull fails with manifest unknown. That's a benign failure. The fix is re-running the mirror workflow via `workflow_dispatch` and retrying, not debugging the mirror.

book/src/02_development/11_ci_docker_mirror.md
nebasuke: Done!

.github/workflows/edr-npm-release.yml
nebasuke: I added a reference.

book/src/02_development/11_ci_docker_mirror.md
nebasuke: Done!

nebasuke:
> I like the intent of decoupling CI from Docker Hub's rate limits, but this change also removes a safety property we currently have: at release time, we build directly on an official Docker Hub image. The release workflow is the one we most care about protecting, since it produces the binaries that end up executed on users' machines.
> 
> With this approach, if a colleague's credentials get stolen (any token with write:packages, or any workflow with packages: write), an attacker could push a compromised image to the mirror under the same tag we build on — and the release flow would happily run on it, potentially ending with us publishing compromised binaries to the npm registry.
> 
> The digest check added in `mirror-docker-images.yml` isn't enough for this: it only validates the copies that the mirror job itself publishes. It says nothing about what's in the mirror at the moment the release workflow pulls from it.
> 
> I think we should validate the digest at release time, against Docker Hub. Resolving a tag's digest doesn't count as a pull for rate-limiting purposes (the Hub API is separate infrastructure), so this shouldn't reintroduce the flakiness the mirror is solving.
> 
> Concretely: each docker job resolves the upstream digest from the Docker Hub API, then pulls the mirror by digest instead of by tag. Pull-by-digest is content-addresse: the runtime verifies every byte against the digest, so a tampered mirror tag simply can't be substituted, and there's no check-then-pull race. The matrix stays as is; each docker job gains one step:
> 
> ```yaml
>   - name: Resolve image digest from Docker Hub
>     id: resolve
>     if: ${{ matrix.settings.docker }}
>     shell: bash
>     run: |
>       set -euo pipefail
>       TAG="${IMAGE##*:}"
>       DIGEST=$(curl -sf --retry 3 "https://hub.docker.com/v2/repositories/library/node/tags/${TAG}" | jq -re .digest)
>       echo "image=${IMAGE%:*}@${DIGEST}" >> "$GITHUB_OUTPUT"
>     env:
>       IMAGE: ${{ matrix.settings.docker }}
> ```
> 
> and the image references change to the resolved output:
> 
> ```diff
>   - image: ${{ matrix.settings.docker }}
>   + image: ${{ steps.resolve.outputs.image }}
> ```
> 
> The four test-binding jobs follow the same pattern with the tag built from matrix.node. The GHCR login stays as is, pulling a private package by digest still needs auth.
> 
> One behavior to document in `11_ci_docker_mirror.md`: if upstream refreshes a tag in place before the weekly mirror re-sync, the resolved digest won't exist in the mirror yet and the pull fails with manifest unknown. That's a benign failure. The fix is re-running the mirror workflow via `workflow_dispatch` and retrying, not debugging the mirror.

Here I've taken a different approach. Dockerhub mainly fails for us as we pull a decent amount of images, and there's a limit of 100 pulls per 6 hours. See https://docs.docker.com/docker-hub/usage/. That should be more than fine if we use it JUST for release. So what I've done instead of pulling by digest, I've instead let actual release (those with a tag) use the original images on Dockerhub rather than those that are cached.

nebasuke:
> LGTM. Thanks for the fix!
> 
> One quality of life improvement would be for the `edr-npm-release.yml` job to add a step that blocks execution until the `mirror-docker-images.yml` workflow has completed successfully. I employ this technique for the HH3 regression benchmark workflow—it waits for the edr-ci.yml job to succeed before starting.
> 
> This would avoid the user from having to re-run CI after modifying the `mirror-docker-images.yml`.

Sure, let me tackle that as a follow-up!

.github/workflows/edr-npm-release.yml
```diff
@@ -133,25 +138,36 @@ jobs:
...
       - name: Build in docker (gnu)
         uses: NomicFoundation/docker-run-action@63f044457cfb71a5c63fa589218c89a418565d9c # Fork of v3 with updated Docker (https://github.com/addnab/docker-run-action/issues/62)
-        if: ${{ matrix.settings.docker && matrix.settings.flavor == 'gnu'}}
+        if: ${{ matrix.settings.docker_tag && matrix.settings.flavor == 'gnu'}}
         with:
-          image: ${{ matrix.settings.docker }}
+          image: ${{ steps.select-image.outputs.ref }}
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
```
anaPerezGhiglia, line 171: At first glance I thought this was wrong. Then, thanks to Claude, I understood that the registry and credentials are not used it the image ref points to Docker hub. 
Perhaps we can add a comment on these lines so that's clear for the reader that the login is unused if the image declares no registry. Ideally we would not even send this arguments, but that might be to cumbersome to achieve, so I think a comment will be enough

anaPerezGhiglia (approved):
Thanks for tackling the supply chain risk, I like your proposal of using official hub on _real_ releases better.

I jus left a small suggestion.
Also, noting here to gain awareness, not not to tackle in this PR: the trick of bypassing steps with the `tag` output is growing too much and I fear it because I think it's too easy to make a mistake with it. I think we should re-think this workflow and gate that in a more robust way.

.github/workflows/edr-npm-release.yml
nebasuke: Thanks for the suggestions, I added comments!

nebasuke:
> Thanks for tackling the supply chain risk, I like your proposal of using official hub on _real_ releases better.
> 
> I jus left a small suggestion. Also, noting here to gain awareness, not not to tackle in this PR: the trick of bypassing steps with the `tag` output is growing too much and I fear it because I think it's too easy to make a mistake with it. I think we should re-think this workflow and gate that in a more robust way.

Agreed... I'll see if I can combine it with @Wodann's follow-up. Filed a quick AI assisted issue here: https://github.com/NomicFoundation/edr/issues/1593. Will follow-up with the fixes soon.

PR #1589, opened 2026-07-29
ci: update HH3 regression benchmark default filter

Wodann:
https://github.com/NomicFoundation/hardhat/pull/8467 adds new benchmark scenarios.

This PR updates our default benchmarks filter to include these:
- renamed `mocha test` to `test mocha` to match naming pattern
- added `test vitest` to include new type of test runner


popescuoctavian (approved):
LGTM, thanks!

PR #1594, opened 2026-07-31
chore(edr_tool_cli): remove the broken `gen-execution-api` subcommand

### LINKED ISSUE #1570: `edr_tool_cli`: `gen-execution-api` targets a crate deleted in 2024 — remove it?
## What it is

`cargo run -p edr_tool_cli -- gen-execution-api` downloads `refs-openrpc.json` from `ethereum/execution-apis` and runs `npx open-rpc-typings` to regenerate Rust typings into `crates/eth_execution_api/src`.

## Why it can't work

Found while adding the `gen-solx-fixtures` subcommand (#1552). The command is broken twice over:

1. **Its target crate no longer exists.** `crates/eth_execution_api` was deleted in #460 ("ci: remove synthetic benchmarks", May 2024). The subcommand starts by reading that crate's `Cargo.toml` for the execution-apis version to fetch, so it fails immediately — and has for over two years. Nothing has touched `execution_api.rs` since except mechanical dependency bumps.
2. **Its path resolution broke separately.** #1139 (Nov 2025) moved the CLI crate from `crates/tools/` to `crates/tool/cli/` without updating `update::project_root()`, which has resolved to `crates/` instead of the repo root since. This half is already fixed in #1552 because `gen-solx-fixtures` shares the helper — but it also means the breakage in (1) was doubly masked.

Smaller signs of rot: `generate(_mode: Mode)` ignores its mode argument, so the `Verify` variant (a CI staleness check by design) was never wired up here.

## Proposal

Delete the subcommand: `execution_api.rs`, the `GenExecutionApi` variant/arm in `main.rs`, and the four dependencies only it uses (`reqwest`, `toml`, `tempfile`, `cfg-if`). `update.rs` stays — `gen-solx-fixtures` uses it.

Alternative, if generated execution-API typings are wanted again someday: resurrecting the consumer crate is a prerequisite anyway, and the generator could be re-added alongside it in whatever form fits then. Keeping dead tooling around until that day costs a compile-time dependency tree (reqwest in particular) and misleads readers into thinking the flow works.

Happy to open a small PR for the removal if the team agrees.

nebasuke:
Closes #1570.

`cargo run -p edr_tool_cli -- gen-execution-api` (and its `cargo gen-execution-api`
alias) downloaded `refs-openrpc.json` from `ethereum/execution-apis` and ran
`npx open-rpc-typings` to regenerate Rust typings into `crates/eth_execution_api/src`.

That crate was deleted in #460 ("ci: remove synthetic benchmarks", May 2024). The
subcommand's first step is to read the crate's `Cargo.toml` for the execution-apis
version to fetch, so it has failed at the first line for over two years. Since then
only mechanical dependency bumps have touched `execution_api.rs`.

The breakage was doubly masked: #1139 moved the CLI crate from `crates/tools/` to
`crates/tool/cli/` without updating `update::project_root()`, so path resolution was
broken independently. That half is already fixed on `main` (#1577), which is why this
PR is only the removal.

## What's removed

- `crates/tool/cli/src/execution_api.rs`
- the `GenExecutionApi` variant and match arm in `main.rs`, plus the now-unused
  `use update::Mode;`
- the `gen-execution-api` alias in `.cargo/config.toml`
- the four dependencies only `execution_api.rs` used: `reqwest`, `toml`, `tempfile`,
  `cfg-if`
- `update::reformat()` — same generator-era leftover, zero callers, and its
  generated-file preamble still pointed at the long-gone `crates/tools` path

Also dropped the stale `#[allow(unused)]` on `update::update()`, which is genuinely
used by `gen-solx-fixtures`. The rest of `update.rs` stays for that subcommand's sake;
`Mode::Verify` keeps its `#[allow(unused)]` since nothing wires up a staleness check
yet.

## On the dependency win

The issue claimed this saves a compile-time dependency tree, "reqwest in particular."
Checked, and that part is wrong — worth stating plainly so nobody re-derives it later:

- `edr_rpc_client` already declares `reqwest = { version = "0.13", features = ["blocking", "json"] }`
  with default features, so the workspace's resolved reqwest feature set is unchanged
  by this PR (verified with `cargo tree -e features -i reqwest@0.13.4 --workspace`
  before and after: identical).
- `toml`, `tempfile`, and `cfg-if` are likewise pulled in by other crates.

`Cargo.lock` therefore changes by exactly four removed entries inside the
`edr_tool_cli` block, with no version churn.

The real win is smaller and non-perf: dead code and a broken command gone, one less
non-workspace `reqwest` version declaration drifting from the workspace pin, and a
smaller Renovate/audit surface on this crate.

## Verification

- `cargo clippy --locked -p edr_tool_cli --all-targets` — clean
- `cargo fmt -p edr_tool_cli -- --check` — clean
- `cargo build --locked --release --bin edr_tool_cli` — reproduces the
  `link-standalone-bins` CI job (this crate links napi with `dyn-symbols` and no Node
  host, so it's the one build that can catch a link regression here)
- `Cargo.lock` diff is deletions only; no new or bumped versions, so nothing for
  cargo-cooldown-check to flag

If generated execution-API typings are ever wanted again, resurrecting the consumer
crate is a prerequisite anyway, and the generator can come back alongside it in
whatever form fits then.


anaPerezGhiglia (approved):
🧹 LGTM

PR #1601, opened 2026-08-04
refactor: introduce EDR-owned protocol hardfork types

anaPerezGhiglia:
This refactor introduces EDR-owned hardfork types `L1Hardfork` and `OpHardfork`, so that EDR's business logic is ruled by its own vocabulary, converted to revm's types only at the EVM boundary. Beyond decoupling EDR from revm's `SpecId` changes, owning the types gives EDR a home for protocol semantics of its own.

## Motivation

EDR must be protocol compliant: it cares about **all** hardforks, including ones with no EVM-semantics changes. Starting with revm `39.0.0` (release `v108`, the "SpecId cleanup" in revm's migration guide), `SpecId` models EVM behavior classes only: the release removed the variants for Frontier Thawing, DAO Fork, Constantinople, Muir Glacier, Arrow Glacier, and Gray Glacier as "EVM-equivalent", and shifted the discriminants of the remaining ones. EDR currently pins revm `38.0.0`, the last version that still models every hardfork; upgrading past it while EDR's protocol logic is keyed on `SpecId` would lose information EDR relies on, because EVM-equivalent is not header-equivalent:

This PR therefore lands **before** the pending revm upgrade (#1590, on hold): against the current pin the conversions introduced here are lossless 1:1, so the refactor is behavior-preserving by construction, and the upgrade's `SpecId` changes will then only touch the `From` conversion impls.

### Owned enums
One enum per chain type, listing every protocol upgrade in activation order. Left SCREAMING variant names keep the existing use sites textually identical.

### Trait layer(`edr_chain_spec`)
Chain-generic code needs to name both vocabularies: `HardforkChainSpec` now carries `EvmHardfork` (what revm consumes: `SpecId` for L1/generic, `OpSpecId` for OP) alongside the protocol-level `Hardfork`. `ProtocolHardfork` bundles the bounds generic code needs and provides `to_evm_spec_id()` for gates that are genuinely about EVM behavior.

### Protocol parameters re-homed
The difficulty bomb delay and block reward are protocol-level parameters, not EVM semantics. Muir Glacier (EIP-2384), Arrow Glacier (EIP-4345), and Gray Glacier (EIP-5133) consisted of nothing but a bomb-delay change, precisely why `SpecId` collapsed them.  EDR must replay protocol history, so these parameters belong on its own hardfork vocabulary.`bomb_delay` and `miner_reward` were free functions keyed on revm's `SpecId`, exactly the coupling that breaks when revm collapses protocol-only forks. They now live on the owned enums as `ProtocolParams` impls (tables moved verbatim).

### Where the design is deliberately not independent yet
Some hardfork gates in chain-generic code still ask revm's EVM classes (`hardfork.to_evm_spec_id() >= EvmSpecId::X`): `ProtocolParams` doesn't answer these gate questions yet, and `L1Hardfork`/`OpHardfork` are deliberately not comparable to each other, so generic code has nothing chain-agnostic to gate on. The intended end state extends the `bomb_delay`/`miner_reward` pattern into a semantic API — instead of `hardfork >= LONDON`, ask something like `hardfork.supports_eip1559_gas_fee()`, but that was consciously left out to scope this PR to the critical changes (see follow-ups). The remaining gates are safe: they are genuinely EVM behavior-class questions, which survive `SpecId` cleanups.

### Solidity test runner. 
To keep the vocabulary consistent across all of EDR's interfaces, the runner's config ingress also parses the owned types (`TestRunnerConfig::try_into_runner_config::<ChainSpecT>()`), converting to the revm type immediately after. The runner itself stays in revm vocabulary: it has only a few hardfork gates, all tied to EVM-semantics forks, and refactoring the vendored foundry port would diverge the backport further without being a blocker for this change.

### One vocabulary per layer
Each layer speaks its own hardfork vocabulary and converts it into the types of the layer below when crossing the seam: the TS/JS surface speaks JS enums and strings, the Rust protocol internals speak the owned types, and only EVM execution speaks revm's behavior classes.

layer-stack ignoreme


crates/edr_chain_l1/src/hardfork.rs
```diff
@@ -0,0 +1,399 @@
...
+    pub const PRAGUE: &str = "Prague";
+    /// String identifier for the Osaka hardfork
+    pub const OSAKA: &str = "Osaka";
+    /// String identifier for the Amsterdam hardfork
+    pub const AMSTERDAM: &str = "Amsterdam";
+    /// String identifier for the latest hardfork
+    pub const LATEST: &str = "Latest";
+}
+
+impl FromStr for L1Hardfork {
```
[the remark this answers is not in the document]
anaPerezGhiglia: This is intentional: the name module is a verbatim copy of revm-primitives, and revm's own `SpecId::from_str` also rejects `Latest`. We mirror both so hardfork strings parse identically before and after this refactor (there's a test pinning the rejection). The const has no consumers, so if it reads as misleading we can drop it in a follow-up

crates/edr_chain_spec/src/lib.rs
```diff
@@ -107,10 +107,41 @@ pub trait ContextChainSpec {
...
+
+/// Capabilities required of a chain's protocol-level hardfork type by code
+/// that creates blocks: conversion to the EVM behavior class, activation
+/// ordering, and protocol-level parameters.
+pub trait ProtocolHardfork: Copy + Into<EvmSpecId> + PartialOrd + ProtocolParams {
+    /// Converts the hardfork into the EVM specification identifier of its EVM
+    /// behavior class.
+    fn to_evm_spec_id(self) -> EvmSpecId {
+        self.into()
+    }
```
Wodann: Suggestion: Since we are defining the hardfork types ourselves, `Into<EvmSpecId>` can be used. The only caveat is that the type system can get confused when there are two possible `Into` traits implemented for the source type. This can be avoided by explicitly annotating the type. I did that in [this commit](https://github.com/NomicFoundation/edr/commit/261e4c5cc52a10ad00345df580fcff7c5ae1cdfa) that can be cherry-picked.

This also voids the need for `trait ProtocolHardfork` in general, as there is only one usage site left.

crates/edr_chain_spec/src/lib.rs
```diff
@@ -107,10 +107,41 @@ pub trait ContextChainSpec {
...
+    /// The EVM-level hardfork type consumed by revm for this chain; e.g.
+    /// `SpecId` for L1, `OpSpecId` for OP.
+    ///
+    /// Unlike [`Self::Hardfork`], this models EVM behavior classes rather than
+    /// protocol upgrades.
+    type EvmHardfork: Copy + Into<EvmSpecId>;
+
+    /// The chain's protocol-level hardfork type.
+    type Hardfork: Default + Into<Self::EvmHardfork> + ProtocolHardfork;
+}
```
Wodann, line 122: Suggestion: to keep things maximally composable and avoid overconstraining type bounds, we've been moving to "minimal" traits, where we each trait only specifies the types and/or functions they themselves require as well as the sub-traits that are required for their own definition.

Proposed change cherry-pickable from https://github.com/NomicFoundation/edr/commit/54bda97f794d4968ca5238738757d11ea4c35700#diff-bc45d8eec77a969161b951f574904bd0b2ad4aa86406075f796a13d996d9a12e

This allows you to only require `EvmHardforkChainSpec` in the parts of the code that deal with `EvmHardfork` alone, making the code maximally reusable.

I added docs to the book in [this PR](https://github.com/NomicFoundation/edr/pull/1618) to outline the best practices I used for the multi-chain architecture and their rationale.

crates/edr_chain_l1/src/hardfork.rs
```diff
@@ -0,0 +1,399 @@
...
+
+use edr_chain_spec::{EvmSpecId, ProtocolParams};
+use edr_primitives::UnknownHardfork;
+
+/// Ethereum L1 hardfork.
+///
+/// Models protocol upgrades, including ones without EVM-semantics changes,
+/// unlike [`EvmSpecId`] which models EVM behavior classes.
+#[repr(u8)]
+#[allow(non_camel_case_types)]
```
Wodann, line 14: style: Since this is now within our control, I suggest applying a single reviewable commit that removes the `#[allow(non_camel_case_types)]` and converts all entries to `RustEnumCase`.

crates/edr_chain_l1/src/hardfork.rs
```diff
@@ -0,0 +1,399 @@
...
+    Eq,
+    PartialOrd,
+    Ord,
+    Hash,
+    serde::Serialize,
+    serde::Deserialize,
+)]
+pub enum L1Hardfork {
+    /// Frontier hardfork
+    FRONTIER = 0,
```
Wodann, line 35: suggestion: We are now in control of hardforks, so can use this as an opportunity to remove unsupported hardforks at the type level.

Cherry-pickable changes: https://github.com/NomicFoundation/edr/commit/54eca329dee4f5398abeb83a224a2dfd36fbe718

crates/edr_chain_l1/src/hardfork.rs
```diff
@@ -0,0 +1,399 @@
...
+use edr_chain_spec::{EvmSpecId, ProtocolParams};
+use edr_primitives::UnknownHardfork;
+
+/// Ethereum L1 hardfork.
+///
+/// Models protocol upgrades, including ones without EVM-semantics changes,
+/// unlike [`EvmSpecId`] which models EVM behavior classes.
+#[repr(u8)]
+#[allow(non_camel_case_types)]
+#[derive(
```
Wodann, line 15: suggestion: derive `Display` and `FromStr` using [strum](https://crates.io/crates/strum)'s `Display` and `EnumStrings`.

The `Display` will allow us to generate better error message than we do now. REVM types don't implement `Display`, which requires us to depend on its `Debug` implementation—which is more verbose but less user-friendly.

crates/edr_chain_l1/src/hardfork.rs
```diff
@@ -0,0 +1,399 @@
...
+            L1Hardfork::OSAKA => EvmSpecId::OSAKA,
+            L1Hardfork::AMSTERDAM => EvmSpecId::AMSTERDAM,
+        }
+    }
+}
+
+/// String identifiers for L1 hardforks.
+pub mod name {
+    /// String identifier for the Frontier hardfork
+    pub const FRONTIER: &str = "Frontier";
```
Wodann, line 134: suggestion: This is an opportunity to align with Hardhat's naming convention for hardfork names, rather than sticking with REVM's mix-and-match scheme.

This would be a breaking change in EDR's API but is encapsulated within Hardhat. [This](https://github.com/NomicFoundation/hardhat/blob/main/packages/hardhat/src/internal/builtin-plugins/network-manager/edr/types/hardfork.ts#L5) is Hardhat's enum for `L1HardforkName`s and `OpHardforkName`s. Conversion logic lives [here](https://github.com/NomicFoundation/hardhat/blob/main/packages/hardhat/src/internal/builtin-plugins/network-manager/edr/utils/convert-to-edr.ts#L74).

Hardhat 2 uses a similar [conversion function](https://github.com/NomicFoundation/hardhat/tree/v2/packages/hardhat-core/src/internal#L46). Names for their hardfork names (L1 only) live [here](https://github.com/NomicFoundation/hardhat/blob/v2/packages/hardhat-core/src/internal/util/hardforks.ts#L8).

crates/edr_op/src/hardfork.rs
```diff
@@ -13,6 +13,156 @@ pub mod generated;
 /// OP chain configs
 pub mod op;
 
+/// OP Stack hardfork.
+///
+/// Models protocol upgrades, including ones without EVM-semantics changes,
+/// unlike [`op_revm::OpSpecId`] which models EVM behavior classes.
+#[repr(u8)]
+#[allow(non_camel_case_types)]
```
Wodann, line 24: suggestion: same two suggestions regarding `strum` and naming convention as for `L1Hardfork`: https://github.com/NomicFoundation/edr/pull/1601/changes#r3759869382

crates/edr_generic/src/rpc/receipt.rs
```diff
@@ -97,7 +97,7 @@ impl RpcTypeFrom<L1BlockReceipt<TypedEnvelope<edr_receipt::Execution<FilterLog>>
         value: &L1BlockReceipt<TypedEnvelope<edr_receipt::Execution<FilterLog>>>,
         hardfork: Self::Hardfork,
     ) -> Self {
-        let transaction_type = if hardfork >= edr_chain_l1::Hardfork::BERLIN {
+        let transaction_type = if hardfork >= Self::Hardfork::BERLIN {
```
Wodann: nit: import `Hardfork`

crates/edr_chain_spec/src/lib.rs
```diff
@@ -107,10 +107,41 @@ pub trait ContextChainSpec {
...
+
+/// Trait for protocol-level parameters that are determined by the hardfork but
+/// are not derivable from its EVM behavior class.
+pub trait ProtocolParams {
```
Wodann: suggestion: This is introducing a trait for all chain specs but it only really makes sense for L1 (and generic), as those contain pre-merge hardforks (L2 didn't exist until post-merge). This is evident in the OP implementation, as it throws a panic and returns `None` for the miner reward.

I think the abstraction layer can be one layer higher:

1. when determining the block's default `difficulty`. Cherry-pickable change [here](https://github.com/NomicFoundation/edr/commit/e9bd1afe6ae32f4501ea1f64831009ff860f3fab).
2. when finalising the block header, let an `L1BlockBuilder` add rewards and leave out rewards in the `OpBlockBuilder`. All required data is already available inside the respective block builders to do so. Cherry-pickable change [here](https://github.com/NomicFoundation/edr/commit/c7fb63101e5cb501d938582863a5fc0b474d90bc)

The two changes make `ProtocolParams` obsolete, so it can be removed.

crates/edr_chain_spec/src/lib.rs
```diff
@@ -107,10 +107,41 @@ pub trait ContextChainSpec {
...
+    fn bomb_delay(self) -> u64;
+
+    /// Returns the static block reward for the hardfork, or `None` post-merge.
+    fn miner_reward(self) -> Option<u128>;
+}
+
+/// Capabilities required of a chain's protocol-level hardfork type by code
+/// that creates blocks: conversion to the EVM behavior class, activation
+/// ordering, and protocol-level parameters.
+pub trait ProtocolHardfork: Copy + Into<EvmSpecId> + PartialOrd + ProtocolParams {
```
Wodann: Suggestion: The `Copy` trait bound was only required for the old `Hardfork` type because IIRC REVM originally required `Copy` for its `HardforkT`.

The `Copy` trait is more restrictive than `Clone` (`trait Copy: Clone`), allowing implicit copying of `Clone`able types. Usually, it's only implemented for types up to 64-bits. As such, `Clone` is common but `Copy` is not. We should avoid unnecessarily restricting users of our API and opt for `Clone` since we have the choice.

After checking REVM, I learned that they (no longer) require the `Copy` trait bound for `HardforkT`. As such we can also slacken the trait bound from `Copy` -> `Clone` for both the EVM & protocol hardfork types. I've done that in [this cherry-pickable commit](https://github.com/NomicFoundation/edr/commit/f22e4e2563b2497a66bc8168abb8c36d416e1c6a).

Wodann (changes requested):

crates/edr_chain_l1/src/hardfork.rs
anaPerezGhiglia: I agree. I decided to leave it out of the scope of this PR to avoid changing all the caller places and make this PR even bigger. I was thinking on doing this as a followup PR that only does the rename so its easier to review.

crates/edr_chain_l1/src/hardfork.rs
anaPerezGhiglia: This is great! However I'm not sure about including these changes in this PR.  What do you think about having this as a follow-up PR that's based on top of these changes? Similarly to what I did for #1603. That way we make sure this PR changes are lossless, a 1:1 replacement, so the refactor is behavior-preserving by construction.

crates/edr_chain_l1/src/hardfork.rs
anaPerezGhiglia: similar to previous comment. I was tempted to do this, but I focused on making this refactor a 1:1 replacement and leave all other improvements for followup PRs.

crates/edr_chain_l1/src/hardfork.rs
anaPerezGhiglia: Applied in [7ddfe9e](https://github.com/NomicFoundation/edr/pull/1601/commits/7ddfe9eb3da995ec3e50c975803f15e18bf68e52): Display/FromStr are now strum-derived with per-variant serialize attrs. 

However, regarting the motivation behind adopting the change: this PR's owned types already implemented `Display` manually for both chains. So the derive is a boilerplate reduction rather than new capability.

crates/edr_chain_spec/src/lib.rs
anaPerezGhiglia: Love it, thanks for the commit! I've already applied it in this branch

crates/edr_op/src/hardfork.rs
anaPerezGhiglia: Applied in [7ddfe9e](https://github.com/NomicFoundation/edr/pull/1601/commits/7ddfe9eb3da995ec3e50c975803f15e18bf68e52)

crates/edr_chain_spec/src/lib.rs
anaPerezGhiglia: isn't it worth to have the umbrella trait defined to avoid having to repeat the `: Clone + Into<EvmSpecId> + PartialOrd` bounds every time? I originally had created it exactly for that purpose, then I ended up creating the `to_evm_spec_id ` function (which I agree we can get rid off)

crates/edr_generic/src/rpc/receipt.rs
anaPerezGhiglia: Addressed by [637f537](https://github.com/NomicFoundation/edr/pull/1601/commits/637f53727b22d7f638b4477f5a52eaab13122238)

crates/edr_chain_spec/src/lib.rs
anaPerezGhiglia: I cherry-picked both commits. On the difficulty one I kept the named `ProtocolHardfork` bounds so the commit stays scoped to the difficulty move. Whether to inline the bounds and drop the `to_evm_spec_id` sugar is being discussed separately in https://github.com/NomicFoundation/edr/pull/1601#discussion_r3778737519, and can land as its own commit if that's where we end up.

crates/edr_chain_spec/src/lib.rs
anaPerezGhiglia: Cherry-picked in the commit with one deviation: I removed the `to_evm_spec_id` sugar method as suggested, but kept the `ProtocolHardfork` trait itself, reduced to a blanket-implemented alias for its bounds. Here the commit in this branch: [c7a1fdf](https://github.com/NomicFoundation/edr/pull/1601/commits/c7a1fdffa7c4f83f3b90e88f6b7bf3dc0e66369a)

The "only one usage site left" holds on your branch, where the ["difficulty" commit](https://github.com/NomicFoundation/edr/commit/e9bd1afe6ae32f4501ea1f64831009ff860f3fab) inlined the bounds (check for the changes in `crates/blockchain/fork/src/lib.rs` for example), but on this branch there are thirteen more: the hardfork-generic functions in `block/header`, `block/local`, `block/storage`, `blockchain/local` and `blockchain/fork`. Those crates have no `ChainSpecT` in scope, so the bounds can't ride on `ProtocolHardforkChainSpec::ProtocolHardfork` there, it's either repeating the bound list at each site or naming the bundle once. The blanket-implemented alias trait is a recognized solution for this type of situation, it's the same shape as foundry's `HardforkTr`, which bundles the hardfork bounds for the solidity-test stack.

crates/edr_chain_spec/src/lib.rs
anaPerezGhiglia: I'd lean towards keeping `Copy`. Some context I gathered while thinking about this:

This isn't a restriction this PR introduced at the API level: `HardforkChainSpec` on `main` already required `Copy` on the hardfork associated type. What `main` had was looser bounds (`Clone`) one layer deeper, in the chain-spec-agnostic crates like `block/header`, but any hardfork type flowing through the system was already forced to be `Copy` by the trait above. 
The PR aligned those layers with the existing requirement, which also cleaned the `.clone()` calls. Keeping the umbrella trait also makes that alignment visible: `ProtocolHardforkChainSpec`'s associated type is now literally "the deeper layers' hardfork contract (`ProtocolHardfork`) plus `Default` and `Into<Self::EvmHardfork>`", so the two layers can't silently drift apart again.

One thing I noticed checking the revm history: `Cfg::Spec` didn't go from `Copy` to `Clone`, it had no capability bound originally, and `Clone` was added in [#2067](https://github.com/bluealloy/revm/pull/2067) as the minimum their internals need (their own `SpecId` stay `Copy`).

The one thing `Clone` bounds would genuinely enable is consuming the block crates standalone, outside the chain-spec stack, with a non-`Copy` hardfork type. But we don't publish EDR as a Rust library today, and I'm not sure that case pays for scattering `.clone()` through the codebase now. I'd rather keep the layers consistent and loosen the bound if a concrete need shows up down the road.

crates/edr_chain_l1/src/hardfork.rs
Wodann: Either way works. Up to you.

Wodann: To add to my prior message, I do think it's important to land this as soon as possible, while we're still working on this.

If you want to cut a release without the breaking change and do another release right afterwards, that's fine by me. I'd do both at once to save time (and do the changes on the Hardhat 3 side myself at the same time).

crates/edr_chain_l1/src/hardfork.rs
Wodann: That works for me.

I have the same thought on it as I shared here: https://github.com/NomicFoundation/edr/pull/1601#discussion_r3786141044

crates/edr_chain_spec/src/lib.rs
Wodann: As I mentioned in my original comment, we indeed previously had the `Copy` trait bound on the chain spec's associated type. We didn't have it at (most of) the individual function trait bounds, though. In that sense, this PR is unnecessarily tightening the trait bounds. There is no reason for us to require `Copy`; merely `Clone`.

From the Rust docs:

> What’s the difference between Copy and Clone?
> 
> Copies happen implicitly, for example as part of an assignment y = x. The behavior of Copy is not overloadable; it is always a simple bit-wise copy.

We don't need the `Hardfork` type to be copied implicitly, nor bit-wise copyable. Therefore, I stand by keeping the `Clone` trait bound.

That also means that the `ChainSpecT::Hardfork` trait bound of `Copy` was too restrictive. I don't exactly recall why I required it originally. As I mentioned, I thought it was due to a trait bound in REVM. Either way, it's no longer necessary.

If we are making changes, the right direction would be to slacken the trait bound from `Copy` -> `Clone` rather than going the opposite way and making all code require `Copy`.

crates/edr_chain_spec/src/lib.rs
Wodann: The proposed removal of `ProtocolHardfork` was probably driven by me first reinstating the `Clone` trait bounds to the original pre-PR values. After that there was only one instance of `ProtocolHardfork` left.

I'm happy to keep the umbrella trait, but I think only `Clone + Into<EvmSpecId>` are required for the type. Only one place requires `PartialOrd` which should be specific to that area. Otherwise we are over-constraining all other APIs/libraries/crates.

For the `Clone` discussion, see https://github.com/NomicFoundation/edr/pull/1601#discussion_r3786218552

Wodann (commented):
I left comments on the remaining open discussions. I think there are two more kinks to agree on.

crates/edr_chain_spec/src/lib.rs
anaPerezGhiglia: I suspected it wasn't just one place that's simple to bound: the chain of `PartialOrd` consumers impacts more layers than the direct comparison sites, but I implemented the suggestion to validate it to be sure. The result: the minimal version needs `+ PartialOrd` at all 13 `ProtocolHardfork` bound sites plus the chain-spec associated type, which are all the consumers of the trait. Removing it from any single one fails to compile. So after the split, every occurrence of the trait reads `ProtocolHardfork + PartialOrd`, and nothing is actually less constrained, no consumer exists that wants a hardfork without its ordering.

Which matches the semantics IMO: hardforks are defined by their activation order: activation tables, base-fee schedules, and hardfork gates are built on it. It only looks narrowly used because most gates still compare via `EvmSpecId`. As gates migrate to the owned vocabulary eventually, this ordering is what they'll be built on.

WRT `Clone`, I'll reply in the other thread to keep it separated

Wodann: If that's what you found, that's a valid usage of the rule:

> - type bounds that are required for the trait's definition OR **are always required at usage sites**

source: https://github.com/NomicFoundation/edr/pull/1618

so feel free to include `PartialOrd`.

crates/edr_chain_spec/src/lib.rs
anaPerezGhiglia: Bound relaxed in [68ed4c2](https://github.com/NomicFoundation/edr/pull/1601/commits/68ed4c2e7d944410fc2a7945833fa5633300ae16). 

Notes: 
- restoring `Clone` at the function bounds matches `main`
- dropping `Copy` from the chain-spec associated type goes further than `main` which relies on it for implicit copies in all chain-spec-generic code, so this adds ~15 explicit `.clone()` calls `main` doesn't have

Fine by me considering your comment that the assoc-type `Copy` is unnecessary, just making it explicit.

crates/edr_chain_l1/src/hardfork.rs
anaPerezGhiglia: tracked by #1634

crates/edr_chain_l1/src/hardfork.rs
anaPerezGhiglia: tracked by #1633

crates/edr_chain_l1/src/hardfork.rs
anaPerezGhiglia: tracked by #1633

anaPerezGhiglia:
@Wodann I left the umbrella `ProtocolHardfork` trait, but relaxed the type bound form `Copy` -> `Clone`. For the out-of-scope comments I created follow-up issues. This is ready for a re-review

Wodann (approved):
Thanks for addressing all of the feedback. LGTM!

Issue #434, opened 2024-05-06
Publish all napi-rs builds with `@nomicfoundation/edr`


fvictorio:
Due to [an npm bug](https://github.com/npm/cli/issues/4828) that doesn't seem like it's going to be fixed any time soon, users can have issues with the way NAPI-RS uses optional dependencies.

As a workaround, we can make the `@nomicfoundation/edr` package always have all the builds available.

It's not clear what's the best way to do this: we could either include all the builds directly in that package, or keep the current approach but make the dependencies be actual dependencies and not optional dependencies. We'll need to do some research.

fvictorio:
Closed by https://github.com/NomicFoundation/edr/pull/449

PR #1615, opened 2026-08-11
build(npm): redefine platform packages as `optionalDependencies`

### LINKED ISSUE #1576: [npm module] Define platform-specific packages as optionalDependencies again
## Context

As a workaround for npm/cli#4828, #434 made the `@nomicfoundation/edr` platform-specific packages regular `dependencies`. See `[scripts/prepublish.sh`](https://github.com/NomicFoundation/edr/blob/main/scripts/prepublish.sh) workaround of renaming the `optionalDependencies` key.

 npm/cli#4828 has since been fixed by npm/cli#8184, released in npm 11.3.0 (2025-04-08).

 We could now roll back to `optionalDependencies` so users only download the build for their own platform instead of all of them:

Image

## Tradeoff to consider
The npm fix only ships in `npm ≥ 11.3.0`, and Node 22 (still LTS until April 2027) bundles `npm 10.x`, which never got the backport. 
However, affected users wouldn't be left out: npm 11 supports Node `^20.17.0 || >=22.9.0`, so they can get the fix by updating npc cli `npm install -g npm@11` while staying on Node 22.

anaPerezGhiglia:
closes #1576

## Changes

- `scripts/prepublish.sh`: removed the `optionalDependencies` → `dependencies` rename, so the platform-specific `@nomicfoundation/edr-*` packages are published as `optionalDependencies` again and installs only
  download the build for the current platform. 
   - The rename was a workaround (#434) for npm/cli#4828
   - npm fixed it in version `11.3.0`.
- `scripts/publish_to_verdaccio.sh`: now runs `prepublish.sh` instead of hand-wiring a single platform dependency, keeping the local publish flow as close as possible to the real release. The entries for platforms not published locally are left in place: npm and pnpm skip `optionalDependencies` they can't resolve.

## Validations

Ran the publish flow end to end against a local Verdaccio 6 registry  on `linux-arm64-gnu`:

  - **Publish**: `scripts/publish_to_verdaccio.sh --version 0.15.0-local.<sha>` — runs the same `prepublish.sh` as the release workflow and published both the platform package and the root package. 
  - **Published shape**: `npm view` on the registry shows all 7 platform packages under `optionalDependencies` and no `dependencies` key.
  - **Consumer installs**: fresh projects installing from the registry with npm and pnpm each downloaded only `@nomicfoundation/edr` plus the current platform's package.
  - **TypeScript client**: a `tsc` project that at runtime constructed an `EdrContext` and  registered the generic provider factory, confirming the native binding loads through the optional platform package.

### DIFF
--- a/.github/workflows/edr-npm-release.yml
+++ b/.github/workflows/edr-npm-release.yml
@@ -542,10 +542,6 @@ jobs:
         uses: ./.github/actions/setup-node
       - name: Install dependencies
         run: pnpm install --frozen-lockfile --prefer-offline
-      - name: Install sponge # needed for prepublish script
-        run: |
-          sudo apt-get update
-          sudo apt-get install -y moreutils
       - name: Download all artifacts
         uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
         with:
--- a/scripts/prepublish.sh
+++ b/scripts/prepublish.sh
@@ -12,5 +12,3 @@ set -o pipefail
 #   --skip-optional-publish  — don't run `npm publish` for the platform pkgs.
 #   --no-gh-release          — don't try to create a GitHub release.
 pnpm napi pre-publish -t npm --skip-optional-publish --no-gh-release
-
-jq 'with_entries(if .key == "optionalDependencies" then .key = "dependencies" else . end)' package.json | sponge package.json
--- a/.changeset/ten-lines-sort.md
+++ b/.changeset/ten-lines-sort.md
@@ -1,0 +1,5 @@
+---
+"@nomicfoundation/edr": minor
+---
+
+Changed the platform-specific `@nomicfoundation/edr-*` packages from `dependencies` to `optionalDependencies`, so installs only download the build for the current platform instead of all of them.
--- a/.changeset/ten-lines-sort.md
+++ b/.changeset/ten-lines-sort.md
@@ -2,4 +2,4 @@
 "@nomicfoundation/edr": minor
 ---
 
-Changed the platform-specific `@nomicfoundation/edr-*` packages from `dependencies` to `optionalDependencies`, so installs only download the build for the current platform instead of all of them.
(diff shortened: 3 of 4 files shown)


.github/workflows/edr-npm-release.yml
```diff
@@ -542,10 +542,6 @@ jobs:
         uses: ./.github/actions/setup-node
       - name: Install dependencies
         run: pnpm install --frozen-lockfile --prefer-offline
-      - name: Install sponge # needed for prepublish script
-        run: |
-          sudo apt-get update
-          sudo apt-get install -y moreutils
```
anaPerezGhiglia, line 548: it was only needed by the removed rename `optionalDependencies -> dependencies`

scripts/prepublish.sh
```diff
@@ -12,5 +12,3 @@ set -o pipefail
 #   --skip-optional-publish  — don't run `npm publish` for the platform pkgs.
 #   --no-gh-release          — don't try to create a GitHub release.
 pnpm napi pre-publish -t npm --skip-optional-publish --no-gh-release
```
Wodann, line 14: Question: Do we still need `prepublish.sh` if all it now does is `pnpm napi pre-publish -t npm --skip-optional-publish --no-gh-release`?

If not, it seems like a candidate for inlining.

Wodann (approved):
One question to consider before merging this. Otherwise looks good to me. Thanks for making this quality of life improvement!

scripts/prepublish.sh
anaPerezGhiglia: it could be inlined, but right now we have two callers (`edr-nom-release` job and `publish_to_verdaccio` script) that would need to execute the same command to stay on sync. Considering that the script was already defined and it contained some explanatory comments when I implemented it I decided to leave it for now. I have no strong opinions about it, so if you think we would be better off without it we can make the change.

Back in the day this was called as part of `prepublishOnly` lifecycle script, but I'm not sure that's still applicable. As I did not have that much time to dedicate to these changes I decided to go with the most conservative approach and just drop the workaround.

PR #1604, opened 2026-08-06
build: upgrade REVM to v41.0.0 (tag v113)

anaPerezGhiglia:
Supersedes #1590.

## Changes
- Upgrades all `revm` workspace dependencies to the versions released at revm tag `v113`, which unifies every revm crate at `41.0.0`.
- `op-revm` moves from crates.io 19.0.0 to the version released with **op-reth v2.4.1**, consumed as a git dependency on the `ethereum-optimism/optimism` monorepo pinned at that release's commit (details below).
- `foundry-fork-db` bumps to the plain crates.io release 0.27 (pins revm ^41).
- **No public API changes.** revm's "SpecId cleanup" removed several historical hardfork variants, but EDR's own `L1Hardfork`/`OpHardfork` types (introduced in #1601) retain all of them; the removals are absorbed at the revm conversion boundary.

### Dependency changes

| Crate | Old | New |
|---|---|---|
| revm | 38.0.0 | 41.0.0 |
| revm-bytecode | 10.0.0 | 41.0.0 |
| revm-context | 16.0.1 | 41.0.0 |
| revm-context-interface | 17.0.1 | 41.0.0 |
| revm-database-interface | =11.0.1 | =41.0.0 |
| revm-handler | 18.1.0 | 41.0.0 |
| revm-inspector | 19.0.0 | 41.0.0 |
| revm-interpreter | 35.0.1 | 41.0.0 |
| revm-precompile | 34.0.0 | 41.0.0 |
| revm-primitives | 23.0.0 | 41.0.0 |
| revm-state | 11.0.1 | 41.0.0 |
| op-revm | 19.0.0 (crates.io) | op-reth v2.4.1 release, git pin `ethereum-optimism/optimism@a9a8dad` |
| revm-inspectors | 0.39.0 | 0.41.2 (0.42 requires revm 42) |
| foundry-fork-db | 0.26 | 0.27 (crates.io) |
| c-kzg | 2.1.4 | ^2.1.7 (required by revm-precompile 41) |

The locked alloy 2.x family moves 2.0.5 → 2.3.0 — a floor inherited from the monorepo's vendored `op-alloy` (see the `op-revm` section).

<details>
<summary><b>op-revm: git dependency on the optimism monorepo</b></summary>

**Why not crates.io.** No op-revm release supports revm 41: the last crates.io release (20.0.0) still pins `revm = "^38"`.

[...]

PR #1616, opened 2026-08-12
build: use `op-revm` shallow-mirror instead of optimism monorepo

anaPerezGhiglia:
- Switches the `op-revm` source from the git dependency on the `ethereum-optimism/optimism` monorepo (introduced in #1604) to a shallow per-snapshot mirror: [`anaPerezGhiglia/op-revm`](https://github.com/anaPerezGhiglia/op-revm) at tag `op-reth-v2.4.1-a9a8dad`. The mirror is a byte-identical snapshot of `rust/op-revm` at the same monorepo commit (`a9a8dad`, the op-reth v2.4.1 release): only the packaging changes, no EDR code changes.
- **Cache footprint.** The monorepo adds ~1.1 GB to cargo's cache (git db + checkout), per machine and per CI cache — stable cargo cannot shallow-clone git dependencies. The mirror adds ~1 MB.
- **op-alloy.** The monorepo is also where `op-alloy` is developed (it is the source of truth of the published crate), and `op-revm` depends on it by path. A git dependency on `op-revm` therefore resolved the in-repo `op-alloy-consensus`: same name and version (2.0.0) as the crates.io release, but including unreleased commits and a different manifest (alloy floors at `^2.1.1` instead of `^2.0.0`). With this PR:
  - `op-alloy-consensus 2.0.0` appears once (crates.io) instead of twice (crates.io + git). Two same-version copies are distinct types to rustc; no op-alloy type crosses the EDR↔op-revm boundary today, but any future code that does would hit a type mismatch.
  - `reth-codecs`, `reth-codecs-derive`, and `reth-zstd-compressors` 0.5.2 are removed. They were optional, never-compiled dependencies of the git copy. Their 0.3.0 counterparts stay: those belong to the crates.io copy and are also present on `main`.
  - Locked alloy versions stay at 2.3.0 (#1604 raised them from 2.0.5 via the vendored floors). The mirror only requires `^2.0.0`, which 2.3.0 satisfies, and this PR does not downgrade any locked version.
  - Merge-order note: if this PR lands before #1604 reaches `main`, it is worth rebuilding #1604's `Cargo.lock` from `main`'s (`git checkout main -- Cargo.lock`, then a `cargo check`): with the vendored floors gone, nothing forces the alloy family above `main`'s 2.0.5, so the upgrade merges without the alloy bumps.

- The mirror currently lives under a personal account; if this approach is approved, the suggestion is to move the repo to the **NomicFoundation** organization (see Follow-up).

## The mirror

A per-snapshot mirror of `rust/op-revm`, one tag per upstream commit, named `<upstream release>-<upstream sha>` so the manifest documents its own provenance. It includes a `PROVENANCE.md` (update procedure) and a `verify.sh` that re-checks byte-identity against the GitHub API. Tags are never moved, new snapshots get new tags, and reproducibility comes from `Cargo.lock`, which records the exact mirror commit.

## Follow-up

- ~If this approach is approved: migrate the mirror repo to the NomicFoundation organization so ownership and the update procedure are team-managed~ Done


Cargo.toml
```diff
@@ -215,9 +215,10 @@ foundry-compilers = { version = "0.18.0", default-features = false, features = [
 
 ## revm
 # No op-revm release supports revm 41 (op-revm 20.0.0 on crates.io pins revm
-# ^38), so a instead we rely on ethereum-optimism monorepo commit matching
-# the op-reth v2.4.1 release.
-op-revm = { git = "https://github.com/ethereum-optimism/optimism", rev = "a9a8dad3f1500a4cc2e4077edb480848bfdef29a", default-features = false, features = [
+# ^38), so a shallow mirror of the ethereum-optimism monorepo commit matching
+# the op-reth v2.4.1 release (workspace pins revm 41.0.0, the revm version at
+# tag v113) is used instead. See the mirror's PROVENANCE.md.
+op-revm = { git = "https://github.com/anaPerezGhiglia/op-revm", tag = "op-reth-v2.4.1-a9a8dad", default-features = false, features = [
```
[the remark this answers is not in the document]
anaPerezGhiglia: The tag is on a mirror repo we control, following a one-tag-per-snapshot convention (documented in its PROVENANCE.md). Tags should never be moved. Reproducibility comes from Cargo.lock, which records the exact commit.
Keeping the tag as is unless someone from the teams disagrees.

popescuoctavian: Suggestion: add this to the Renovate ignore list. IIUC Renovate won't be able to propose updates for tags, but it may become an issue if we switch to a commit SHA in the future.

popescuoctavian (approved):
LGTM, thanks! Just a minor suggestion for future-proofing.

Cargo.toml
anaPerezGhiglia: Mmh, I'm not sure about this, I think I would skip it. The rule matches by package name regardless of source, so if we ever return to a crates.io `op-revm` it would silently suppress those updates.

popescuoctavian: Fair point. I'm good with that given that we are hosting the shallow copy.

PR #1618, opened 2026-08-12
doc: document best practices for traits as part of style guide

Wodann:
Adds a section to the book's style guide explaining best practices and rules for using `trait`s that are implicit in the existing design of (most of) the multi-chain architecture.


book/src/02_development/06_style_guide.md
```diff
@@ -79,3 +79,196 @@ Rust functions that panic on the Node.js main thread will not return a result an
 Annotating with the `catch_unwind` macro attribute will turn the panic into a `napi::Error`, which can then be returned as a `napi::Result`.
 
 Note that this will not capture panics in background threads.
+
+## Defining traits
+
+When defining traits, try to keep traits composable by only defining the generic types, associated types and/or functions needed for a specific use case.
+
+Furthermore, avoid overconstraining traits by only requiring:
```
anaPerezGhiglia, line 87: This sentence is not easy to read. Is not clear until you read both bulletpoints which way is intended, since it can be interpreted both ways: 

- you are overconstraining when you only require (which should be avoided)
- you should avoid overconstraining by requiring only the listed elements

Perhaps change it to

```suggestion
Furthermore, avoid overconstraining traits. Require only:
```

book/src/02_development/06_style_guide.md
```diff
@@ -79,3 +79,196 @@ Rust functions that panic on the Node.js main thread will not return a result an
...
+E.g. when a member needs the chain spec itself—to call one of its associated functions—the type ends up carrying both the decomposed parameters and the chain spec, which is worse than either alone.
+In cases like that, it is acceptable to keep the chain spec parameter instead.
+
+### Rationale
+
+When a user is calling a function or using a type with trait bounds, they will have to satisfy all the constraints of the trait bounds.
+By using the least constraining option, you can reduce the number of constraints that need to be satisfied at usage sites; thus making the function more reusable and composable.
+
+Moreover, this limits the number of obligations the trait solver has to discharge and—because well-formedness obligations feed region inference—the number of lifetime constraints that come with them.
+Historically, we have run into problems where an unrelated trait bound resulted in an `outlives` requirement and the error blamed a lifetime that was in fact valid.
```
anaPerezGhiglia, line 273: this is almost a copy-paste from "Defining traits" section. I think it can be unified. If you prefer keeping a Rationale subsection for each section, then I would suggest editing them so the reader can spot where the difference is

anaPerezGhiglia (approved):
The CI is failing, and I left some suggestions that are worth considering before merging, but I'm approving it since is just a doc entry, to avoid blocking the merge

anaPerezGhiglia:
by the way, thanks for putting this together! 🙇

book/src/02_development/06_style_guide.md
Wodann: Addressed by https://github.com/NomicFoundation/edr/pull/1618/changes/e1754e96d2d70d3fe9d388c7e6f5f31296f2b530

book/src/02_development/06_style_guide.md
Wodann: Addressed by https://github.com/NomicFoundation/edr/pull/1618/changes/14ad9e599fe6870a32c528a3c2a38833bc6f9eb3

PR #1619, opened 2026-08-13
fix: collect stack traces for `setUp()` failures in `CollectStackTraces::Always`

### LINKED ISSUE #1605: `setUp()` failures in `CollectStackTraces::Always` produce empty stack traces due to skipped re-execution fallback
### Summary

In `CollectStackTraces::Always` mode, a failing `setUp()` produces a stack trace built from step-less trace arenas, so the stack-trace heuristics have nothing to work with. The re-execution fallback that would have produced a correct trace is skipped.

### Details

`ContractRunner::run_tests` deliberately downgrades tracing for `setUp()`, because setup traces are normally only needed to map contract codes to addresses:

```rust
// crates/edr_solidity_tests/src/runner.rs:634-637
let prev_tracer = executor.inspector_mut().tracer.take();
if prev_tracer.is_some() || has_invariants {
    executor.set_tracing(TracingMode::WithoutSteps);
}

let setup_time = Instant::now();
let mut setup: TestSetup<HaltReasonT> = self.setup(&mut executor, call_setup);

executor.inspector_mut().tracer = prev_tracer;   // <-- restored here
```

The original `WithSteps` tracer is restored *before* the setup-failure branch consults it:

```rust
// crates/edr_solidity_tests/src/runner.rs:650-656
setup.stack_trace_result = if executor.tracer_records_steps() {
    // We collected steps during setup, so we can generate the stack trace
    get_stack_trace(
        &*self.contract_decoder,
        setup.traces.iter().map(|(_, arena)| &arena.arena),
        None,
    )
```

`tracer_records_steps()` reports the *configured* mode of the restored tracer (`true` in `Always` mode), not the mode that was actually in effect while `setUp()` ran (`WithoutSteps`). The comment — "We collected steps during setup" — is false on this path.

Consequence: `get_stack_trace` converts an arena whose `CallTraceNode::trace.steps`are all empty. `convert_from_arena` reads `step.pc` per step (`crates/edr_solidity/src/nested_trace/conversion.rs:58-62`), so the resulting `NestedTrace` has no steps and the inference degrades — instead of taking the re-execution branch at `runner.rs:665-681`, which builds a fresh executor with `TracingMode::WithSteps` and re-runs setup precisely to get usable steps.

`CollectStackTraces::OnFailure` is unaffected: `prev_tracer` is `None` there (unless the suite has invariants), so `tracer_records_steps()` is `false` and the re-execution branch is taken correctly.

### Expected

A failing `setUp()` under `CollectStackTraces::Always` yields the same source-level stack trace as under `OnFailure`.

### Suggested fix

Track whether steps were actually recorded *during setup*, rather than asking the restored tracer. Either capture the answer before the downgrade, or don't downgrade setup tracing when `collect_stack_traces == Always`.

### Test coverage gap

`always_mode_produces_stack_trace_for_failing_test` (`crates/edr_solidity_tests/tests/it/repros.rs:686-730`) covers the unit, table, fuzz and invariant paths in `Always` mode, but not a failing `setUp()`. A new fixture with a reverting `setUp()` would catch this.

popescuoctavian:
This PR fixes missing Solidity test stack trace when `setUp()` fails and stack traces are collected with `CollectStackTraces::Always`.

## Problem

In `CollectStackTraces::Always` mode, a failing `setUp()` produced a useless stack trace (`OtherExecutionError` with no source reference) instead of a source-mapped one.

`ContractRunner::run_tests` downgrades tracing to `TracingMode::WithoutSteps` during `setUp()` as an optimization, but restored the original tracer *before* the setup-failure branch consulted `executor.tracer_records_steps()`. In `Always` mode the restored tracer is `WithSteps`, so the check wrongly reported that steps had been collected during setup, and stack traces were inferred from step-less trace arenas instead of taking the re-execution fallback. `OnFailure` mode was unaffected because its tracer never records steps.

## Fix

Capture whether steps were actually recorded during `setUp()` before restoring the original tracer, and use that in the setup-failure branch. A failing `setUp()` under `Always` now takes the same re-execution fallback as `OnFailure` and yields the same source-mapped stack trace.

## Tests

- New fixture contract with a reverting `setUp()` in `StackTraceAlwaysMode.t.sol`, asserted in both `Always` mode (extends `always_mode_produces_stack_trace_for_failing_test`) and `OnFailure` mode (new test) — the latter path was previously untested too.
- Strengthened `assert_execution_error_stack_trace` to require a source-mapped error entry: the degraded trace surfaces as a nominally successful `OtherExecutionError { source_reference: None }`, which the previous assertion would have accepted. Verified red before the fix, green after.

closes #1605

### DIFF
--- a/.changeset/stack-trace-failing-setup-always-mode.md
+++ b/.changeset/stack-trace-failing-setup-always-mode.md
@@ -1,0 +1,5 @@
+---
+"@nomicfoundation/edr": patch
+---
+
+Fixed missing Solidity test stack trace when `setUp()` fails and stack traces are collected with `CollectStackTraces::Always`.
--- a/crates/edr_solidity_tests/src/runner.rs
+++ b/crates/edr_solidity_tests/src/runner.rs
@@ -640,6 +640,10 @@ impl<
         let mut setup: TestSetup<HaltReasonT> = self.setup(&mut executor, call_setup);
         debug!("finished setting up in {:?}", setup_time.elapsed());
 
+        // Whether steps were actually recorded while `setUp()` ran; must be
+        // captured before the original tracer (whose mode may differ) is
+        // restored.
+        let setup_recorded_steps = executor.tracer_records_steps();
         executor.inspector_mut().tracer = prev_tracer;
 
         if setup.reason.is_some() {
@@ -647,7 +651,7 @@ impl<
             // these numbers to reason about the performance of their code.
             let elapsed = start.elapsed();
 
-            setup.stack_trace_result = if executor.tracer_records_steps() {
+            setup.stack_trace_result = if setup_recorded_steps {
                 // We collected steps during setup, so we can generate the stack trace
                 get_stack_trace(
                     &*self.contract_decoder,
--- a/crates/edr_solidity_tests/tests/it/repros.rs
+++ b/crates/edr_solidity_tests/tests/it/repros.rs
@@ -668,16 +668,20 @@ fn assert_execution_error_stack_trace<'suite, HaltReasonT: HaltReasonTrait>(
     };
 
     // A non-empty trace that reaches the failing call's execution error
-    // (exact variant depends on the compilation mode).
+    // (exact variant depends on the compilation mode). The entry must be
+    // source-mapped: a trace inferred from step-less arenas degrades to a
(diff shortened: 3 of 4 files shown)


Wodann (approved):
Thanks for the quick fix! LGTM

PR #1621, opened 2026-08-13
bench: add a memory benchmark for the Solidity test runner

Wodann:
Hardhat 3 users report out-of-memory crashes when running Solidity tests at verbosity `-vvv` and above. Before fixing that, we need a way to measure peak memory use reliably and compare builds. This PR adds a `solidity-tests-memory` benchmark command that does exactly that.

## What it measures

Peak resident memory (RSS) of a full test run, for each combination of repository and Hardhat verbosity level. Verbosity is the input that matters because Hardhat maps it to the EDR options that control trace collection (`includeTraces` / `collectStackTraces`): `-vv` collects nothing, `-vvv` records EVM steps for every test and keeps traces of failing tests, `-vvvv` keeps traces of all tests.

The benchmarked repositories cover the different execution paths: **solady** (1557 tests, fuzz-heavy), **uniswap-v4-core** (598 tests), and **morpho-blue** (145 tests including invariant testing).

## Design

- **One child process per measurement.** `maxRSS` is a per-process high-water mark that never goes down, so a second measurement in the same process would inherit the first one's peak. The driver therefore spawns a fresh process for every (repo, verbosity) cell.
- **A dedicated internal entry script** (`solidity-tests-memory-child.ts`) runs the measurement. The driver passes its parameters as a single JSON argument. The child reports its result as a tagged JSON line on stdout.
- **Two independent memory readings.** The child reads its own peak via `process.resourceUsage()`; the driver additionally wraps the child in `/usr/bin/time -v`, which observes the peak from outside. The external reading is the important one: it also covers a child that the OOM killer terminates before it can report anything.
- **OOM is a result, not a failure.** If a child is killed, the driver records the cell as `OOM` with the last externally observed RSS ("the true requirement is at least this much") and moves on to the next cell.
- **Warm-up compile.** Each repo is compiled in a throwaway child first, so the first measured run doesn't pay one-time solc costs that later runs would get from the artifact cache.
- **Bounded parallelism.** Children are capped at 4 rayon threads. Peak RSS scales with how many test suites run concurrently, and with unbounded parallelism the current runner OOMs a 16 GiB machine at `-vvv` on solady — every cell would just read "OOM", leaving nothing to compare fixes against. A fixed cap keeps every cell measurable and comparable across builds.
- **stdout is data, stderr is progress.** The command prints per-cell progress to stderr and exactly one artifact to stdout: a markdown table (rows = repos, columns = verbosity levels), ready to paste into a PR. Full results are also written to a JSON file.

## Usage

```bash
cd js/benchmark
pnpm soltestsMemory                          # all repos × verbosity 2, 3, 4
pnpm soltestsMemory --repo solady            # one repo
pnpm soltestsMemory --verbosity 3            # one verbosity level
```

## Example output table

| repo | -vv | -vvv | -vvvv |
|---|---|---|---|
| solady | 510 MiB / 3.1s | 4750 MiB / 17.7s | OOM (killed at ≥ 10.3 GiB) |
| uniswap-v4-core | 387 MiB / 1.6s | 2308 MiB / 1.8s | 3652 MiB / 1.9s |
| morpho-blue | 283 MiB / 2.4s | 1681 MiB / 10.9s | 1900 MiB / 10.8s |

## Also in this PR

Repo setup needed two fixes to run the new repos: submodules pinned by the benchmarked commit are now initialized after checkout (the shallow clone doesn't fetch them), and `npm install` falls back to `--ignore-scripts` for repos whose lifecycle scripts need tools we don't have (morpho-blue's `prepare` runs `forge install`, which is redundant here because the clone already checks out submodules).

### DIFF
--- a/js/benchmark/src/index.ts
+++ b/js/benchmark/src/index.ts
@@ -26,6 +26,7 @@ import {
   setupRepo,
   REPOS,
   runSolidityTestsMemoryBenchmark,
+  MEMORY_REPOS,
   MEMORY_VERBOSITIES,
 } from "./solidity-tests.js";
 
@@ -189,10 +190,7 @@ async function main(): Promise<boolean> {
       return false;
     }
   } else if (args.command === "solidity-tests-memory") {
-    const repos =
-      args.repo !== undefined
-        ? args.repo.split(",")
-        : ["solady", "uniswap-v4-core", "morpho-blue"];
+    const repos = args.repo !== undefined ? args.repo.split(",") : MEMORY_REPOS;
     const verbosities =
       args.verbosity !== undefined ? [args.verbosity] : MEMORY_VERBOSITIES;
     const memoryResults = await runSolidityTestsMemoryBenchmark(
--- a/js/benchmark/src/index.ts
+++ b/js/benchmark/src/index.ts
@@ -195,11 +195,20 @@ async function main(): Promise<boolean> {
         : ["solady", "uniswap-v4-core", "morpho-blue"];
     const verbosities =
       args.verbosity !== undefined ? [args.verbosity] : MEMORY_VERBOSITIES;
-    await runSolidityTestsMemoryBenchmark(
+    const memoryResults = await runSolidityTestsMemoryBenchmark(
       repos,
       verbosities,
       benchmarkOutputPath
     );
+    const failed = memoryResults.filter((result) => result.kind === "error");
+    if (failed.length > 0) {
+      console.error(
+        `Error: ${failed.length} measurement(s) failed: ${failed
+          .map((result) => `${result.repo} at verbosity ${result.verbosity}`)
+          .join(", ")}`
+      );
+      return false;
+    }
   } else if (args.command === "compare-forge") {
     if (args.csv_output === undefined) {
       console.error(
--- a/js/benchmark/src/index.ts
+++ b/js/benchmark/src/index.ts
@@ -190,7 +190,13 @@ async function main(): Promise<boolean> {
       return false;
     }
   } else if (args.command === "solidity-tests-memory") {
-    const repos = args.repo !== undefined ? args.repo.split(",") : MEMORY_REPOS;
+    const repos =
+      args.repo !== undefined
+        ? args.repo
+            .split(",")
+            .map((repo) => repo.trim())
+            .filter((repo) => repo.length > 0)
+        : MEMORY_REPOS;
     const verbosities =
       args.verbosity !== undefined ? [args.verbosity] : MEMORY_VERBOSITIES;
     const memoryResults = await runSolidityTestsMemoryBenchmark(
--- a/js/benchmark/src/index.ts
+++ b/js/benchmark/src/index.ts
@@ -137,7 +137,7 @@ async function main(): Promise<boolean> {
   });
   parser.add_argument("--verbosity", {
     type: "int",
-    help: "Hardhat verbosity level for the solidity-tests-memory command. Repeat the command for each level, or omit to measure all of them.",
+    help: `Hardhat verbosity level (0-5, i.e. -v to -vvvvv) for the \`solidity-tests-memory\` command. Pass one level per run; omit to measure verbosities ${MEMORY_VERBOSITIES.join(", ")}, the ones with distinct trace-collection behaviour.`,
   });
   const args: ParsedArguments = parser.parse_args();
 
--- a/js/benchmark/src/index.ts
+++ b/js/benchmark/src/index.ts
@@ -25,6 +25,8 @@ import {
   runForgeTests,
   setupRepo,
   REPOS,
+  runSolidityTestsMemoryBenchmark,
+  MEMORY_VERBOSITIES,
 } from "./solidity-tests.js";
 
 const {
@@ -42,6 +44,7 @@ interface ParsedArguments {
     | "report-provider-benchmark"
     | "solidity-tests-benchmark"
     | "solidity-tests"
+    | "solidity-tests-memory"
     | "compare-forge"
     | "report-forge";
   grep?: string;
@@ -55,6 +58,7 @@ interface ParsedArguments {
   csv_input?: string;
   // eslint-disable-next-line @typescript-eslint/naming-convention
   forge_path?: string;
+  verbosity?: number;
 }
 
 interface BenchmarkScenarioResult {
@@ -94,6 +98,7 @@ async function main(): Promise<boolean> {
       "report-provider-benchmark",
       "solidity-tests-benchmark",
       "solidity-tests",
+      "solidity-tests-memory",
       "compare-forge",
       "report-forge",
     ],
@@ -128,6 +133,10 @@ async function main(): Promise<boolean> {
   parser.add_argument("--forge-path", {
     type: "str",
     help: "Path to forge executable (default: 'forge')",
+  });
+  parser.add_argument("--verbosity", {
+    type: "int",
+    help: "Hardhat verbosity level for the solidity-tests-memory command. Repeat the command for each level, or omit to measure all of them.",
   });
   const args: ParsedArguments = parser.parse_args();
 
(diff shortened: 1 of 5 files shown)


js/benchmark/src/solidity-tests.ts
```diff
@@ -715,3 +732,374 @@ function assertNoFailures(results: SuiteResult[]) {
...
+          repo: repoName,
+          verbosity,
+          peakRssBytes: lastSeenRss,
+          elapsedMs: 0,
+          suiteCount: 0,
+          testCount: 0,
+          failureCount: 0,
+          oom: true,
+        };
+      }
```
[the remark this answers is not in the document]
Wodann: Addressed by https://github.com/NomicFoundation/edr/pull/1621/changes/06bf379bcc31e260782fd7ba19aca035ef17a39b

js/benchmark/src/solidity-tests.ts
```diff
@@ -715,3 +732,374 @@ function assertNoFailures(results: SuiteResult[]) {
...
+        return "—";
+      }
+      // Prefer the externally observed peak; it also covers allocations made
+      // after the in-process read.
+      const peak = result.peakRssBytesExternal ?? result.peakRssBytes;
+      if (result.oom) {
+        return peak > 0 ? `OOM (>= ${displayMiB(peak)})` : "OOM";
+      }
+      return `${displayMiB(peak)} / ${result.elapsedMs}ms`;
+    }),
```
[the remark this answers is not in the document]
Wodann: Addressed by https://github.com/NomicFoundation/edr/pull/1621/changes/06bf379bcc31e260782fd7ba19aca035ef17a39b

We report `ms` for sub-second timings. Otherwise, `s` are reported.

js/benchmark/src/solidity-tests.ts
```diff
@@ -715,3 +732,395 @@ function assertNoFailures(results: SuiteResult[]) {
...
+}
+
+/** Extract `Maximum resident set size (kbytes): N` from `/usr/bin/time -v`. */
+function parseGnuTimeMaxRss(stderr: string): number | undefined {
+  const match = stderr.match(/Maximum resident set size \(kbytes\): (\d+)/);
+  return match === null ? undefined : Number(match[1]) * 1024;
+}
+
+/** Whether a child process exited because it ran out of memory. */
+function isOomError(
```
popescuoctavian, line 878: Question: Should this also handle Javascript V8 heap OOM? From what I understand, that aborts (`JavaScript heap out of memory`, `SIGABRT`, exit code 134).

js/benchmark/src/solidity-tests.ts
```diff
@@ -715,3 +732,395 @@ function assertNoFailures(results: SuiteResult[]) {
...
+/** Compile a repo's contracts + tests without running anything. */
+export async function compileSolidityTestsInput(repoPath: string) {
+  await createSolidityTestsInput(repoPath);
+}
+
+/**
+ * Node arguments invoking the internal child entry point with the given
+ * parameters. `process.execArgv` carries the tsx loader over.
+ */
+function memoryChildNodeArgs(params: MemoryChildParams): string[] {
```
popescuoctavian, line 969: Question: IIUC, this runs with the default V8 heap limit; for other benchmarks we extend that with `--max-old-space-size=28000`. Just making sure this is on purpose.

js/benchmark/src/solidity-tests.ts
```diff
@@ -715,3 +732,395 @@ function assertNoFailures(results: SuiteResult[]) {
...
+    console.error(
+      "note: /usr/bin/time not available, relying on process.resourceUsage() alone"
+    );
+  }
+
+  const results: MemoryRunResult[] = [];
+  for (const repoName of repoNames) {
+    for (const verbosity of verbosities) {
+      console.error(`measuring ${repoName} at verbosity ${verbosity}...`);
+      const result = await runMemoryChildProcess(
```
popescuoctavian, line 934: Suggestion: use `try-catch` to avoid one rejecting cell from discarding the whole run.

js/benchmark/src/solidity-tests.ts
```diff
@@ -715,3 +732,395 @@ function assertNoFailures(results: SuiteResult[]) {
...
+        repoName,
+        repoPaths.get(repoName)!,
+        verbosity,
+        useGnuTime
+      );
+      if (result.oom === true) {
+        console.error(`  OOM: killed at >= ${displayMiB(result.peakRssBytes)}`);
+      } else {
+        console.error(
```
popescuoctavian, line 943: Suggestion: using `peakRssBytes` here and `peakRssBytesExternal` at the end (line 1074) may be confusing when reading the logs. Can we use `peakRssBytesExternal` here as well (if available)?

js/benchmark/src/solidity-tests.ts
```diff
@@ -642,12 +652,19 @@ export async function setupRepo(
     }
   }
 
-  await execAsync("npm install", { cwd: repoPath });
+  try {
+    await execAsync("npm install", { cwd: repoPath });
+  } catch {
+    console.error(
```
popescuoctavian, line 658: Suggestion: log the error here to help debug potential situations where running with `--ignore-scripts` leaves the repository in a bad state.

js/benchmark/src/index.ts
popescuoctavian, line 116: Issue: this argument means different things for `solidity-tests` (path to repo) vs `solidity-tests-memory` (list of repo names). We should update the help text to reflect that.

js/benchmark/src/index.ts
```diff
@@ -179,6 +188,18 @@ async function main(): Promise<boolean> {
       console.error("Error: --repo is required for solidity-tests command");
       return false;
     }
+  } else if (args.command === "solidity-tests-memory") {
+    const repos =
+      args.repo !== undefined
+        ? args.repo.split(",")
```
popescuoctavian, line 194: Suggestion: also apply `trim` to accept cases like `--repo solady, morpho-blue`.

js/benchmark/src/index.ts
```diff
@@ -179,6 +188,18 @@ async function main(): Promise<boolean> {
...
+        : ["solady", "uniswap-v4-core", "morpho-blue"];
+    const verbosities =
+      args.verbosity !== undefined ? [args.verbosity] : MEMORY_VERBOSITIES;
```
popescuoctavian, line 197: Nit: validate verbosity argument.

js/benchmark/src/index.ts
```diff
@@ -179,6 +188,18 @@ async function main(): Promise<boolean> {
       console.error("Error: --repo is required for solidity-tests command");
       return false;
     }
+  } else if (args.command === "solidity-tests-memory") {
+    const repos =
+      args.repo !== undefined
+        ? args.repo.split(",")
+        : ["solady", "uniswap-v4-core", "morpho-blue"];
```
popescuoctavian, line 195: Nit: move the repo list to an export next to `MEMORY_VERBOSITIES` to make it easier to find/extend.

popescuoctavian (changes requested):
Thanks for adding this! LGTM overall, apart from a minor issue and some suggestions/questions.

js/benchmark/src/solidity-tests.ts
Wodann: Good call; yes. V8 aborts with FATAL ERROR: … JavaScript heap out of memory (SIGABRT, exit 134 / GNU time reports signal 6). However, in the ended I decided to match the message rather than SIGABRT, which has other causes (e.g. a Rust abort).

Addressed by 65b300492. Verified against a real V8 OOM under /usr/bin/time.

js/benchmark/src/solidity-tests.ts
Wodann: Not on purpose. 

Addressed by 18caa3ef7. The soltestsMemory script now passes the same node flags as the other benchmark scripts (incl. --max-old-space-size=28000); children inherit them via process.execArgv, so the OS rather than V8's default heap cap decides when a child runs out of memory, which is what we want to measure.

js/benchmark/src/solidity-tests.ts
Wodann: Good idea. Addressed by 61e611c1

js/benchmark/src/solidity-tests.ts
Wodann: Addressed by 61e611c1

js/benchmark/src/solidity-tests.ts
Wodann: Addressed by fef215d9c

js/benchmark/src/index.ts
Wodann: Addressed by 508017a6f

js/benchmark/src/index.ts
Wodann: Addressed by 963bb7ff1. Lists the default repos dynamically from `MEMORY_REPOS`.

js/benchmark/src/index.ts
Wodann: Incorporated in f636e0de2. A good usage of [Postel's law](https://en.wikipedia.org/wiki/Robustness_principle).

js/benchmark/src/index.ts
Wodann: Addressed by 701637a3c

popescuoctavian (approved):
Thanks for integrating the feedback!

PR #1622, opened 2026-08-14
refactor: rename the solx compiler type to slangSolx

kanej:
We rename `CompilerType::Solx` to `SlangSolx` as an accepted compiler type. This is to mirror the change in the renamed `hardhat-slang-solx` plugin.

EDR reads a build info's `compilerType` field to decide how to recover source locations: `solc` means `evm.{deployed,}Bytecode.sourceMap`, and the solx variant means `evm.{deployed,}Bytecode.debugInfo`.

The Hardhat plugin that produces those build infos has been renamed to `hardhat-slang-solx`, and the compiler type it registers and writes is now `slangSolx`. That value triggered `to_compiler_type`'s unknown-value path and fell back to `Solc`, where we look for a `sourceMap` that solx does not emit. Solidity stack traces were silently lost: every failing test under a solx build profile printed `Stack Trace Warning: Instruction not found at PC 2` and no frames.

Currently `hardhat-slang-solx` does a remapping of `slangSolx` to `solx` to support EDR. When EDR changes to the new compiler type value, we will update the `hardhat-slang-solx` plugin. No backwards compatibility is required.

### DIFF
--- a/.changeset/great-hornets-invite.md
+++ b/.changeset/great-hornets-invite.md
@@ -2,4 +2,4 @@
 "@nomicfoundation/edr": patch
 ---
 
-Update supported compiler types to include slangSolx.
+Changed `solx` compiler type string to `slangSolx`.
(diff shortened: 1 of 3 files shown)


Wodann (approved):
Thanks for the rename!

PR #1623, opened 2026-08-15
refactor: name the failing arena explicitly in `get_stack_trace`

Wodann:
`get_stack_trace` used to take one flat iterator of trace arenas and convert whichever came last, relying on a doc comment — "assumes last trace is the error one" — to carry the whole contract. Every arena before the last was only ever walked to collect the creation/runtime bytecode of deployed contracts, needed to decode the failing trace.

This PR makes that split explicit in the signature:

- `failing_trace` — the arena that is converted and analysed;
- `code_sources` — arenas walked only for their CREATE nodes, to map contract addresses to code;
- `deployed_code` — pre-computed address→code mappings, seeded before the arenas are walked. This replaces the previous third parameter, which could only seed runtime code; the provider now passes `DeployedCode { runtime: Some(..) }` with unchanged behaviour.

The test runner's `collect_stack_trace` helper requires the failing arena the same way. Its callers split it off `execution_traces` themselves, so the reasoning about whether a failing arena was recorded now lives at each call site, where it can actually be justified: the unit, table and fuzz sites run right after the failing call's arena was recorded under an active tracer; the one site where no arena may exist — an invariant campaign that failed before recording any EVM call — handles that case explicitly.

## Why

Two reasons:

1. **Correctness by construction.** The "last arena is the failing one" assumption was easy to break silently: any caller appending an arena after the failing one would change which trace gets analysed, with no compiler help. Now the failing arena is named at every call site.
2. **It unlocks the memory work.** This establishes the invariant that only the failing arena's recorded EVM steps are ever read — every other arena contributes just its CREATE nodes. The follow-up PRs (fixing the OOM at Hardhat verbosity `-vvv`+) rely on exactly that to strip steps from every retained arena.

## Behaviour change (one, deliberate)

An invariant campaign that fails without recording any EVM call (e.g. an ABI error, or too many `vm.assume` rejects) now reports no stack trace. Previously the "last arena" fallback silently pointed the stack-trace heuristics at the last setup arena — a call that necessarily succeeded, since a failed `setUp()` never reaches invariant execution — producing a heuristic failure at best. No arena describes such a failure; the test result's `reason` string already explains it ("failed to set up invariant testing environment: …").

Everything else is a pure refactor: same arenas walked in the same order, same seeding precedence (pre-computed code → code sources → failing trace, later wins).

## Testing

- `cargo test -p edr_solidity_tests --test it` (123 passing) — including `always_mode_produces_stack_trace_for_failing_test`, which exercises stack-trace generation through the unit, table, fuzz, invariant and invariant-campaign-failure paths.
- Benchmarked runtime-neutral against the memory benchmark from #1621 (that harness was built to measure this PR's follow-ups; this refactor sits within run-to-run noise on every cell).

### DIFF
--- a/crates/edr_provider/src/error.rs
+++ b/crates/edr_provider/src/error.rs
@@ -25,7 +25,7 @@ use edr_runtime::{overrides::AccountOverrideConversionError, transaction};
 use edr_signer::SignatureError;
 use edr_solidity::{
     contract_decoder::{ContractDecoder, ContractDecoderError},
-    solidity_stack_trace::{get_stack_trace, StackTraceCreationResult},
+    solidity_stack_trace::{get_stack_trace, DeployedCode, StackTraceCreationResult},
 };
 use edr_state_api::StateError;
 use foundry_evm_traces::CallTraceArena;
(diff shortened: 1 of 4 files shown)


anaPerezGhiglia (approved):
LGTM

PR #1628, opened 2026-08-19
misc: correct semver level of change

Wodann:
The changeset that was submitted for the original change erroneously had the patch-level. Since it's a breaking change, it should have been a minor change. This PR fixes it in the changelog.

I already fixed the corresponding GitHub release.


popescuoctavian (approved):
Good catch @Wodann!

PR #1629, opened 2026-08-19
build: add `os`/`cpu`/`libc` metadata to platform packages

Wodann:
#1615 redefined the platform-specific `@nomicfoundation/edr-*` packages as `optionalDependencies` so installs would only download the build for the current platform. They carry no `os`, `cpu` or `libc` fields, though, so nothing filters them — all seven still install:

```
$ npm install --dry-run @nomicfoundation/edr@0.16.0   # linux-x64
add @nomicfoundation/edr-win32-x64-msvc 0.16.0
add @nomicfoundation/edr-darwin-arm64 0.16.0
… all 7 …
added 8 packages
```

This restores the metadata that caa6f8502 removed when it made the packages hard dependencies. Note it also re-exposes npm/cli#4828 for npm < 11.3.0 — the bug caa6f8502 was avoiding — which #1615's changeset already documents.


popescuoctavian (approved):
Thanks for fixing this!

PR #1630, opened 2026-08-19
fix: prune absent platform packages from the Verdaccio publish

Wodann:
The HH3 regression benchmark has failed on all six baselines since #1615 (runs 455, 457, 460, 462, 463, 468), every one at the `Run regression benchmark` step. The two baselines before it passed.

#1615 pointed `publish_to_verdaccio.sh` at `prepublish.sh`, which lists every platform package as an `optionalDependency`. A local build only produces — and this script only publishes — the host's, so the rest resolve to nothing in the local registry.

npm, pnpm and bun skip `optionalDependencies` they can't resolve. Yarn does not: Classic fails with `Couldn't find any versions`, Berry with `YN0082: No candidates found`. That takes out the four Yarn scenarios — `1inch-aqua`, `1inch-cross-chain-swap`, `aave-v4` and `lidofinance-core`.

This prunes the platform packages a run won't publish, keeping the `optionalDependencies` shape #1615 introduced. Verified locally: both Yarn flavours go from a hard resolution failure to a clean install with the binding linked.

### Validation

- [x] Validating the fix using a [regression benchmark run](https://github.com/NomicFoundation/edr/actions/runs/32240052630).

### DIFF
--- a/scripts/publish_to_verdaccio.sh
+++ b/scripts/publish_to_verdaccio.sh
@@ -4,12 +4,15 @@
 #
 # It stages the prebuilt native binary into its platform package, compiles the
 # bundled TypeScript helpers, pins the packages to the requested version, wires
-# the platform packages as optionalDependencies — the same package shape as a
-# real release — and publishes the main package and this platform's package
-# (platform package first).
+# this host's platform package in as an optionalDependency — mirroring a real
+# release's package shape — and publishes the main package and this platform's
+# package (platform package first).
 #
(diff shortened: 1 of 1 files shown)


Wodann:
/bench scenarios=1inch-aqua,1inch-cross-chain-swap,aave-v4,lidofinance-core

popescuoctavian (approved):
LGTM, given that the regression benchmark succeeds.

PR #1632, opened 2026-08-19
chore(renovate): change schedule to weekends, rebase only on conflicts, add edrextended as reviewer

popescuoctavian:
**Note:** After further investigation into renovate's branch worker code, I've decided to drop the `updateNoteScheduled: false` setting at this point, since it also overwrites the `rebaseWhen: conflicted` setting. This could lead to a situation where, if you have multiple renovate PRs, merging the first one causes conflicts for the others, and a rebase isn't triggered automatically if you are outside the schedule. We'll reassess after observing the outcome of this PR.


Wodann (approved):
LGTM. Thanks!

PR #1637, opened 2026-08-20
chore: rename hardfork variants to follow Rust naming conventions

anaPerezGhiglia:
##  Scope
**Internal refactor - not NAPI user-visible**: The public string representation  is unchanged, so this has no effect on the npm package

Renaming the hardfork name strings to match Hardhat's definitions is intentionally out of scope and will be done in a separate branch.

## Changes
- Renamed `L1Hardfork` and `OpHardfork` enum variants from SCREAMING_SNAKE_CASE to PascalCase. The types were introduced in #1601 modeled on revm's SpecId, which uses SCREAMING_SNAKE_CASE variants; this brings them in line with standard Rust style and cleans up artifacts of mirroring revm.
- Dropped the `#[strum(serialize = …)]` attributes that repeated the variant name. strum's default emits the variant identifier verbatim, so these were redundant. 
- Removed the now-unnecessary `#[allow(non_camel_case_types)]`
- Removed the `serde::Serialize/Deserialize` derives from both enum. 

## 👀 Note for the reviewer
- the `serde::Serialize/Deserialize` derives gave the types a second string representation, serde serialized variant identifiers while strum `Display/FromStr` uses the public hardfork name strings, and the two disagree for the multi-word L1 hardforks (e.g. "FrontierThawing" vs "Frontier Thawing"). Since nothing exercises the serde impls, and Hardhat passes hardforks across N-API as strings parsed with strum `FromStr`, they were removed rather than kept in an inconsistent, unused state.
- commits autocontained, so this PR can be reviewed commit by commit


Wodann (approved):
LGTM. Thanks!

popescuoctavian:
Merging this to be included in today's release, per @anaPerezGhiglia's request.

PR #1639, opened 2026-08-21
fix(tools): avoid uppercasing OpHardfork variants in op-chain-config generator

popescuoctavian:
The hardfork variants were renamed to PascalCase in #1637, but the generator still emitted `OpHardfork::BEDROCK`-style names, breaking the ["Check generated files" CI workflow](https://github.com/NomicFoundation/edr/actions/runs/32454139339).


Wodann (approved):

Wodann:
Thanks for the quick fix. Looks good apart from styling

PR #1641, opened 2026-08-21
fix(tools): repair no-op `--check` in op-chain-config generator and update generated configs

popescuoctavian:
The --check of op-chain-config generator which runs in CI was always succeeding. I've discovered this while testing #1639.

### Problem

The cargo op-chain-config -- --check step in the "Check generated files" workflow has never detected out-of-date configs. It used git diff -G'<regex>' --exit-code, which was broken three ways:

- The single quotes were part of the argument (Command::arg passes it verbatim — no shell strips them), so the pattern could never match.
- The slashes were doubled (https:////github.com//...) as if they needed escaping.
- -G selects files whose changed lines match the regex; the intent ("ignore the // source: <sha> line") requires --ignore-matching-lines (-I), which discounts matching lines instead.

### Changes

Generator (crates/tool/op_chain_config_generator):
- Use git diff --ignore-matching-lines='^// source: ...' with a corrected regex, scoped to the generated paths, so SHA-only changes pass and any real drift fails.
- Wipe the generated directory before regenerating, so chains removed from the registry no longer leave stale modules behind.
- Fix the --check rollback: it missed generated.rs (the generated/* pathspec doesn't cover it) and couldn't remove newly generated untracked files (now handled with a scoped git clean).

Workflow (.github/workflows/check-generated-files.yml):
- Removed the redundant cargo op-chain-config generation step before the --check step — a leftover of the workflow's original implementation; --check regenerates on its own.

Regenerated configs (first regeneration since December):
- Added: Celo Sepolia.
- Removed (removed upstream from the registry): Arena-Z, Swell, Creator Chain testnet, Ethernity Sepolia, Base (mainnet + Sepolia).

Dropped built-in Base support (crates/edr_op):
- Beyond the regenerated configs, removed the hand-written Base hardfork activations (hardfork/base.rs) and the Base-specific tests. Base no longer follows the OP stack protocol, so EDR's OP chain configs cannot represent it correctly. If EDR supports Base again, it should model Base's own protocol.
- Added a changeset for the removed built-in chain configurations.


anaPerezGhiglia (approved):
LGTM! thanks for pushing for fixing this! 🙇

PR #1642, opened 2026-08-21
test(bench): add the ability to temporarily pin Hardhat to an open PR

popescuoctavian:
This PR adds a Hardhat compatibility pin to the HH3 regression benchmark

### Why

When an EDR change breaks compatibility with current Hardhat `main`, the regression benchmark starts failing until the matching Hardhat PR merges. This PR lets an EDR branch pin the benchmark to a specific commit on an open Hardhat PR.

### How it works

An EDR PR can check in `.github/hardhat-compat-pin.json`:

```json
{ "pr": <Hardhat PR number>, "sha": "<full 40-hex commit sha>", "reason": "..." }
```

Any run that would otherwise default to Hardhat `main` — push baselines, and `workflow_dispatch` / `/bench` runs without an explicit `hardhat-ref` — resolves the pin (read at the EDR ref being benchmarked) and:

- **Hardhat PR open** → benchmarks against the pinned sha; `/bench` status comments note the active pin.
- **Hardhat PR merged** → reverts to `main` automatically and emits a notice that the pin file can be removed.
- **Hardhat PR closed without merging** → reverts to `main` with a warning.
- **Malformed pin** (bad JSON, missing/invalid `pr` or `sha`) → fails loudly instead of silently benchmarking `main` — that's exactly the incompatible state the pin exists to avoid. On `/bench`, the error is reported back to the PR and the run is skipped.

An explicit `hardhat-ref` (dispatch input or `hardhat-ref=` comment arg) always wins over the pin. The pinned PR must live in `NomicFoundation/hardhat` (not a fork) so the runner can check the sha out directly. The `hardhat-ref` dispatch input's `default: "main"` is dropped (empty now means "main, or the pin if active").

### Testing

- Unit tests in `resolve-regression-trigger.test.cjs` covering all pin states (open / merged / closed / malformed) across all three triggers, plus explicit-ref precedence.
- Verified live via `workflow_dispatch` on this branch with a pin targeting an open Hardhat PR ([run](https://github.com/NomicFoundation/edr/actions/runs/32716610698)): the resolver logged `Hardhat compat pin active: NomicFoundation/hardhat#8548 @ d57964b…` and the benchmark checked out Hardhat at the pinned sha. Also verified the merged-PR case reverts to `main`. The test pin file is not part of this PR.

### DIFF
--- a/.github/scripts/resolve-regression-trigger.cjs
+++ b/.github/scripts/resolve-regression-trigger.cjs
@@ -244,16 +244,14 @@ module.exports = async ({ github, context, core }) => {
           }
         }
 
-        // Gate on EDR CI being green for the PR head before spending
-        // ~3h on the self-hosted runner.
-        const green =
-          pinError === undefined && (await waitForEdrCi(pr.head.sha));
         if (pinError !== undefined) {
           await postComment(
             `⚠️ Could not resolve the Hardhat compat pin, so the regression ` +
               `benchmark was not started: ${pinError}`
           );
-        } else if (green) {
+          // Gate on EDR CI being green for the PR head before spending
+          // ~3h on the self-hosted runner.
+        } else if (await waitForEdrCi(pr.head.sha)) {
           shouldRun = true;
           // Only mention a filter that actually narrows the run (`*` = all).
           const filterNotes = [
--- a/.github/scripts/resolve-regression-trigger.cjs
+++ b/.github/scripts/resolve-regression-trigger.cjs
@@ -28,25 +28,17 @@ const CI_POLL_INTERVAL_MS = 30 * 1000; // 30 seconds
 const DEFAULT_SCENARIO_FILTER = "*";
 const DEFAULT_BENCHMARK_FILTER = "test solidity,test mocha,test vitest";
 
-// The Hardhat repository the benchmark runs against.
-const HARDHAT_OWNER = "NomicFoundation";
-const HARDHAT_REPO = "hardhat";
-
-// Optional Hardhat compatibility pin, checked into the EDR repo while a
-// breaking EDR change needs a not-yet-merged Hardhat counterpart. Format:
-//
-//   {
-//     "pr": <Hardhat PR number>,
-//     "sha": "<full 40-hex commit sha on that PR>",
-//     "reason": "<optional free-form note>"
-//   }
-//
-// While the Hardhat PR is open, runs that don't name an explicit hardhat-ref
-// benchmark against the pinned sha; once it's merged (or closed) they revert
-// to `main` automatically. Delete the file after the Hardhat PR merges.
-// The pinned PR must live in NomicFoundation/hardhat (not a fork), so the
-// benchmark job can check the sha out directly.
-const HARDHAT_PIN_PATH = ".github/hardhat-compat-pin.json";
+// Optional Hardhat compatibility pin (see hardhat-compat-pin.cjs for the file
+// format). While the pinned Hardhat PR is open, runs that don't name an
+// explicit hardhat-ref benchmark against the pinned sha; once it's merged (or
+// closed) they revert to `main` automatically. Delete the file after the
+// Hardhat PR merges.
+const {
+  HARDHAT_OWNER,
+  HARDHAT_REPO,
+  HARDHAT_PIN_PATH,
+  parseHardhatPin,
+} = require("./hardhat-compat-pin.cjs");
 
 module.exports = async ({ github, context, core }) => {
   const { owner, repo } = context.repo;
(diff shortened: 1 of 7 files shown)


popescuoctavian:
Running `/bench` here picks up the workflow from `main`, so I've tested this by running the following command on a separate branch containing a Hardhat pin (open & merged PR):

```
gh workflow run hh3-regression-benchmark.yml \
  --repo NomicFoundation/edr \
  --ref popescuoctavian/hh3-regression-benchmark-pin-branch
```

Runs: 
- https://github.com/NomicFoundation/edr/actions/runs/32719980549/job/97409132258

image

- https://github.com/NomicFoundation/edr/actions/runs/32720373878/job/97410317694
image

.github/scripts/resolve-regression-trigger.cjs
```diff
@@ -66,6 +92,69 @@ module.exports = async ({ github, context, core }) => {
     return false;
   }
 
+  // Resolve the Hardhat ref for runs that didn't name one explicitly: `main`,
+  // unless a compat pin (HARDHAT_PIN_PATH) exists on `ref` and its Hardhat PR
+  // is still open. Returns { ref, pin? }, with `pin` set only when the pinned
+  // sha is actually used. Throws on a malformed pin (or an unreadable pinned
+  // PR) so misconfiguration fails loudly instead of silently benchmarking
+  // `main` — which is exactly the incompatible state the pin exists to avoid.
```
Wodann, line 92: suggestion: can we also add a CI check (that can be gated on a change to `.github/hardhat-compat-pin.json`) that validates the formatting of the JSON file's contents?

This would allow us to catch erroneous formatting before merging (e.g. if `/bench` was not run in the PR that added the compatibility pin).

.github/scripts/resolve-regression-trigger.cjs
```diff
@@ -152,14 +244,40 @@ module.exports = async ({ github, context, core }) => {
...
-        const green = await waitForEdrCi(pr.head.sha);
-        if (green) {
+        const green =
+          pinError === undefined && (await waitForEdrCi(pr.head.sha));
+        if (pinError !== undefined) {
+          await postComment(
+            `⚠️ Could not resolve the Hardhat compat pin, so the regression ` +
+              `benchmark was not started: ${pinError}`
+          );
+        } else if (green) {
```
Wodann: nit: Doing both:

```
        const green =
          pinError === undefined && (await waitForEdrCi(pr.head.sha));
```

and 

```
if (pinError !== undefined) {
  // ..
} else if (green) {
```

feels double. You could do either/or.

1. First check `if (pinError != undefined)` already rules out that `else if (green)` occurs, so you can keep `const green = await waitForEdrCi(pr.head.sha);`
2. Move `if (pinError != undefined)` into the `else` branch of `if (green)` while keeping `const green = pinError === undefined && (await waitForEdrCi(pr.head.sha));`

.github/scripts/resolve-regression-trigger.test.cjs
```diff
@@ -39,11 +53,28 @@ function makeDeps({ eventName, sha, payload = {}, ci, pr } = {}) {
         }),
       },
       pulls: {
-        get: async () => {
+        get: async ({ repo: pullRepo }) => {
+          if (pullRepo === "hardhat") {
+            if (hardhatPr === undefined)
```
Wodann: nit: add curly braces to clarify the if-statement's body.

.github/scripts/resolve-regression-trigger.test.cjs
```diff
@@ -39,11 +53,28 @@ function makeDeps({ eventName, sha, payload = {}, ci, pr } = {}) {
...
+              throw new Error("hardhat pulls.get not expected");
```
Wodann: question: The error message is not clear to me. What does it mean?

Wodann (approved):
Thanks for the quick turnaround on this.

I left some comments, but feel free to merge without re-review if you deem the suggested improvements worth addressing.

.github/scripts/resolve-regression-trigger.cjs
popescuoctavian: Good idea! Addressed by: 22b96711e54745d6aad45898ba8b83aea42ff889

.github/scripts/resolve-regression-trigger.cjs
popescuoctavian: Addressed by: 560a4a481931245c5d3d2cc28ca56c0ebadf7ea0

.github/scripts/resolve-regression-trigger.test.cjs
popescuoctavian: Addressed by 560a4a481931245c5d3d2cc28ca56c0ebadf7ea0

.github/scripts/resolve-regression-trigger.test.cjs
popescuoctavian: It means that this case is not expected. Improved error message in 560a4a481931245c5d3d2cc28ca56c0ebadf7ea0

.github/scripts/resolve-regression-trigger.cjs
popescuoctavian: Tested here: https://github.com/NomicFoundation/edr/actions/runs/32736181211/job/97459433974?pr=1642

PR #1664, opened 2026-08-24
fix(scripts): narrow napi targets instead of staging placeholder binaries

popescuoctavian:
Fixes the `Regression benchmark` job on main, broken since #1650 bumped
`@napi-rs/cli` to 3.8.6 (first surfaced on #1648's run, the first to actually
execute the job after the bump):

    Internal Error: Release package @nomicfoundation/edr-darwin-arm64 is incomplete: missing edr.darwin-arm64.node

`@napi-rs/cli` >= 3.8 validates during `pre-publish` that every target in
`napi.targets` has its `.node` binary staged — unconditionally, regardless of
`--skip-optional-publish`. #1650 fixed this for the release workflow by moving
`pnpm artifacts` before `prepublish.sh`, but the Verdaccio flow only ever
builds one binary, so validation fails on the first missing platform.

Fix: stage empty placeholder `.node` files for the platforms that weren't
built, run `prepublish.sh` unmodified — so `pre-publish` sees the exact same
configuration as a real release — then delete the placeholders. They can never
be published: `prepublish.sh` passes `--skip-optional-publish`, the script only
publishes the host's platform package, and they're removed right after
`pre-publish` returns. The existing `optionalDependencies` prune loop is kept
unchanged. This avoids hardcoding the platform→triple mapping and keeps the
Verdaccio route as close as possible to the real publish flow.

Verified locally against `@napi-rs/cli` 3.8.6: without placeholders,
`pre-publish` reproduces the exact CI error; with them it succeeds, and the
resulting package state (versions, `optionalDependencies`) is identical to
what the script produced before the CLI bump.

### DIFF
--- a/.github/workflows/edr-ci.yml
+++ b/.github/workflows/edr-ci.yml
@@ -295,6 +295,8 @@ jobs:
 
       - name: Install package
         run: pnpm install --frozen-lockfile --prefer-offline
+      - name: Test repo scripts
+        run: pnpm run test:scripts
       - name: Run build script
         run: pnpm run build:dev
       - name: Run lint script
--- a/package.json
+++ b/package.json
@@ -4,11 +4,13 @@
   "author": "Nomic Foundation",
   "devDependencies": {
     "@changesets/cli": "^2.29.5",
+    "@types/node": "^22.0.0",
     "prettier": "^3.2.5",
-    "syncpack": "^15.0.0"
+    "syncpack": "^15.0.0",
+    "typescript": "~5.8.2"
   },
   "engines": {
-    "node": ">=22.13.0",
+    "node": ">=22.18.0",
     "pnpm": ">=11"
   },
   "license": "SEE LICENSE IN EACH PACKAGE'S LICENSE FILE",
@@ -17,10 +19,12 @@
   "scripts": {
     "build": "pnpm run --recursive build",
     "build:dev": "pnpm run --recursive build:dev",
-    "lint": "syncpack lint && pnpm run prettier && pnpm run --recursive lint",
+    "lint": "syncpack lint && pnpm run prettier && pnpm run tsc:scripts && pnpm run --recursive lint",
     "lint:fix": "syncpack format && pnpm run prettier:fix && pnpm run --recursive lint:fix",
-    "prettier": "prettier --check \".github/**/*.{yml,ts,js}\" \"**/*.md\"",
+    "prettier": "prettier --check \".github/**/*.{yml,ts,js}\" \"scripts/**/*.{ts,json}\" \"**/*.md\"",
     "prettier:fix": "pnpm run prettier --write",
-    "test:workflows": "node --test \".github/scripts/*.test.cjs\""
+    "test:scripts": "node --test \"scripts/**/*.test.ts\"",
+    "test:workflows": "node --test \".github/scripts/*.test.cjs\"",
+    "tsc:scripts": "tsc --noEmit -p scripts/tsconfig.json"
   }
 }
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -18,12 +18,18 @@ importers:
       '@changesets/cli':
         specifier: ^2.29.5
         version: 2.31.1(@types/node@22.20.1)
+      '@types/node':
+        specifier: ^22.0.0
+        version: 22.20.1
       prettier:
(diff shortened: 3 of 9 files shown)


Wodann (commented):
Is there a fix for this problem that:
1. doesn't duplicate the supported platform "triples" into this script; and
2. keeps the verdaccio publish route as close as possible to the real publish script?

popescuoctavian:
@Wodann I've pushed a different approach, which adds placeholders for missing binaries. This keeps the implementation closer to real publishing.

Wodann:
### Alternative: narrow `napi.targets` for this one invocation

Pushed as https://github.com/NomicFoundation/edr/pull/1664/changes/b9a65f476540a408d6ee625e2b3b1d9fc92eed27.

There's no flag to skip the pre-publish validation — `prePublish()` loops over every target from `readNapiConfig` and calls `validateReleasePackageContents` unconditionally (`src/api/pre-publish.ts:342`), and none of the documented options (`--skip-optional-publish`, `--dry-run`, `--npm-dir`, …) gate it. But `--config-path` overrides `napi.targets` for that invocation, so the other platforms are never considered in the first place.

**Change**

- `scripts/write_napi_host_config.ts` writes a config naming only this host's target, looked up in `napi.targets` via napi's exported `parseTriple` — so adding a platform still needs no change here. It follows the Hardhat repo's precedent of running TypeScript scripts directly with `node`.
- `scripts/publish_to_verdaccio.sh` calls it into a `mktemp` file (removed by an `EXIT` trap), checks the file is non-empty, and passes `--config-path`.
- `scripts/prepublish.sh` forwards `"$@"` to `napi pre-publish`. The release workflow calls it with no args, so that path is unchanged.
- `scripts/tsconfig.json` + `tsc:scripts` / `test:scripts` give `scripts/` the type-checking, formatting and test coverage Hardhat's scripts have. `scripts/package.json` marks the directory ESM — the repo root can't be `type: module` because `config/eslint/eslintrc.js` and `.syncpackrc.js` are CommonJS.
- `engines.node` moves to `>=22.18.0`, the first release where Node runs TypeScript without a flag. Every CI job is already on 24.

**Why this over placeholders**

- No fake `.node` files on disk. Validation currently only does `existsSync`, but it already reads binary contents for wasm targets, so empty placeholders are betting on that not spreading to node targets.
- The `npm pkg delete optionalDependencies.…` pruning loop becomes unnecessary — `resolveRootOptionalDependencies` wires exactly the configured targets, so only this host's is ever added.
- Smaller blast radius on the working tree: **2** tracked files modified (root + host platform `package.json`) instead of **8** — with the full target list, napi version-syncs all 7 `npm/*/package.json` even though 6 are never published.

**Tests** (24, ~0.8s, no Rust build and no Verdaccio needed)

- The suffix→target lookup, and the emitted config's exact bytes.
- The CLI's exit codes and its refusal to write a config when no target matches.
- The emitted config fed to napi's own `readNapiConfig`, asserting `binaryName` survives — see caveat 1.
- The invariant that `napi.targets` and `crates/edr_napi/npm/` stay in sync, generated from the real directory listing, so adding a platform to one and not the other fails with a named test.

**Two things worth a look in review:**

1. napi prints

   > Both napi field in `crates/edr_napi/package.json` and [NAPI-RS config](…) file are found, the NAPI-RS config file will be used.

   The message overstates it — the merge is a shallow `Object.assign` over `pkgJson.napi` (`src/utils/config.ts:283`), so `binaryName` and everything else survive and only `targets` is overridden. That's load-bearing, so there's a test pinning it: if an upgrade made it a true replacement, `binaryName` would silently fall back to `"index"` and pre-publish would look for the wrong `.node`.

2. `scripts/package.json` is a 5-line ESM marker, not a workspace package (the pnpm globs don't cover `scripts/`, and it's `private`). Without it, `node scripts/*.ts` warns `MODULE_TYPELESS_PACKAGE_JSON` on every run.

**Unrelated sharp edge found while testing:** after running `publish_to_verdaccio.sh` locally, `crates/edr_napi/package.json` is left at the local version with an injected `optionalDependencies`, and `pnpm run lint` is **green** on that state — no prettier glob or syncpack rule covers `crates/**/package.json`. A `git commit -a` would ship a version downgrade of the published package. The script's header does say to run `git checkout -- crates/edr_napi` afterwards; might be worth making that harder to miss.

Wodann (approved):

PR #1666, opened 2026-08-25
rename compiler type to slang-solx

kanej:
The Hardhat `hardhat-slang-solx` plugin is changing the compiler type value to `slang-solx` (from `slangSolx`).

This is now the value that is expected on the wire from Hardhat, and off of which conditional per compiler execution will be based.

`slangSolx` will no longer be recognized.

### DIFF
--- a/crates/edr_solidity/src/artifacts.rs
+++ b/crates/edr_solidity/src/artifacts.rs
@@ -151,6 +151,7 @@ pub enum CompilerType {
     #[default]
     Solc,
     /// Slang solx compiler
+    #[strum(serialize = "slang-solx")]
     SlangSolx,
 }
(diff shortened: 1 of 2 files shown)


crates/edr_solidity/src/artifacts.rs
```diff
@@ -151,6 +151,7 @@ pub enum CompilerType {
     #[default]
     Solc,
     /// Slang solx compiler
+    #[strum(serialize = "slang-solx")]
```
kanej, line 154: Hardhat is going to send `slang-solx` on the wire. This line overrides the camel casing default.

Wodann (approved):

PR #1668, opened 2026-08-25
ci(bench): notify Slack on perf regressions in main baseline runs

popescuoctavian:
## Problem

The Hardhat 3 regression benchmark records every push to `main` as the new
baseline and never fails the run ("never break main"). That means a perf
regression merged to `main` was silently absorbed: the regressed numbers became
the baseline, subsequent `/bench` comparisons ran against them, and nobody was
notified — the existing Slack notification only fires for genuinely failed
baseline runs (infra errors, validation failures), never for an exceeded
threshold.

## Fix

Adds a Slack notification for perf regressions detected on `main` baseline
runs, while keeping the existing behavior of always recording the new baseline
and keeping the run green:

- `fail-on-alert` is now on for all runs (previously only for comparison runs).
  `github-action-benchmark` raises the alert *after* pushing the new data
  point, so the baseline is still recorded first.
- On baseline runs the store step gets `continue-on-error`, so the alert fails
  the step but not the job — `main` stays green.
- A new **Notify perf regression on main** step fires on
  `steps.benchmark.outcome == 'failure'` for baseline runs, reusing the
  existing Slack webhook with a distinguishable `workflow_name`
  (`… — perf regression on main (baseline was updated)`).

Comparison runs (PR `/bench`, `workflow_dispatch`) are unchanged: they still
fail on alert and never push.

## Notes

- No double notifications: an absorbed (`continue-on-error`) failure doesn't
  set the job status, so `failure()` stays false and the generic "Notify
  failures" step is skipped; conversely, if an earlier step fails, the store
  step is skipped and only the generic notification fires.
- Known trade-off: a genuine failure of the store step on `main` (e.g. a push
  error to the results repo) now also surfaces as the regression ping instead
  of the generic failure one — you still get exactly one notification either
  way.
- The Slack side is untouched; since the webhook template reads
  "Github Action workflow {workflow_name} failed", the regression message will
  read slightly awkwardly. A dedicated Slack variable could clean that up
  later if desired.

### DIFF
--- a/.github/workflows/hh3-regression-benchmark.yml
+++ b/.github/workflows/hh3-regression-benchmark.yml
@@ -392,7 +392,15 @@ jobs:
             || echo "(no regression-report.json produced)"
 
       - name: Store benchmark result
+        id: benchmark
         uses: benchmark-action/github-action-benchmark@52576c92bccf6ac60c8223ec7eb2565637cae9ba # v1.22.1
+        # On baseline runs the alert fires *after* the new data point is pushed,
+        # so absorbing the failure here keeps main green while still recording
+        # the baseline; the step's `outcome` drives the Slack regression alert
+        # below. Side effect: a genuine failure of this step on main (e.g. a
+        # push error) also surfaces as that regression ping instead of the
+        # generic failure one.
+        continue-on-error: ${{ needs.setup.outputs.is_baseline == 'true' }}
         with:
           tool: customSmallerIsBetter
           output-file-path: hardhat/regression-report.json
@@ -408,9 +416,27 @@ jobs:
           # Only push a new baseline on main-branch pushes; PRs/dispatch only compare.
           auto-push: ${{ needs.setup.outputs.is_baseline == 'true' }}
           alert-threshold: "110%"
-          # Fail the run on a regression for comparisons; never break main.
-          fail-on-alert: ${{ needs.setup.outputs.is_baseline != 'true' }}
+          # Fail the step on a regression. For comparisons (PR `/bench`,
+          # dispatch) this fails the run; for baseline runs `continue-on-error`
+          # above keeps main green and the failed outcome only triggers the
+          # Slack notification below.
+          fail-on-alert: "true"
           summary-always: true
+
+      - name: Notify perf regression on main
+        # Baseline runs record the new (regressed) data point as the baseline
+        # and stay green, so ping Slack explicitly when the store step raised
+        # a regression alert.
+        if: ${{ needs.setup.outputs.is_baseline == 'true' && steps.benchmark.outcome == 'failure' }}
+        uses: slackapi/slack-github-action@0d95c9a7becc1e6e297d76df9bc735c44f4cbcbc # v3.0.5
+        with:
+          webhook: ${{ secrets.GH_ACTION_NOTIFICATIONS_SLACK_WEBHOOK_URL }}
+          webhook-type: webhook-trigger
+          payload: |
+            {
+              "workflow_name": "${{ github.workflow }} — perf regression on main (baseline was updated)",
+              "run_url": "${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}"
+            }
 
       - name: Notify failures
         # Only ping Slack for failed baseline runs (pushes to main)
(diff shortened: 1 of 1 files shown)


Wodann (approved):
As discussed, I pushed a revision as a single commit on top of your branch that I think we should fold in. It avoids reduces the amount of errors that are indistinguishable from a reported regression. Details below:

## The problem

`continue-on-error` is all-or-nothing per step. Absorbing the regression alert means also absorbing every *other* way that step can fail — a malformed `regression-report.json`, an unreachable results repo, an expired `BENCHMARK_GITHUB_TOKEN`, a rejected push.

So as written, a genuine store failure on `main` leaves the run **green** and pings Slack with `— perf regression on main (baseline was updated)`, which is false on both counts: there was no regression, and no baseline was recorded. The PR notes the mislabelled ping as a known trade-off, but the run staying green is the part that bites. If the PAT expires, `main` stays green, baselines silently stop being recorded, and Slack reports a perf regression on every push. Nothing looks broken, so the pipeline can be dead for days.

## The fix

Split the single invocation into a read and a write, so the two failure classes get independent policies:

| Step | `auto-push` | `fail-on-alert` | `continue-on-error` | Token |
|---|---|---|---|---|
| Compare benchmark result against baseline | `false` | `true` | baseline runs only | `GITHUB_TOKEN` |
| Record baseline in the results repo | `true` | `false` | never | `BENCHMARK_GITHUB_TOKEN` |

A failure in `Compare` is (almost always) the alert, and is absorbed so `main` stays green. A failure in `Record` is always a real store failure, and reddens the run. Comparison runs (`/bench`, dispatch) are unchanged — `Record` is gated on `is_baseline`, so they still just compare and still fail on a regression.

Behaviour across the failure modes:

| | pre-PR single step | this PR's single step | split |
|---|---|---|---|
| `fail-on-alert` on baseline | off | on + `continue-on-error` | on (compare) / off (record) |
| Regression | green, **no ping** | green, regression ping ✅ | green, regression ping ✅ |
| Unreadable report | red, generic ✅ | **green**, "regression" ❌ | red, generic ✅ |
| Results repo unreachable | red, generic ✅ | **green**, "regression" ❌ | red, generic ✅ |
| Push/auth failure | red, generic ✅ | **green**, "regression" ❌ | red, generic ✅ |
| Transient read-phase error | red, generic ✅ | green, "regression" ❌ | green, "regression" ❌ |

The pre-PR version had correct infra handling and no regression detection; this PR trades the former for the latter. The split gets both, at the cost of the last row.

## Also in the commit

- **The two Slack steps collapse into one**, `Notify failures and regressions`, keyed on `failure() || steps.compare.outcome == 'failure'`. With the split, those two cases are cleanly separable, so a single step with a conditional suffix is simpler than two steps that only avoid double-firing by accident. The regression wording additionally requires `Record` to have succeeded, so an infra error that trips both steps reads as a plain failure.
- **`errors: true` on the Slack step.** It defaults to `false`, meaning a rejected delivery only logs a warning. On the regression path the run is deliberately green, so Slack is the only trace — a dropped ping would leave nothing at all.
- **A `Clear the results-repo clone` step between the two invocations.** The action clones the results repo to `./benchmark-data-repository` and only runs its cleanup on the push-retry path, so the directory survives; `git clone` then exits 128 on the non-empty destination and the action throws. Without this, `Record` would fail on every baseline run.
- **A stale comment fix** on the job's `permissions:` block, which referenced the now-deleted `Store benchmark result` step.

## Known limitations

- **A transient read-phase failure still reports as a regression** (last row above). `Compare` can fail on a fetch or clone error rather than on the alert, and if that error has cleared by the time `Record` runs, you get a green run and a "perf regression" ping with no regression. The step's error annotation on the run says which it was, so it's a false alarm rather than a wrong conclusion. I don't think it's closable without an output the action doesn't expose — it declares none.
- **The Slack message still reads awkwardly.** The template is `Github Action workflow {workflow_name} failed`, so the regression ping asserts a failure on a run that deliberately succeeded. The parenthetical is what stops it parsing as "the regression failed". Fixing it properly needs a second template variable on the Slack side; worth a follow-up issue rather than blocking this.
- **`edr-benchmark.yml` has the identical gap** — same action, same results repo, same 110% threshold, `fail-on-alert` still off for `main`. Out of scope here, but it'll want the same treatment.

## Verification

Checked against the pinned action's `dist/` at `52576c9` rather than from memory: the push happens before `handleAlert` throws; `prevBench` is the most recent entry with a *different* commit id, so the comparison is invariant to whether `Record` has already pushed this SHA (which is why the ordering is safe); `handleAlert` early-returns when both `comment-on-alert` and `fail-on-alert` are off, so `Record` does no comparison work and writes no duplicate summary. `actionlint`, `zizmor` and `prettier` all pass.

PR #1675, opened 2026-08-27
refactor(tests): extract shared provider helpers

anaPerezGhiglia:
- create new `tests::common::provider` module
- extract `new_provider` and similar customizable functions, together with `send_transaction` helper into it

### DIFF
--- a/crates/edr_provider/tests/common/mod.rs
+++ b/crates/edr_provider/tests/common/mod.rs
@@ -1,8 +1,6 @@
 /// Test utilities for blob transactions.
 #[allow(dead_code)]
 pub mod blob;
-/// Test utilities for constructing an L1 provider and issuing common requests.
-#[cfg(feature = "test-utils")]
 pub mod provider;
 
 pub fn help_test_method_invocation_serde<MethodInvocation>(call: MethodInvocation)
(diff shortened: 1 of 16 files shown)


crates/edr_provider/tests/common/mod.rs
```diff
@@ -1,6 +1,9 @@
 /// Test utilities for blob transactions.
 #[allow(dead_code)]
 pub mod blob;
+/// Test utilities for constructing an L1 provider and issuing common requests.
+#[cfg(feature = "test-utils")]
```
Wodann: suggestion: move this inside the `provider.rs` file as a module-level:

```
//! Test utilities for constructing an L1 provider and issuing common requests.
#![cfg(feature = "test-utils")]
```

If the file ever gets moved, the comment and cfg go with it. Moreover, all information is localised to the file itself, making it easier to parse (no jumping to other places to get the full context).

Wodann (approved):
Thanks for cleaning this up. Much better!

One suggestion, but you can address that without asking for a re-review.

crates/edr_provider/tests/common/mod.rs
anaPerezGhiglia: addressed by [754972a](https://github.com/NomicFoundation/edr/pull/1675/commits/754972afde219225372733674ef60e5eebe07381)

PR #1686, opened 2026-08-29
fix: validate interval mining and always restart its timer

Wodann:
Follow-up to #1486, and stacked on it.

#1486 moved JSON-RPC request handling onto a dedicated provider thread. In review, @anaPerezGhiglia found two pre-existing interval-mining validation holes that the new design escalates from "degraded" to "provider unusable", because the thread that mines is now the thread that serves requests. She asked for them as a separate PR, which is this one.

## The defects

**A `[0, 0]` range starved the provider.** `generate_interval()` returned 0, `crossbeam_channel::after(Duration::ZERO)` is delivered instantly, and the event loop's interval arm is checked before its request arm — so requests were never serviced and their JS promises never settled while the loop hot-mined. Before #1486 the same configuration busy-mined but requests still progressed, because the data mutex was fair.

**A `min > max` range panicked the provider.** `rand::random_range` panics on an empty range. The JSON-RPC path rejected it, but the provider-construction path did not, and Hardhat forwards such a configuration without complaint. Before #1486 that panic killed a detached task; now it kills the thread that owns `ProviderData`.

**`evm_setIntervalMining` stopped restarting the timer for an unchanged value.** Not raised in review. #1486 re-armed only when the new configuration differed, so a client polling `evm_setIntervalMining(5000)` every 4s mined every 5s where it previously never mined. Equality was also the wrong question to ask: a range draws a new interval before each block, so setting `[1000, 5000]` twice kept the first draw instead of re-drawing.

## The fix

An interval range is now a type with private, ordered, non-zero bounds, and `generate_interval` returns `NonZeroU64`. A zero-length timer and an empty range are unrepresentable rather than merely unreached. `#[serde(try_from = ..)]` routes deserialization through an unchecked shim, which closes a third entry point that is easy to miss: a recorded scenario file carries a serialized `MiningConfig` that the scenario runner deserializes straight into a live configuration.

The wire format is unchanged — a newtype variant wrapping a struct with `min`/`max` serializes identically to the struct variant it replaces — so existing recorded scenarios keep loading. A test pins both directions.

Reconfiguration is now signalled by a take-once flag on `ProviderData` rather than inferred from value equality. That is the actual predicate, and it handles rejection correctly for free: the conversion error propagates before the setter is reached, so an invalid value never restarts the timer.

## No scheduling change

The review also asked whether a valid-but-small interval throttles request handling. It does not, once zero is impossible. `crossbeam_channel::after` computes its deadline at the call site, which is *after* mining completes, so the next deadline is always a full interval away and every cycle leaves a window in which the timer is not ready. Requests get roughly `interval / (interval + mining time)` of the loop. Only a zero interval breaks that, which is what this PR makes unrepresentable.

## Breaking change

`[0, 0]` and `[0, N]` interval ranges are no longer accepted. Neither worked as configured before, but a configuration that is forwarded today will now fail when the provider is created.

## Validation

- `cargo test -p edr_provider --features test-utils`: 302 tests pass, including 17 new unit tests covering every accepted and rejected input, the wire format in both directions, and that deserialization is closed.
- `evm_set_interval_mining_restarts_on_an_unchanged_interval` mines 3 blocks instead of 0 against the previous logic, and its positive control fails if nothing re-arms at all.
- `evm_set_interval_mining_rejects_an_invalid_range` asserts the provider is still usable after a rejection — the regression test for the panic.
- Scenario replay is unchanged: 5079 success / 193 failure on `safe-contracts`, the same as before.

## Reviewer

Question: Given that this is a bug fix, does this qualify as a "BREAKING CHANGE"?

### DIFF
--- a/crates/edr_provider/src/config.rs
+++ b/crates/edr_provider/src/config.rs
@@ -153,9 +153,8 @@ impl From<IntervalRangeConfig> for IntervalConfig {
 
 /// An inclusive range of interval-mining intervals, in milliseconds.
 ///
-/// Non-empty and free of zeroes by construction, so
-/// [`IntervalRangeConfig::generate_interval`] can neither panic nor return
-/// zero. Deserialization upholds the same invariant.
+/// Non-empty and free of zeroes by construction, including through
+/// deserialization.
 #[derive(Clone, Debug, Deserialize, Serialize)]
 #[serde(try_from = "UncheckedIntervalRange")]
 pub struct IntervalRangeConfig {
@@ -190,6 +189,8 @@ impl IntervalRangeConfig {
 
     /// Draws an interval uniformly from the range, in milliseconds. Both
     /// bounds are inclusive.
+    ///
+    /// Never returns zero and never panics.
     pub fn generate_interval(&self) -> NonZeroU64 {
         // `min <= max` is an invariant of the type, so the subtraction cannot
         // underflow and `0..=span` is never empty.
--- a/crates/edr_provider/src/config.rs
+++ b/crates/edr_provider/src/config.rs
@@ -115,49 +115,145 @@ pub struct LocalConfig {
 }
 
 /// Configuration for interval mining.
+///
+/// Every representable value yields a non-zero interval.
 #[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
-#[serde(rename_all_fields = "camelCase")]
 pub enum IntervalConfig {
+    /// Mine a block every `n` milliseconds.
     Fixed(NonZeroU64),
-    Range { min: u64, max: u64 },
+    /// Mine a block every `n` milliseconds, where `n` is drawn from the range
+    /// anew before each block.
+    Range(IntervalRangeConfig),
 }
 
 impl IntervalConfig {
-    /// Generates a (random) interval based on the configuration.
-    pub fn generate_interval(&self) -> u64 {
+    /// Generates a (random) interval in milliseconds, based on the
+    /// configuration.
+    pub fn generate_interval(&self) -> NonZeroU64 {
         match self {
-            IntervalConfig::Fixed(interval) => interval.get(),
-            IntervalConfig::Range { min, max } => rand::rng().random_range(*min..=*max),
+            IntervalConfig::Fixed(interval) => *interval,
+            IntervalConfig::Range(range) => range.generate_interval(),
         }
     }
 }
 
-/// An error that occurs when trying to convert [`IntervalConfigRequest`] to an
-/// `Option<IntervalConfig>`.
+impl From<NonZeroU64> for IntervalConfig {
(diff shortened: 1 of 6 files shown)


crates/edr_provider/src/config.rs
```diff
@@ -115,49 +115,145 @@ pub struct LocalConfig {
 }
 
 /// Configuration for interval mining.
-#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
-#[serde(rename_all_fields = "camelCase")]
+///
+/// Every representable value yields a non-zero interval.
+#[derive(Clone, Debug, Deserialize, Serialize)]
```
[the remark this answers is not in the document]
Wodann: No API stability guaranteed in Rust

crates/edr_provider/src/config.rs
```diff
@@ -115,49 +115,145 @@ pub struct LocalConfig {
...
-    Range { min: u64, max: u64 },
+    /// Mine a block every `n` milliseconds, where `n` is drawn from the range
+    /// anew before each block.
+    Range(IntervalRangeConfig),
 }
 
 impl IntervalConfig {
-    /// Generates a (random) interval based on the configuration.
-    pub fn generate_interval(&self) -> u64 {
+    /// Generates a (random) interval in milliseconds, based on the
```
anaPerezGhiglia: nitpick: let's update the doc since we are at it. On fixed config or interval with equal bounds the interval is not random

crates/edr_provider/src/config.rs
```diff
@@ -115,49 +115,145 @@ pub struct LocalConfig {
...
+impl From<IntervalRangeConfig> for IntervalConfig {
+    fn from(value: IntervalRangeConfig) -> Self {
+        Self::Range(value)
+    }
+}
+
+/// An inclusive range of interval-mining intervals, in milliseconds.
+///
+/// Non-empty and free of zeroes by construction, so
+/// [`IntervalRangeConfig::generate_interval`] can neither panic nor return
```
anaPerezGhiglia: Let's avoid mentioning the caller in the struct documentation. If we want to document this, then it should be in the caller's site, not here

anaPerezGhiglia (approved):
Thanks for fixing this!

crates/edr_provider/src/config.rs
Wodann: Addressed by https://github.com/NomicFoundation/edr/pull/1686/changes/16613d07bf54fb2e0d91b8ff6290acbc02676800

crates/edr_provider/src/config.rs
Wodann: Addressed by https://github.com/NomicFoundation/edr/pull/1686/changes/4aace64380ab69c0cd698f2fe9c64b7d74c100d6

PR #1689, opened 2026-08-30
feat(solidity-tests): implement `vm.broadcastRawTransaction`

### LINKED ISSUE #1690: vm.broadcastRawTransaction not implemented
edr does not implement `vm.broadcastRawTransaction`

would be great to have it for parity with [foundry](https://www.getfoundry.sh/reference/cheatcodes/broadcast-raw-transaction) and being able to for example setup deterministic factories using Nick's method (https://github.com/Arachnid/deterministic-deployment-proxy)

wighawag:
Fixes #1690.

## Summary

Implements the `vm.broadcastRawTransaction(bytes)` cheatcode in the Solidity test runner. It was declared in the `Vm` interface but routed to the unsupported-cheatcode handler alongside the script-only broadcast cheatcodes.

The cheatcode decodes an RLP-encoded signed transaction, recovers the sender from its signature, and executes it against the current EVM state as its own transaction. Nothing is broadcast to a network. This is the same behavior as Foundry's cheatcode in a test context; the broadcast queue it feeds in scripts does not exist in EDR, so that half is omitted.

## Motivation

Submitting an already-signed transaction is the only way to replay a pre-signed deterministic-deployment bootstrap (Nick's method). The deployed address is bound to a keyless sender recovered from a fixed signature at nonce 0, so the deployment cannot be re-encoded as an ordinary call or reproduced with `vm.prank`.

## Changes

- **`foundry_evm_core::backend`**: new `CheatcodeBackend::transact_from_tx`, the non-forked counterpart of `transact`. It commits the in-flight journaled state, executes the request in a nested EVM with the cheatcode inspector at the current call depth plus one, commits the result, and refreshes the loaded accounts. `CowBackend` forwards to it. This mirrors upstream Foundry's implementation.
- **`foundry_cheatcodes`**: `broadcastRawTransaction` decodes with `TxEnvelope::decode`, recovers the signer up front so an unrecoverable signature reports its actual reason, converts to a `TransactionRequest`, and calls `transact_from_tx`. The sender is taken from the request's `from`, as upstream does. Removed from the unsupported list.
- **Spec**: the cheatcode moves to a Scripting section with a doc comment describing what it actually does and its non-obvious semantics (below). `cheatcodes.json` regenerated.
- **Changeset**: minor bump for `@nomicfoundation/edr`.


popescuoctavian:
@wighawag thank you for the PR!

I've made a few changes:
- Fixed uncommitted test state leaking into the database through the cheatcode, an issue also present in Foundry. The nested EVM now runs on a seeded clone, and only what the transaction touched is mirrored back. Added tests for this.
- Ported five tests from upstream Foundry. Script tests were skipped (`startBroadcast` is unsupported).
- Dropped the caller parameter from `transact_from_tx`: the sender comes from the request's `from`.
- Small doc rewrites.

crates/foundry/evm/core/src/backend/mod.rs
```diff
@@ -1722,6 +1767,71 @@ impl<
...
+        // Flush the in-flight journal into the database so the nested EVM sees
+        // the state as of the current point in the test.
+        self.commit(journaled_state.state.clone());
+
+        let res = {
+            configure_tx_req_env(&mut env.tx, tx, Some(caller))?;
+            let mut env = self.env_with_handler_cfg(env);
+
+            let db = self.clone();
+            let mut journal = Journal::<_, JournalEntry>::new(db);
```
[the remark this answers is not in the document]
popescuoctavian: Addressed by e17ab10e872d1163b417796779602f860002b9f3.

crates/foundry/cheatcodes/src/evm.rs
```diff
@@ -2851,6 +2854,63 @@ impl Cheatcode for getStateDiffJsonCall {
...
+    ) -> Result {
+        let Self { data } = self;
+
+        let tx = TxEnvelope::decode(&mut data.as_ref())
+            .map_err(|err| fmt_err!("failed to decode RLP-encoded transaction: {err}"))?;
+
+        // The conversion below recovers the signer and sets it as `from`, so
+        // the transaction is executed from the address that signed it, and
+        // deliberately *not* from the caller of the cheatcode.
+        tx.recover_signer()
```
Wodann: suggestion: Seemingly "leaked" API documentation
```suggestion
        tx.recover_signer()
```

crates/foundry/evm/core/src/backend/mod.rs
```diff
@@ -1722,6 +1768,87 @@ impl<
...
+
+        let res = {
+            configure_tx_req_env(&mut env.tx, tx, None)?;
+            let mut env = self.env_with_handler_cfg(env);
+
+            // Execute against a copy of the backend seeded with the current
+            // journaled state, so the transaction sees the state at this point
+            // of the test. Seeding the copy rather than `self` keeps the test's
+            // uncommitted changes out of the database: they may still be
+            // reverted by the test, in which case nothing would undo them.
```
Wodann: suggestion: a (seemingly) over-verbose Claude comment

```suggestion
            // Use a clone of the backend, seeded with the current journaled
            // state, to keep uncommitted changes out of the database if a
            // test reverts.
```

crates/foundry/evm/core/src/backend/mod.rs
```diff
@@ -1722,6 +1768,87 @@ impl<
...
+            EvmBuilderT::evm_with_journal_and_inspector(journal, env, inspector)
+                .inspect_tx(tx)
+                .wrap_err("backend: failed committing transaction")?
+        };
+
+        // Only *validation* failures are errors; a transaction that reverts
+        // during execution is still an included transaction, so its changeset
+        // (nonce bump and fee payment) is committed either way. Traced because
+        // there is otherwise no signal that the transaction did not succeed.
+        if !res.result.is_success() {
```
Wodann: suggestion: seemingly, a Claude explanation for this code. I don't think there needs to be an explanation for why this is traced.

```suggestion
```

crates/foundry/evm/core/src/backend/mod.rs
```diff
@@ -1722,6 +1768,87 @@ impl<
...
+            trace!(result = ?res.result, "broadcast transaction did not succeed");
+        }
+
+        // Commit the transaction's changes and mirror them into the loaded
+        // accounts, but only the accounts and slots the transaction touched.
+        // The database never received the test's uncommitted changes, so for
+        // everything else the journaled state is the only source of truth and
+        // must not be overwritten from the database.
```
Wodann: suggestion: (seemingly) overly verbose Claude comment

```suggestion
        // Only overwrite the journaled state for accounts and slots that the
        // transaction touched.
```

crates/foundry/cheatcodes/src/evm.rs
Wodann, line 2895: issue: as far as I can tell, this [function is immutable](https://docs.rs/alloy-consensus/2.4.2/alloy_consensus/transaction/trait.SignerRecoverable.html) and from is never set.

Am I missing something?

Wodann (changes requested):

crates/foundry/evm/core/src/backend/mod.rs
popescuoctavian: Applied inline suggestion: e02d2b0f29aac51313a3948216d637be04d132ca

crates/foundry/evm/core/src/backend/mod.rs
popescuoctavian: Applied suggestion: 13c2a0cfdc07565df60b9be837db7ef9847ac700

crates/foundry/evm/core/src/backend/mod.rs
popescuoctavian: Applied inline suggestion: e9f4a82beb8803bb28dd67c733a4594529b18501

crates/foundry/cheatcodes/src/evm.rs
popescuoctavian: This code is a validation step and is meant to fail early with the cheatcode's own error message when the signature is invalid - it doesn't provide any functionality. The `tx.into()` below recovers the signer and calls `.from(signer)` on the request before returning it (ends up in the `k256` feature branch below):
```rust
TxEnvelope::Legacy(tx) => {
    #[cfg(feature = "k256")]
    {
        let from = tx.recover_signer().ok();
        let tx: Self = tx.strip_signature().into();
        if let Some(from) = from { tx.from(from) } else { tx }
    }
    #[cfg(not(feature = "k256"))]
    { tx.strip_signature().into() }
}
```

That conversion absorbs recovery failures with `.ok()`. Without the previous check, a transaction with a bad signature would surface later without a meaningful error message.

This is what the comment above was trying to explain, I'll try to make it clearer.

popescuoctavian: Initial comment was meant to describe what `tx.into()` below does, but did not explain why `tx.recover_signer()` was added. I've removed that and replaced it with `tx.recover_signer()`'s purpose: e7d7bb113452e1d03e8c134f159643ba1e0e353e

Wodann (approved):

PR #1700, opened 2026-08-31
fix: don't panic on a fuzz counterexample without a recorded arena

Wodann:
Two commits: the first adds an end-to-end reproduction of the panic and fails until the second fixes it (validated at each commit).

**`test: reproduce the fuzz counterexample panic on an executor error`**

An executor error during a fuzz run — as opposed to a reverting call — produces a counterexample without a recorded trace arena, and the stack-trace computation in `FunctionRunner::run_fuzz_test` then panics on the missing arena when the tracer records steps.

Reproduce it end to end: the fuzzed call runs on a fork whose mock RPC serves fork creation and every account fetch except the one address that only the fuzzed call reads, so `call_raw` itself fails with a database error. The test pins the intended outcome — the failure's reason carries the executor error and there is no stack trace — and fails with the panic until the following commit fixes it.

**`fix: don't panic on a fuzz counterexample without a recorded arena`**

`FunctionRunner::run_fuzz_test` expects the counterexample's call to have recorded a trace arena whenever the tracer records steps. An executor error — as opposed to a reverting call — breaks that expectation: `single_fuzz` turns it into a test-case failure without a counterexample outcome, so the counterexample is built without traces, `execution_traces` stays empty, and the `split_last().expect(..)` panics. It became reachable when the failing arena was named explicitly, because the suite's setup arenas are no longer chained into the same list.

Handle it like the invariant campaign that retains no arena: nothing describes the failure, so there is no stack trace to compute, and the result's `reason` carries the executor error — now with its error chain: `to_string` on the `eyre` report yielded only the outermost "EVM error" layer, hiding the cause. An EVM-level failure of the call itself surfaces as a revert with a recorded arena; only an executor error, such as the fork database failure in the preceding commit's reproduction, produces a traceless counterexample. That reproduction now passes.

---
Part of the Solidity test runner memory-reduction series; stacked on #1623 (merged).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3604,6 +3604,7 @@ dependencies = [
  "futures",
  "humantime-serde",
  "itertools 0.15.0",
+ "mockito",
  "num-traits",
  "once_cell",
  "parking_lot",
--- a/crates/edr_solidity_tests/Cargo.toml
+++ b/crates/edr_solidity_tests/Cargo.toml
@@ -54,6 +54,7 @@ paste = "1.0"
 serde_json.workspace = true
 semver.workspace = true
 serial_test = "4.0.0"
+mockito = { version = "1.7", default-features = false }
 tempfile.workspace = true
 tracing-subscriber = { workspace = true, features = ["env-filter", "fmt"] }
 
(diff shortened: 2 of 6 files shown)


popescuoctavian (approved):
LGTM, thanks for addressing this!

PR #1701, opened 2026-08-31
docs: state the contracts the stack-trace computation relies on

Wodann:
## Problem / context

#1623 made `get_stack_trace` take the failing arena explicitly. That turned two implicit assumptions into contracts the code relies on, and it silently changed one documented contract. None of this was written down where the next reader needs it.

## Solution

Documentation only, plus one regression test:

- `TestResult::execution_traces` is ordered by execution and its last arena is the failing call. Six `split_last` sites across two crates rely on this, so the field doc now states it. The same holds for a failed suite's `SetupTraces`.
- `stack_trace_result` (and the napi `stackTrace()` mirror) claimed `None` implies the test succeeded or was skipped. A failure is also `None` when no retained arena describes it; `reason` then explains the failure. That happens for an assume-rejected invariant campaign, a campaign that failed before its first call, and a fuzz counterexample produced by an executor error (#1700).
- `get_stack_trace` returns `Ok` with an empty vec when the heuristics cannot explain the failure; callers present that as a failed heuristic. It also needs the failing arena to carry recorded EVM steps, since only the calldata- and metadata-based inferences work without them.

## Validation

- New repro `always_mode_reports_no_stack_trace_without_a_failing_arena`: an invariant campaign whose calls are all rejected by `vm.assume` fails with a reason and no stack trace. It pins the previously untested `None`-on-failure case.

## Notes for reviewer

- Stacked on #1700 (merged); part of the Solidity test runner memory-reduction series.
- No behavior changes. The one code addition is the test fixture (`AlwaysStackTraceInvariantNoCallTest`).

### DIFF
--- a/crates/edr_napi/index.d.ts
+++ b/crates/edr_napi/index.d.ts
@@ -146,7 +146,10 @@ export declare class TestResult {
    * Compute the error stack trace.
    * The result is either the stack trace or the reason why we couldn't
    * generate the stack trace.
-   * Returns null if the test status is succeeded or skipped.
+   * Returns null if the test succeeded or was skipped. Also returns null
+   * when the failure left no trace behind, for example an invariant
+   * campaign whose calls were all rejected by `vm.assume`; the `reason`
+   * field then explains the failure.
    * Cannot throw.
    */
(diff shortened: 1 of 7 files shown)


popescuoctavian (approved):
LGTM, thanks for adding this, it makes the functionality easier to follow and understand.

PR #1702, opened 2026-08-31
perf: free unconsumed trace arenas as soon as each test finishes

Wodann:
## Problem / context

Every test's `SparsedTraceArena`s stayed resident in `TestResult::execution_traces` until the whole suite completed and the napi conversion filtered them — regardless of `include_traces`. Under `CollectStackTraces::Always` (Hardhat `-vvv` and above) each arena also carries per-opcode EVM steps, which is the difference between megabytes and gigabytes: solady peaked at 5.2 GiB at `-vvv` and OOMed a 16 GiB machine at `-vvvv`.

## Solution

Introduce `TraceRetentionPolicy`, which decides what a finished test's arenas are still needed for, and free the rest as soon as the test returns. The freeing lives in `TestResult::free_unconsumed_traces`, which destructures the result exhaustively: adding a field to `TestResult` then fails to compile there, forcing a decision on the new field's retention. It runs after stack-trace generation — the last consumer of the arenas it frees — and frees `execution_traces` when neither `include_traces` nor the gas report will consume them, plus the counterexample arenas, which have no consumer once the test has finished.

The gas-report samples move off `TestResult` entirely, since their one consumer is the gas-report pass in `run_test_suite`. `FunctionRunner` returns a `TestRunOutcome` carrying them beside the result, and `run_tests` returns the outcomes as a `SuiteRunOutcome`. The suite pass destructures each outcome, analyzing or dropping its samples within that loop iteration, and assembles the `SuiteResult` from the results afterwards. A sample therefore cannot outlive its test's turn in that pass; nothing frees a field afterwards, and no assertion is needed to tie samples to their consumer. Without a gas report the field is `None`: `FunctionRunner` gates the collection on the retention policy, so nothing is collected only to be dropped. Without a gas report the samples are dropped as each test finishes, on the rayon worker that allocated them. The old `clear()` sat inside the decode loop of `MultiContractRunner::run_test_suite`, so it never ran at `IncludeTraces::None` and the loop's `continue` skipped it for passing tests at `Failing`.

Since the retention decision now destroys arenas, the decode loop and the napi filter route through the same `IncludeTraces::should_include` predicate. The two hand-written variants disagreed on skipped tests: the decode loop decoded a skipped test's arenas at `Failing` even though the napi filter then dropped them. The fuzz and table-test paths also stop cloning the failing arena into `BaseCounterExample::traces`, which nothing reads; the setup-failure path moves rather than clones the setup traces.

## Validation

- Measured on its own (solady, peak RSS, 4 rayon threads): `-vvv` 5.2 → 4.5 GiB. `-vvvv` still OOMs; #1703 addresses that by stripping steps from the arenas that are retained.
- New regression tests in `repros.rs` cover freeing under `Always` mode, including that arenas nothing consumes are freed and that stack traces still come out.

## Notes for reviewer

- Stacked on #1701; part of the Solidity test runner memory-reduction series.
- The policy runs after stack-trace generation on purpose: that generation is the last reader of the freed arenas.
- The unified `should_include` predicate slightly changes skipped-test handling, as described above; it removes wasted decoding rather than any surfaced data.

### DIFF
--- a/crates/edr_solidity_tests/src/runner.rs
+++ b/crates/edr_solidity_tests/src/runner.rs
@@ -17,6 +17,7 @@ use edr_artifact::ArtifactId;
 use edr_chain_spec::{EvmHaltReason, HaltReasonTrait};
 use edr_decoder_revert::RevertDecoder;
 use edr_solidity::{
+    config::IncludeTraces,
     contract_decoder::SyncNestedTraceDecoder,
     solidity_stack_trace::{get_stack_trace, DeployedCode, StackTraceEntry},
 };
@@ -44,7 +45,9 @@ use foundry_evm::{
         invariant::{CallDetails, InvariantContract},
         CounterExample, FuzzFixtures,
     },
-    traces::{load_contracts, SetupTraceKind, SetupTraces, SparsedTraceArena, TracingMode},
+    traces::{
+        load_contracts, CallTraceArena, SetupTraceKind, SetupTraces, SparsedTraceArena, TracingMode,
+    },
 };
 use itertools::Itertools;
 use proptest::test_runner::{FailurePersistence, RngAlgorithm, TestError, TestRng, TestRunner};
@@ -56,8 +59,9 @@ use crate::{
     error::TestRunnerError,
     fuzz::{invariant::BasicTxDetails, BaseCounterExample, FuzzConfig},
     multi_runner::TestContract,
-    result::{SuiteResult, TestResult, TestSetup},
+    result::{SuiteResult, SuiteRunOutcome, TestResult, TestRunOutcome, TestSetup, TestStatus},
     revm::context::result::HaltReason,
+    trace_retention::TraceRetentionPolicy,
     TestFilter, TestFunctionConfigOverride,
 };
 
@@ -109,6 +113,8 @@ pub struct ContractRunner<
     /// Whether gas reports are being generated. When enabled, isolation cannot
     /// be disabled by function-level overrides.
     generate_gas_report: bool,
+    /// Which trace arenas to retain once a test has finished.
+    trace_retention: TraceRetentionPolicy,
 
     #[allow(clippy::type_complexity)]
     _phantom: PhantomData<fn() -> (EvmBuilderT, HaltReasonT, TransactionErrorT)>,
@@ -134,6 +140,8 @@ pub struct ContractRunnerOptions<'a> {
     /// Whether gas reports are being generated. When enabled, isolation cannot
     /// be disabled by function-level overrides.
     pub generate_gas_report: bool,
+    /// Which test results carry call traces to the caller.
+    pub include_traces: IncludeTraces,
 }
 
 /// Contract artifact related arguments to the contract runner.
@@ -197,6 +205,7 @@ impl<
             invariant_config,
             test_function_overrides,
             generate_gas_report,
+            include_traces,
         } = options;
 
         Self {
@@ -217,6 +226,7 @@ impl<
             span,
             test_function_overrides,
             generate_gas_report,
+            trace_retention: TraceRetentionPolicy::new(include_traces, generate_gas_report),
             _phantom: PhantomData,
         }
     }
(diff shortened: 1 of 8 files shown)


crates/foundry/evm/evm/src/executors/fuzz/mod.rs
```diff
@@ -272,7 +272,10 @@ impl<
         let last_run_traces = if run_result.is_ok() {
             traces.pop()
         } else {
-            call.traces.clone()
+            // Nothing reads `BaseCounterExample::traces`, so the failing
+            // arena can move into `FuzzTestResult::traces` rather than be
+            // cloned for both.
+            call.traces
```
Wodann, line 278: The type-level removal lands later in this series: #1709 deletes `BaseCounterExample::traces` and its constructor parameters. It cannot happen here because the invariant replay still fills the field; #1704 retires that last producer.

crates/foundry/evm/evm/src/executors/fuzz/mod.rs
```diff
@@ -322,7 +325,8 @@ impl<
                         Some(CounterExample::Single(BaseCounterExample::from_fuzz_call(
                             calldata,
                             &args,
-                            call.traces,
+                            // Nothing consumes counterexample arenas; see above.
+                            None,
```
Wodann, line 329: Same here: the `traces` parameter and the field it fills are removed at the type level in #1709, once #1704 has retired the invariant replay's writes to it.

crates/edr_solidity_tests/src/runner.rs
```diff
@@ -1081,7 +1121,9 @@ impl<
                                 .expect("args have valid abi encoding"),
                         ),
                         &args,
-                        raw_call_result.traces.clone(),
+                        // Counterexample arenas are never consumed; the
+                        // failing arena lives on in `execution_traces`.
+                        None,
```
Wodann, line 1123: Like the fuzz-path `None`: the `traces` parameter and the field it fills are removed at the type level in #1709, once #1704 has retired the invariant replay's writes to it.

crates/edr_solidity_tests/src/runner.rs
```diff
@@ -1557,7 +1610,10 @@ impl<
             None
         };
 
-        self.result
+        TestRunOutcome {
```
popescuoctavian, line 1623: Nit: I believe you can use `self.outcome()` here, which keeps it consistent with the rest of the file.

popescuoctavian (approved):
LGTM, thanks for adding this! It is a nice improvement on memory consumption and I like the retention policy model.

crates/edr_solidity_tests/src/runner.rs
Wodann: Tried it: `Self::outcome` takes `self` by value, and `self.executor` was moved into `fuzzed_executor` at the start of the run, so this fails with E0382 (use of partially moved value). The struct literal moves only the two remaining fields, which the partial-move rules allow. Added a comment at the site explaining this.

crates/edr_solidity_tests/src/runner.rs
```diff
@@ -1435,8 +1474,12 @@ impl<
             }
         }
 
+        self.gas_report_samples = self
+            .cr
+            .trace_retention
+            .retains_gas_report_samples()
+            .then_some(invariant_result.gas_report_traces);
```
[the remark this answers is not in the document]
Wodann: This compiles: moving `gas_report_traces` out and later moving `cases`/`metrics` are moves of distinct fields, which Rust allows for any non-`Drop` struct. `invariant_result` is never used as a whole afterwards. CI and `clippy -D warnings` are green on this commit.

crates/foundry/evm/evm/src/executors/fuzz/mod.rs
```diff
@@ -272,7 +272,10 @@ impl<
         let last_run_traces = if run_result.is_ok() {
             traces.pop()
         } else {
-            call.traces.clone()
+            // Nothing reads `BaseCounterExample::traces`, so the failing
+            // arena can move into `FuzzTestResult::traces` rather than be
+            // cloned for both.
+            call.traces
         };
```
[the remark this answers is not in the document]
Wodann: This compiles: `call.traces` and the later `call.labels`/`call.indeterminism_reasons` are moves of distinct fields of a non-`Drop` struct, which the partial-move rules allow; `call` is never used as a whole after the first move. Same for line 330. CI and `clippy -D warnings` are green on this commit.

crates/edr_solidity_tests/src/runner.rs
popescuoctavian: Got it, missed that. Thanks for checking!

PR #1703, opened 2026-08-31
perf: strip recorded EVM steps from arenas retained for call traces

Wodann:
## Problem / context

Recorded EVM steps — one entry per executed opcode — are only consumed by stack-trace generation, which has already run by the time a test finishes. Yet every arena retained for call traces kept its steps until the suite completed. At `IncludeTraces::All` (Hardhat `-vvvv`) that held gigabytes of steps nothing would read, and solady OOMed a 16 GiB machine.

## Solution

Add `strip_arena_steps` and `SparsedTraceArena::strip_steps`, and apply them to every arena that is kept once a test finishes: the retained `execution_traces` (in `TestResult::free_unconsumed_traces`) and the gas-report samples (in `run_tests`, where they are kept beside the result), which at `CollectStackTraces::Always` carry steps too. Every remaining consumer reads the call tree, the logs and their ordering: trace decoding never looks at steps, the gas report walks nodes, and the napi conversion discards `TraceMemberOrder::Step`.

`strip_steps` first resolves the ignored `pauseTracing`/`resumeTracing` ranges in place, because they are keyed by position in each node's `ordering` and dropping step entries would shift them. Resolving in place avoids `resolve_arena`'s clone, so the arena's footprint does not double at the moment it is being shrunk; the arena also becomes non-sparse, so the napi conversion's later `resolve_arena` is a borrow. The `ordering` vectors are shrunk as well, since `retain` keeps the capacity that grew with one entry per step. `clear_node` moves out of `resolve_arena` so both the cloning and the in-place resolution can call it.

## Validation

- Measured on top of #1702 (peak RSS, 4 rayon threads): solady `-vvvv` no longer OOMs and peaks at 4.4 GiB; uniswap-v4-core `-vvvv` 3.4 → 2.4 GiB. `-vvv` is unchanged within noise, since only failing tests retain arenas there.
- `strip_steps_matches_resolving_then_dropping_steps` pins that in-place resolution plus stripping equals the old resolve-then-drop result, node by node.

## Notes for reviewer

- Stacked on #1702; part of the Solidity test runner memory-reduction series.
- `strip_steps` must not run on a still-sparse arena, since the ignored ranges are positional; the doc on `strip_arena_steps` states when each entry point is safe.

### DIFF
--- a/crates/foundry/evm/traces/src/lib.rs
+++ b/crates/foundry/evm/traces/src/lib.rs
@@ -64,106 +64,38 @@ impl SparsedTraceArena {
         if self.ignored.is_empty() {
             Cow::Borrowed(&self.arena)
         } else {
-            fn clear_node(
-                nodes: &mut [CallTraceNode],
-                node_idx: usize,
-                ignored: &HashMap<(usize, usize), (usize, usize)>,
-                cur_ignore_end: &mut Option<(usize, usize)>,
-            ) {
-                // Prepend an additional None item to the ordering to handle the beginning of
-                // the trace.
-                let node = nodes
-                    .get(node_idx)
-                    .expect("node_idx should be within nodes bounds");
-                let items = std::iter::once(None)
-                    .chain(node.ordering.clone().into_iter().map(Some))
-                    .enumerate();
-
-                let mut internal_calls = Vec::new();
-                let mut items_to_remove = BTreeSet::new();
-                for (item_idx, item) in items {
-                    if let Some(end_node) = ignored.get(&(node_idx, item_idx)) {
-                        *cur_ignore_end = Some(*end_node);
-                    }
-
-                    let mut remove = cur_ignore_end.is_some() & item.is_some();
-
-                    match item {
-                        // we only remove calls if they did not start/pause tracing
-                        Some(TraceMemberOrder::Call(child_idx)) => {
-                            let node = nodes
-                                .get(node_idx)
-                                .expect("node_idx should be within nodes bounds");
-                            let &child_node_idx = node
-                                .children
-                                .get(child_idx)
-                                .expect("child_idx should be within children bounds");
-                            clear_node(nodes, child_node_idx, ignored, cur_ignore_end);
-                            remove &= cur_ignore_end.is_some();
-                        }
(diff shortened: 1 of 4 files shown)


crates/foundry/evm/traces/src/lib.rs
```diff
@@ -215,6 +147,130 @@ impl TracingMode {
...
+}
+
+/// Discards the recorded EVM steps of every node in `arena`, keeping the rest
+/// of the call tree: its nodes, their logs and their ordering. See
+/// [`SparsedTraceArena::strip_steps`] for when this is safe.
+///
+/// Must not be applied to the arena of a still-sparse [`SparsedTraceArena`]:
+/// its ignored ranges are keyed by position in `ordering`, which this shifts.
+/// Use [`SparsedTraceArena::strip_steps`] there instead.
+pub fn strip_arena_steps(arena: &mut CallTraceArena) {
```
Wodann, line 264: The "must not be applied to a still-sparse arena" rule is a convention here, not a type guarantee. A `strip_steps(self) -> CallTraceArena` signature would make the mistake unwritable, which I would prefer.

However, in this PR it buys nothing on its own: the stripped arenas are stored back into `TestResult::execution_traces: Vec<SparsedTraceArena>`, so each returned `CallTraceArena` would be re-wrapped with an empty `ignored` map. The mistake-proofing only materializes once the holder distinguishes the two states in its type.

That holder arrives in #1704, which wraps the collection in an `ExecutionTraces` newtype. There it can be reshaped as:

```rust
pub struct ExecutionTraces {
    /// Displaced arenas: EVM steps stripped, ignored ranges resolved.
    prior: Vec<CallTraceArena>,
    /// The most recent arena — the only one that may carry steps.
    last: Option<SparsedTraceArena>,
}
```

`push` would call `strip_steps(self) -> CallTraceArena` on the displaced `last` and move it into `prior`. A stripped arena is then a `CallTraceArena` by type, and the napi conversion's `resolve_arena` disappears for the prior arenas. The cost is that consumers can no longer iterate one uniform slice: the decode, napi and gas-report loops need a two-part iteration. Setup traces would want the same treatment.

popescuoctavian (approved):
Looks good, thanks!

PR #1704, opened 2026-08-31
perf: strip EVM steps from displaced arenas as traces accumulate

Wodann:
## Problem / context

Arenas kept their recorded steps until the test finished and `TraceRetentionPolicy` stripped them. A test that records several arenas — a `beforeTestSetup` chain, an invariant replay sequence — therefore held all of them step-laden at once. Only the last arena of a collection is ever the failing trace of a stack-trace computation; the earlier ones are walked for their CREATE nodes alone.

## Solution

Introduce a `TraceArenas<T>` newtype that can only grow through `push`, which strips the steps of the arena it displaces. At most one step-laden arena is then alive per collection, and the invariant lives in the type rather than in a convention on a `Vec`. Mutable access is limited to methods that cannot reintroduce steps: `identify_and_decode` for the decoding pass, and the stripping methods. This bounds the transient peak within a test, not just the peak between tests. `ExecutionTraces` and `TestSetup::traces` (`SetupTraces`) are both aliases of it, since `get_stack_trace` and `get_setup_stack_trace` each only name the last arena of their collection. In the invariant replay, a call that turns out not to be the failing one is stripped as soon as that is known.

The replay also stops handing each call's arena to `BaseCounterExample::traces`, which nothing reads (#1702 established that). The arena moves into `execution_traces` instead of being cloned into it, and a dead clone of the `afterInvariant()` arena is dropped. The counterexample's copies were the step-laden originals and would have outlived the strip.

A preparatory commit renames the single-arena `traces` fields (`RawCallResult`, `InspectorData`, `CaseOutcome`, `FailedInvariantCaseData`, `FuzzTestResult`) to `call_trace_arena`, so call sites no longer read as if they carried several arenas.

## Validation

- Unit tests pin the newtype's behavior: pushing strips the displaced arena only, and `strip_steps` on a sparse arena matches resolve-then-drop.
- Measured on top of #1703 for an earlier revision of this change (peak RSS, one run per configuration): solady `-vvv` 5.0 → 4.8 GiB, `-vvvv` 4.4 → 4.3 GiB. The benchmark repos have no failing invariants and no `beforeTestSetup`, so the strip is a no-op there; the deltas are measurement spread.

## Notes for reviewer

- Stacked on #1703; part of the Solidity test runner memory-reduction series. Two commits: the field rename, then the newtypes.

### DIFF
--- a/crates/foundry/evm/evm/src/executors/invariant/replay.rs
+++ b/crates/foundry/evm/evm/src/executors/invariant/replay.rs
@@ -22,7 +22,7 @@ use foundry_evm_fuzz::{
     invariant::{BasicTxDetails, InvariantContract},
     BaseCounterExample,
 };
-use foundry_evm_traces::{load_contracts, SetupTraces, SparsedTraceArena, TracingMode};
+use foundry_evm_traces::{load_contracts, ExecutionTraces, SetupTraces, TracingMode};
 use parking_lot::RwLock;
 use proptest::test_runner::TestError;
 use revm::{
@@ -37,7 +37,7 @@ use super::{
 };
 use crate::executors::{
     stack_trace::{SolidityTestStackTraceError, SolidityTestStackTraceResult},
-    Executor,
+    Executor, RawCallResult,
 };
 
 /// Arguments to `replay_run`.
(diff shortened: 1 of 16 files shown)


crates/foundry/evm/evm/src/executors/invariant/replay.rs
```diff
@@ -140,69 +140,91 @@ pub fn replay_run<
 
     // Replay each call from the sequence, collect logs, traces and coverage.
     for tx in inputs.iter() {
-        let call_result = executor.transact_raw(
+        let RawCallResult {
```
Wodann, line 151: Thought: destructure to ensure compiler errors on addition of new fields

popescuoctavian (approved):
LGTM, thanks!

PR #1705, opened 2026-08-31
perf: don't collect invariant run traces past the gas-report budget

Wodann:
## Problem / context

`InvariantTestRun::run_traces` is only consumed by `InvariantTest::end_run`, which samples it into `gas_report_traces` while that vector holds fewer than `gas_report_samples` entries and drops it otherwise. When `gas_report_samples` is below `runs`, an invariant run accumulated up to `depth` arenas per run only to throw them away.

## Solution

Apply the sampling condition at collection time. The vector only grows in `end_run`, so the condition is constant for the duration of a run and exactly the same samples are kept. Both sites now share `InvariantTest::needs_gas_report_samples`.

The napi layer pins `gas_report_samples` to 0 whenever an `invariant` config is passed, which is what Hardhat does. So that every consumer gets the same benefit, `MultiContractRunner::new` now zeroes `gas_report_samples` when no gas report was requested and leaves it alone when one was.

## Validation

- Measured on top of #1704 (peak RSS, one run per configuration): morpho-blue, the invariant-heavy repo, 1.5 → 0.41 GiB at `-vvv` (−73%) and 1.6 → 0.44 GiB at `-vvvv` (−72%), and ~15% faster there (10.5 → 8.8 s at `-vvv`). The other repos and `-vv` are within noise.

## Notes for reviewer

- Stacked on #1704; part of the Solidity test runner memory-reduction series.
- With the Foundry default (256 samples, 256 runs) every run is still sampled, so Foundry-parity behavior is unchanged; the gate only bites when `gas_report_samples` < `runs`.


crates/edr_solidity_tests/src/multi_runner.rs
```diff
@@ -232,6 +232,9 @@ impl<
             include_traces = IncludeTraces::All;
             // Enable EVM isolation for more accurate gas measurements
             evm_opts.isolate = true;
+        } else {
+            // Nothing consumes gas-report samples, so don't collect them.
```
popescuoctavian, line 236: Suggestion: Should we also set `fuzz.gas_report_samples = 0` here? It currently has limited impact since `FuzzConfig::default()` sets it to 0, so it will only impact callers who explicitly set this. But it keeps the logic consistent across both fuzz and invariant tests.

popescuoctavian (approved):
LGTM, thanks! Just a small suggestion that you may consider integrating.

crates/edr_solidity_tests/src/multi_runner.rs
Wodann: Good catch! Addressed by e969e6215

PR #1715, opened 2026-08-31
feat(solidity-tests): scope inline config directives by test profile

schaable:
Adds test profile support to inline configuration, so Hardhat can offer profiles for Solidity tests.

`SolidityTestRunnerConfigArgs` gains two optional fields, `testProfile` and `declaredTestProfiles`, both defaulting to `default`. A caller that passes neither is unaffected.

A directive's key can now carry a profile prefix:

```solidity
/// forge-config: fuzz.runs = 3       // applies under every profile
/// forge-config: ci.fuzz.runs = 500  // applies only under `ci`
```

Where both set the same key, the prefixed one wins. Prefixes are validated against the *declared* profiles rather than the selected one, so a typo fails whichever profile is running.

### Behaviour changes

- A `default.`-prefixed directive is now scoped to the `default` profile instead of applying unconditionally.
- An unprefixed directive alongside a prefixed one setting the same key is no longer a duplicate-key error.
- `InlineConfigUnsupportedProfile` is renamed to `InlineConfigUndeclaredProfile` and now carries `declaredProfiles`.

The [design doc](https://app.notion.com/p/nomicfoundation/Solidity-test-profiles-3bc578cdeaf58000ba6bffbdc1ca735e) has the full rationale, including why prefixes validate against the declared set and why directives for unselected profiles are still validated.

Hardhat PR: https://github.com/NomicFoundation/hardhat/pull/8590

### DIFF
--- a/crates/edr_napi/src/solidity_tests/config.rs
+++ b/crates/edr_napi/src/solidity_tests/config.rs
@@ -5,6 +5,7 @@ use edr_primitives::hex;
 use edr_solidity_tests::{
     executors::invariant::InvariantConfig,
     fuzz::FuzzConfig,
+    inline_config::DEFAULT_PROFILE,
     inspectors::cheatcodes::{CheatsConfigOptions, ExecutionContextConfig},
     TestFilterConfig,
 };
@@ -223,6 +224,19 @@ pub struct SolidityTestRunnerConfigArgs<'env> {
     /// file and need no entry here; only non-relative paths (package imports)
     /// do.
     pub import_mappings: Option<HashMap<String, String>>,
+    /// The Solidity test profile this run was started with.
+    ///
+    /// An inline-config directive prefixed with a profile name
+    /// (`forge-config: ci.fuzz.runs = 8`) applies only under that profile; an
+    /// unprefixed one applies under every profile, and the prefixed one wins
+    /// where both set the same key. Defaults to `default`.
+    pub test_profile: Option<String>,
+    /// Every Solidity test profile the project declares.
+    ///
+    /// A prefix naming a profile that is not declared is an error, so a
+    /// mistyped one fails whichever profile is selected. `default` is always
+    /// declared. Defaults to `["default"]`.
+    pub declared_test_profiles: Option<Vec<String>>,
 }
 
 impl SolidityTestRunnerConfigArgs<'_> {
@@ -274,6 +288,8 @@ impl SolidityTestRunnerConfigArgs<'_> {
             eip712_canonical_types,
             test_source_paths,
             import_mappings,
+            test_profile,
+            declared_test_profiles,
         } = self;
 
         let test_source_paths = test_source_paths.map_or_else(HashMap::new, |paths| {
@@ -289,6 +305,10 @@ impl SolidityTestRunnerConfigArgs<'_> {
                 .map(|(import_path, path)| (import_path, PathBuf::from(path)))
                 .collect()
         });
+
+        let test_profile = test_profile.unwrap_or_else(|| DEFAULT_PROFILE.to_owned());
+        let declared_test_profiles =
+            declared_test_profiles.unwrap_or_else(|| vec![DEFAULT_PROFILE.to_owned()]);
 
         let test_pattern = TestFilterConfig {
             test_pattern: test_pattern
@@ -417,6 +437,8 @@ impl SolidityTestRunnerConfigArgs<'_> {
             generate_gas_report,
             test_source_paths,
             import_mappings,
+            test_profile,
+            declared_test_profiles,
         };
 
         Ok(config)
--- a/crates/edr_solidity_tests/src/inline_config.rs
+++ b/crates/edr_solidity_tests/src/inline_config.rs
@@ -19,6 +19,18 @@
 //!
 //! Both the `forge-config:` and `hardhat-config:` prefixes are recognized.
 //!
+//! A directive may be scoped to a test profile by prefixing its key with the
+//! profile's name. A prefixed directive applies only under that profile, an
+//! unprefixed one under every profile, and where both set the same key the
+//! prefixed one wins, whatever order they appear in. A prefix must name a
+//! declared profile — see [`InlineConfigProfiles`].
+//!
+//! ```solidity
+//! /// forge-config: fuzz.runs = 3       // under every profile
+//! /// forge-config: ci.fuzz.runs = 500  // only under `ci`
+//! function testFoo(uint256 x) public { /* ... */ }
+//! ```
+//!
 //! The work flows through the submodules as a pipeline:
 //!
 //! ```text
@@ -28,12 +40,16 @@
 //!   - overrides  compose the above into a source's per-contract overrides
 //!   - provider   cache the overrides and serve them
 //! ```
+//!
+//! `profiles` carries the run's selected and declared profiles through that
+//! pipeline, so `directives` can scope each one.
 
 mod directives;
 mod error;
 mod natspec;
 mod overrides;
 mod parse;
+mod profiles;
 mod provider;
 mod resolver;
 
@@ -41,9 +57,10 @@ pub(crate) use self::directives::is_test_function;
 pub use self::{
     error::{
         InlineConfigCollectError, InlineConfigError, InlineConfigErrorItem, InlineConfigErrors,
-        InlineConfigProblem,
+        InlineConfigProblem, InlineConfigProfilesError,
     },
     overrides::{ContractInlineConfig, FunctionOverride},
+    profiles::{InlineConfigProfiles, DEFAULT_PROFILE},
     provider::{CachedInlineConfigProvider, InlineConfigRoot, SharedInlineConfigProvider},
     resolver::ImportResolver,
 };
--- a/crates/edr_solidity_tests/src/inline_config/directives.rs
+++ b/crates/edr_solidity_tests/src/inline_config/directives.rs
@@ -8,8 +8,11 @@
 //! forge-config: default.fuzz.runs = 100
 //! hardhat-config: invariant.fail-on-revert = true
 //! ```
+//!
+//! A directive's key may carry a profile prefix, scoping it to that profile.
+//! See [`parse_inline_config`] for how scoping and precedence work.
 
-use super::{error::InlineConfigError, natspec::NatSpecBlock};
+use super::{error::InlineConfigError, natspec::NatSpecBlock, profiles::InlineConfigProfiles};
 use crate::config::{TestFunctionConfigOverride, TimeoutConfig};
 
 const HARDHAT_CONFIG_PREFIX: &str = "hardhat-config:";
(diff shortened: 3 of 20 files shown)


popescuoctavian:
Thanks for adding this @schaable! This conflicts with the contract-level inline config changes from https://github.com/NomicFoundation/edr/pull/1627. I'll wait for that to merge and then refactor this, but it may take a while. Sorry for that.

crates/edr_solidity_tests/src/inline_config/profiles.rs
```diff
@@ -0,0 +1,143 @@
...
+    pub fn new(
+        selected: impl Into<String>,
+        declared: impl IntoIterator<Item = String>,
+    ) -> Result<Self, InlineConfigProfilesError> {
+        let selected = selected.into();
+
+        // A `BTreeSet` deduplicates and keeps the names sorted, so error
+        // messages list them deterministically.
+        let mut declared: BTreeSet<String> = declared.into_iter().collect();
+        declared.insert(DEFAULT_PROFILE.to_owned());
```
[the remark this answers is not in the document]
popescuoctavian: Addresssed by 8f3fcd216af5e0785fdc1fd1a65e3b965524b9b2.

popescuoctavian (approved):

crates/edr_napi/src/solidity_tests/config.rs
```diff
@@ -223,6 +224,19 @@ pub struct SolidityTestRunnerConfigArgs<'env> {
...
+    /// (`forge-config: ci.fuzz.runs = 8`) applies only under that profile; an
+    /// unprefixed one applies under every profile, and the prefixed one wins
+    /// where both set the same key. Defaults to `default`.
+    pub test_profile: Option<String>,
+    /// Every Solidity test profile the project declares.
+    ///
+    /// A prefix naming a profile that is not declared is an error, so a
+    /// mistyped one fails whichever profile is selected. `default` is always
+    /// declared. Defaults to `["default"]`.
+    pub declared_test_profiles: Option<Vec<String>>,
```
Wodann: issue: "`default` is always declared." implies to me that `Some(vec!["foo"])` should resolve to `vec!["default", "foo"]`. However, that doesn't seem to be the case in the code.

We should either:
- update the doc
- fix the code

crates/edr_solidity_tests/src/inline_config/profiles.rs
```diff
@@ -0,0 +1,224 @@
...
+    /// Fails if a declared name cannot be written as a directive prefix, or if
+    /// the selected profile is not among the declared ones.
+    pub fn new(
+        selected: impl Into<String>,
+        declared: impl IntoIterator<Item = String>,
+    ) -> Result<Self, InlineConfigProfilesError> {
+        let selected = selected.into();
+
+        // A `BTreeSet` deduplicates and keeps the names sorted, so error
+        // messages list them deterministically.
```
Wodann, line 58: nit: I'd move this comment to the `struct InlineConfigProfiles` definition, where someone would edit the type/field.

crates/edr_solidity_tests/src/inline_config/profiles.rs
```diff
@@ -0,0 +1,224 @@
...
+        &self.selected
+    }
+
+    /// Whether `profile` is one of the project's declared profiles.
+    pub(super) fn is_declared(&self, profile: &str) -> bool {
+        self.declared.contains(profile)
+    }
+
+    /// Every declared profile, sorted, for error messages.
+    pub(super) fn declared_names(&self) -> Vec<String> {
```
Wodann, line 87: nit: I'd expose it as a `&[str]` to make it more reusable. Any caller of the API can then choose to convert its ownership model using `.iter().cloned().collect()`

it makes it usable for both use cases.

crates/edr_solidity_tests/src/inline_config/error.rs
```diff
@@ -116,6 +121,50 @@ pub enum InlineConfigError {
...
+    )]
+    UnrepresentableName {
+        /// The offending name.
+        name: String,
+    },
+    /// A declared profile name starts with `-` or `_`.
+    #[error(
+        "invalid Solidity test profile name `{name}`: a profile name must not start with `-` or `_`"
+    )]
+    LeadingSeparator {
```
Wodann, line 154: Question: Is there a reason that this is a separate error variant compared to `UnrepresentableName`? Can they be combined such that all illegal symbols are in one error type?

crates/edr_solidity_tests/src/inline_config.rs
```diff
@@ -19,6 +19,24 @@
...
+//!
+//! ```solidity
+//! /// forge-config: fuzz.runs = 3       // under every profile
+//! /// forge-config: ci.fuzz.runs = 500  // only under `ci`
+//! function testFoo(uint256 x) public { /* ... */ }
+//! ```
+//!
+//! Profiles are resolved separately for the contracts and functions,
+//! and only then are the contract's directives applied underneath the
+//! function's. Level precedence therefore outranks profile precedence: a
```
Wodann, line 36: typo/style: This reads oddly. I'm not 100% sure what is meant:

> and only then are the contract's directives applied underneath the

crates/edr_solidity_tests/tests/it/fuzz.rs
```diff
@@ -460,3 +461,62 @@ async fn test_fuzz_function_overrides() {
...
+        ),
+        (
+            "ci",
+            [
+                ("testFuzz_Unprefixed", 3),
+                ("testFuzz_ProfileWinsOverUnprefixed", 8),
+                ("testFuzz_DefaultProfileOnly", 100),
+            ],
+        ),
+    ];
```
Wodann, line 489: Question: Does this mean that `"ci"` does not inherit from `"default"` in EDR?

crates/edr_solidity_tests/tests/it/fuzz.rs
```diff
@@ -460,3 +461,62 @@ async fn test_fuzz_function_overrides() {
...
+
+    for (selected, cases) in expected_runs {
+        let filter = SolidityTestFilter::new(".*", ".*", ".*fuzz/FuzzProfileOverride.t.sol");
+        let mut config = TEST_DATA_DEFAULT.config_with_mock_rpc();
+        config.fuzz.runs = 100;
```
Wodann, line 494: nit: use a `const` for `GLOBAL_RUNS` like in other tests.

crates/edr_solidity_tests/src/inline_config/directives.rs
```diff
@@ -142,6 +167,33 @@ impl Key {
...
+    /// accepts both sections; each affects only the tests of its kind.
+    fn check_target(
+        self,
+        target: DirectiveTarget<'_>,
+        raw_key: &str,
+    ) -> Result<(), InlineConfigError> {
+        let DirectiveTarget::Function(function) = target else {
+            return Ok(());
+        };
+        let is_fuzz_test = function.starts_with("test");
```
Wodann, line 181: question: Is `"test"` the identifier for a fuzz test?

crates/edr_solidity_tests/src/inline_config/directives.rs
```diff
@@ -557,9 +671,239 @@ mod tests {
...
+        // Under another profile the directive is inert, and since it is the
+        // function's only one, nothing is collected for it at all.
+        assert_eq!(
+            parse_with(&profiles("default", &["ci"]), &directive, "testFoo"),
+            Ok(None)
+        );
+    }
+
+    #[test]
+    fn default_prefix_does_not_apply_under_another_profile() {
```
Wodann, line 728: suggestion: This should be clearly documented in the docs (both on the Hardhat 3 website and in EDR itself; e.g. in the inline config module-level docs).

Wodann (commented):
The main concern I have is around documenting the product decisions. We should do so both at the top-level inline_config.rs's module-level doc comments and on the Hardhat 3 website. Even more so, since we deviate from Foundry.

Ideally, this is written by a human rather than Claude and remains concise, merely stating the rules.

E.g.

```
/// Directives are handled in this order of priority:
///  1. Function-level > Contract-level
///  2. Profile > bare
/// 
/// Custom profiles do not inherit from the "default" profile.
```

crates/edr_solidity_tests/tests/it/fuzz.rs
Wodann: Self-answer: Yes. Custom profiles do not inherit from `"default"` in EDR/Hardhat.

crates/edr_solidity_tests/src/inline_config/profiles.rs
popescuoctavian: This is not possible since the underlying data structure is a `BTreeSet<String>`, which has no contiguous backing storage (i.e. elements live in tree nodes). Returning `&[String]` here would be possible if we switch to a sorted `Vec<String>` instead of a `BTreeSet`, but I think that complicates things unnecessarily (uses binary search). Leaving it as is for now, let me know if you have a better solution in mind.

crates/edr_solidity_tests/src/inline_config/profiles.rs
popescuoctavian: Addressed by a7ece87ed1bc4d2e139f253bcc36047d4486b3e7.

crates/edr_solidity_tests/src/inline_config/error.rs
popescuoctavian: No specific reason, mostly to separate the reasons. But I agree with using a single error, addressed by a9ef8c317b39b5b7509678b2de4a2a046f768b02.

crates/edr_solidity_tests/src/inline_config.rs
popescuoctavian: Addressed by bcd0bdde18f3bab61d9d735a9019219669d6bb50.

crates/edr_solidity_tests/tests/it/fuzz.rs
popescuoctavian: That is correct, added to docs in bcd0bdde18f3bab61d9d735a9019219669d6bb50.

crates/edr_solidity_tests/tests/it/fuzz.rs
popescuoctavian: Addressed by 8ac55d08d9da64d704e5bd64a96c51efe4966369.

crates/edr_solidity_tests/src/inline_config/directives.rs
popescuoctavian: Added to docs here bcd0bdde18f3bab61d9d735a9019219669d6bb50.

crates/edr_solidity_tests/src/inline_config/directives.rs
popescuoctavian: Partially - fuzz tests have start with `"test"` (same as unit tests) AND have at least one parameter. The current implementation accepts both and doesn't report failures when using `fuzz` configs for unit tests, they are simply ignored (same as Foundry).

However, I believe that we should be more restrictive. IIRC in the initial implementation I made, data about parameters was not immediately available, which is why I made this choice. But it is now trivial to obtain from the Slang ast and reject `fuzz` configs in unit tests. I'll address this in a follow-up PR since it's not related to this change. Great catch, thanks!

crates/edr_solidity_tests/src/inline_config/profiles.rs
Wodann: The only alternative then is to return an `impl Iterator<String>`, if we want to keep it extensible, I think.

popescuoctavian: Moved to `impl Iterator<Item = &str>` in 3a489b451e0ca3be09b89ebe97d60c10475773d7.

Wodann (approved):

crates/edr_solidity_tests/src/inline_config/directives.rs
Wodann: @schaable: Just so we are all on the same page. If and when you add website documentation about Solidity test profiles, something like [this](https://github.com/NomicFoundation/edr/commit/bcd0bdde18f3bab61d9d735a9019219669d6bb50) should be added there too.

schaable: Agreed. I've already added [website docs](https://github.com/NomicFoundation/hardhat-website/pull/297) for the feature. I'll review them to make sure this is explicitly covered.

PR #1719, opened 2026-09-01
fix(edr_solidity): accept 64-bit optimizer.runs in compiler input

nebasuke:
Change `runs` in the `solc` `OptimizerSettings` to support `U64`, to support Aave's value, and potentially other projects. 

# Claude summary
## Why

`edr_solidity`'s `OptimizerSettings.runs` was `Option<u32>`, but real projects ship vanity runs values wider than that: aave-v4's config historically used `optimizer.runs: 444444444444` (4.4×10¹¹; the current upstream reduced it to `44444444`, apparently to accommodate tooling with exactly this kind of limit). solc accepts the value — it treats runs as 64-bit, as does `foundry-compilers` (`Optimizer { runs: Option<usize> }` in our own dependency tree) — so build-info containing it is legitimate compiler input.

The impact was disproportionate: the serde error propagates out of `BuildInfoBuffers::parse`, failing the **entire build-info parse** — and with it stack traces — on both the solc and slang-solx paths, over a field the contract decoder never reads.

Found while building a DWARF decode benchmark corpus from aave-v4's real hardhat-slang-solx build-info (#1718): the bench setup hit `invalid value: integer '444444444444', expected u32`.

## What

- `runs: Option<u32>` → `Option<u64>`, matching solc and foundry-compilers. The field is deserialized but never read, so nothing downstream changes.
- Regression test deserializes a minimal `CompilerInput` with `runs: 444444444444` and round-trips it through serialization, so a silent clamp would also fail it. Proven red on the pre-fix code.
- Changeset: patch.

### DIFF
--- a/.changeset/optimizer-runs-u64.md
+++ b/.changeset/optimizer-runs-u64.md
@@ -1,0 +1,5 @@
+---
+"@nomicfoundation/edr": patch
+---
+
+Fixed build-info parsing rejecting compiler input whose `optimizer.runs` exceeds 32 bits. Projects using vanity runs values such as `444444444444` (aave-v4's historical setting) failed the whole build-info parse — disabling stack traces — even though solc accepts the value; the field is now read as 64-bit, matching solc and foundry-compilers.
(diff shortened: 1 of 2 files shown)


Wodann (approved):
Thanks for the fix!

PR #1720, opened 2026-09-01
fix: reclaim a provider's OS thread when it becomes unreachable

Wodann:
## Problem

#1486 gave every provider a dedicated OS thread, joined only when the JS `Provider` wrapper is garbage-collected and its finalizer runs. The wrapper is a handful of bytes on the JS heap, so V8 sees no reason to collect it, and the threads accumulate for the life of the process.

macOS caps threads per process far below Linux, so it hits the wall first. In [this Hardhat test job](https://github.com/NomicFoundation/edr/actions/runs/33519467687/job/99929773105?pr=1670) the suite created 1,998 providers before `pthread_create` started returning `EAGAIN`:

```
thread 'tokio-rt-worker' panicked at crates/edr_provider/src/provider.rs:151:14:
failed to spawn the provider thread: Os { code: 35, kind: WouldBlock,
                                          message: "Resource temporarily unavailable" }
```

A second defect turned that into a job timeout. `Provider::new` used `.expect` on the spawn, and it runs inside the `spawn_blocking` closure that owns `create_provider`'s deferred. napi-rs has no `Drop` for `JsDeferred`, so unwinding left the JavaScript promise pending forever. Each affected stack-trace test then sat out its full 120s timeout — the panics in the log are exactly 120s apart — and the job was cancelled after 16 of them.

A third defect was doing most of the damage, and only became visible once the hang was gone. Hardhat 2 handed EDR a `subscriptionCallback` that reached the wrapper through an event adapter, and the wrapper holds the provider. EDR keeps that callback alive through a threadsafe function — a strong reference V8 cannot see through — so the closure was a GC root and the provider stayed **reachable**, not merely un-pressured. No amount of external-memory reporting collects a reachable object, which is why the GC change alone did not move the wall: 2,006 providers instead of 1,998.

## Fix

**`fix(edr_provider): reject rather than hang when a thread is refused`** — adds `CreationError::ThreadSpawn` and propagates the `io::Error`. `CreationError` already reaches JS as a rejection, so `createProvider` now fails with a real message in milliseconds instead of hanging.

**`fix(edr_napi): let V8 collect providers that pin an OS thread`** — reports the thread's 2 MiB stack reservation as external memory, which is the pressure signal V8 was missing. A `GcTracked<T>` newtype reports it while converting the value for JavaScript, so only a value that became a JS object — and therefore has a finalizer to release it again — is ever counted. Reporting from a constructor cannot make that guarantee: a later failure would leave the finalizer subtracting an amount that was never added.

**`refactor(edr_napi): declare the report and its release together`** — the report and its release still had to be written by hand in two places, which is the asymmetry Copilot flagged in review. A `gc_tracked!` macro now emits the alias, the amount and the finalizer from one declaration:

```rust
gc_tracked! {
    /// A [`Provider`] on its way to JavaScript, reporting its OS thread to V8.
    pub(crate) type GcProvider = GcTracked<Provider>;

    /// The standard library's default thread stack reservation, which
    /// `CancellableThread` does not override.
    const EXTERNAL_MEMORY: i64 = 2 * 1024 * 1024;

    fn drop(self) {
        // whatever `Drop` would have done
    }
}
```

The amount lives on a `HasExternalMemory` trait rather than a const generic, so no second alias can name a figure the finalizer does not release. Two diagnostics come along: a marker blanket-implemented for `T: Drop` turns a hand-written `Drop` into an `E0119` naming where that body belongs, and `on_unimplemented` turns an undeclared type into ``  `X` is not declared with `gc_tracked!` `` rather than an unsatisfied trait bound.

**`fix(hardhat-tests): break Hardhat 2's provider reference cycle`** — patches a `WeakRef` into Hardhat 2 so the subscription callback no longer roots the wrapper, which also removes the event adapter that existed only to work around the construction ordering. Hardhat 3 already does exactly this in `edr-provider.ts`, so this follows its precedent rather than inventing one. The same change is needed on Hardhat's `v2` branch — the patch only covers the vendored suite until that lands. ([HH2 branch](https://github.com/Wodann/hardhat/tree/fix/edr-provider-reference-cycle))

**`test(hardhat-tests): widen the interval-mining timing margin`** that caused [these CI failures](https://github.com/NomicFoundation/edr/actions/runs/33624762704/job/100229797128?pr=1720) — the `evm_setIntervalMining` "using sleep" tests configured a 100 ms interval and slept 170 ms, leaving 70 ms between a block being due and the assertion. macOS missed it in 5 of 6 cases. The bound is two-sided — the sleep has to fall between one and two intervals — so the absolute margin grows only with the interval, not the multiplier; at 250 ms the block has 175 ms of slack, for about 2 s of extra suite time.

## Thread counts

Creating providers back to back on Linux with a debug build, sampling `/proc/self/status`. Through **patched Hardhat 2 against the locally built EDR** — the configuration that fails in CI — over 300 providers:

| | live threads after 300 providers | peak |
|---|---|---|
| before | **333**, one per provider, never reclaimed | 333 |
| after | 34 | 68 |

And with EDR driven directly, isolating the GC change from the cycle, over 400 providers:

| | live threads after 400 providers | peak |
|---|---|---|
| before | **433**, one per provider, never reclaimed | 433 |
| after | 39–67 | 79–83 |

Before each fix, growth is strictly linear and nothing is ever released. After it, the count is flat and settles back down. Three runs each; the peak varied by 4 threads.

Note that the second table only exercises the collectible case: its subscription callback captures nothing, so it never forms the cycle. Give that harness a Hardhat-shaped callback and it returns to 433 threads with the GC change in place, which is what the first table's fix addresses.

Reporting more than 2 MiB was tested and rejected: 8 MiB halves the peak but nearly triples GC count, and 32 MiB is no better than 8 MiB on threads while tripling it again. The returns are sublinear, the GC cost grows with the real heap, and 2 MiB is the only figure that is actually true.

## CI outcome

On the macOS job, in order: cancelled at the 30-minute mark → 957 failing → 5 failing (2,800 passing, 167 pending) → expected green. The last five were the timing-sensitive interval-mining tests, unrelated to the leak; the job was green before #1486 and every run since was cancelled, so they had never been visible.

## Validation

- Both harnesses above, run against builds with and without each change, and again after every refactor of the reporting mechanism.
- Subscription delivery after the `WeakRef` change: `eth_subscribe` plus `evm_mine` through the built Hardhat 2 package still delivers the notification. Hardhat 2 has no provider tests of its own, so that coverage lives here.
- Both diagnostics exercised by hand: adding a `Drop` impl to `Provider`, and wrapping a type the macro never declared.

### DIFF
--- a/.changeset/provider-thread-gc-pressure.md
+++ b/.changeset/provider-thread-gc-pressure.md
@@ -1,0 +1,7 @@
+---
+"@nomicfoundation/edr": patch
+---
+
+Fixed provider threads accumulating until the process ran out of them. Each provider owns a dedicated OS thread, which V8 could not see, so an unreachable provider was never collected and its thread never joined. On macOS, whose per-process thread limit is far below Linux's, a suite creating hundreds of providers exhausted it.
+
+`createProvider` now rejects when the OS refuses a thread, instead of panicking and leaving its promise pending forever.
--- a/crates/edr_napi/src/context.rs
+++ b/crates/edr_napi/src/context.rs
@@ -21,7 +21,7 @@ use crate::{
     config::{resolve_configs, ConfigResolution, ProviderConfig, TracingConfigWithBuffers},
     contract_decoder::ContractDecoder,
     logger::LoggerConfig,
-    provider::{factory::SyncProviderFactory, Provider, ProviderFactory},
+    provider::{factory::SyncProviderFactory, GcProvider, Provider, ProviderFactory},
     solidity_tests::{
         artifact::{Artifact, ArtifactId},
         config::SolidityTestRunnerConfigArgs,
@@ -139,14 +139,14 @@ impl EdrContext {
                     Arc::clone(&contract_decoder),
                 )
                 .map(|provider| {
-                    Provider::new(
+                    GcProvider::from(Provider::new(
                         provider,
                         runtime,
                         contract_decoder,
                         dropped_provider_sender,
                         #[cfg(feature = "scenarios")]
                         scenario_file,
-                    )
+                    ))
                 });
 
             deferred.resolve(|_env| result);
@@ -390,11 +390,13 @@ impl EdrContext {
 impl EdrContext {
     /// Creates a mock provider, which always returns the given response.
     /// For testing purposes.
-    #[napi(async_runtime)]
+    // `GcProvider` only carries the value to JavaScript, which receives a
+    // `Provider`.
+    #[napi(async_runtime, ts_return_type = "Provider")]
     pub fn create_mock_provider(
         &self,
         mocked_response: serde_json::Value,
-    ) -> napi::Result<Provider> {
+    ) -> napi::Result<GcProvider> {
         use crate::mock::MockProvider;
 
         let runtime = runtime::Handle::current();
@@ -404,16 +406,14 @@ impl EdrContext {
             context.provider_deallocator.sender()
         };
 
-        let provider = Provider::new(
+        Ok(GcProvider::from(Provider::new(
             Arc::new(MockProvider::new(mocked_response)),
             runtime,
             Arc::default(),
             dropped_provider_sender,
             #[cfg(feature = "scenarios")]
             None,
-        );
-
-        Ok(provider)
+        )))
     }
 
     /// Creates a provider with a mock timer.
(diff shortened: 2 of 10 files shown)


popescuoctavian (approved):
LGTM!

PR #1722, opened 2026-09-02
perf(edr_napi): report a response's payload to V8

Wodann:
A `Response` holds the JSON-RPC envelope and, when tracing is enabled, the call trace arenas. [PR #1301's analysis](https://github.com/NomicFoundation/edr/pull/1301#issuecomment-4915995062) measured peak RSS going from 445 MB to 1.5 GB with arenas retained, and attributed +1.0s of the slowdown to V8 working against the larger resident set.

None of that weight is visible to V8. The JS wrapper is a handful of bytes, so an unreachable response looks cheap and V8 has no reason to collect it — the same blind spot #1720 fixed for provider threads.

## What changed

`GcTracked`'s amount moves from an associated const to a `HasExternalMemory::external_memory` method, because a response's weight is not a property of its type. `Response` computes its figure once in `Response::new` and stores it. `Provider` keeps the call-trace figure in an `AtomicI64` that follows `verbose_raw_tracing`.

## Design constraints

**The rate is the binding constraint.** `Provider` is created a handful of times per process; `Response` is created once per JSON-RPC request. Measured while building #1720, 2 MiB per provider produced a collection roughly every 32 providers, so V8 behaves as though it wants a hint per ~64 MiB of reported growth. Applying that rate to responses:

| reported per response | collection hint every |
|---|---|
| 16 MiB | ~4 requests |
| 4 MiB | ~16 requests |
| 1 MiB | ~64 requests |
| ~2 KB | ~32,000 requests |

**The payload is bimodal, so no single constant works.** On the common Hardhat path `data` is ~2 KB with no arenas; with tracing on the arenas are tens of MB. Sized for traces a constant wrecks the common path; sized for the common path it does nothing for traces.

**The report and its release must not drift.** An associated const guaranteed that by construction. A method only guarantees it if the value's answer never changes — so the figure is *stored* rather than recomputed, and both ends read one immutable field. That is the whole reason `Response` carries it as a field.

**It has to be O(1), with no arena traversal.** The figure comes from provider configuration, which is known before any request runs, so nothing walks a trace. The trait documents the constant-time expectation for future implementors.

**`IncludeTraces::Failing` is bimodal within one provider** — most responses carry no arenas, failures carry full ones. Keying on `call_trace_arenas.is_empty()` rather than on the configured `IncludeTraces` makes `Failing` exact instead of a compromise, and `is_empty()` is O(1).

**The atomic is `Relaxed` because it publishes nothing.** Stronger orderings exist to make other writes visible alongside the atomic one, and there are none. A response created concurrently with `setVerboseTracing` may read either figure, and both are sound: whichever it reads is what it stores and later releases.

## The figures

Deliberately **uncalibrated**. Memory work is in flight, and a figure calibrated against today's allocations would overshoot once those PRs land. Being wrong is not symmetric — over-reporting costs a collection on every request, under-reporting only forgoes some of the benefit — so these lean low.

They are derived rather than guessed. Measured with `size_of` against `revm-inspectors` 0.41.4: `CallTraceStep` = 192 B, `CallTraceNode` = 384 B, `CallLog` = 104 B. The step size independently confirms #1301's "~200-byte `CallTraceStep`".

| case | figure | derivation |
|---|---|---|
| untraced | `data.len()` | exact for the envelope, and O(1) |
| stack trace present | +4 KiB | a handful of frames of strings |
| envelope too large to marshal as a string | 256 MiB | only reachable above the 250 MB limit, and a `Value` tree costs several times its serialized length, so a floor |
| call traces recorded | +1 MiB | 192 B/step × ~5,400 steps |
| …with stack snapshots | +4 MiB | ~448 B/step (192 flat plus a `Box<[U256]>`) × ~9,300 steps |

Two checks on the traced figures. The **4× ratio is corroborated from the other direction**: #1301 measured +1.05 GB of retained arenas with `StackSnapshotType::Top` against +5.15 GB with `Full`, a ratio of 4.9 — so 4× under-reports the verbose case against the only real data available, and a unit test pins the ratio so the two constants can't drift apart. And **1 MiB is a small trace**: a complex contract call runs tens of thousands of opcodes, so a seaport-sized transaction is plausibly 5–10× that.

The one aggressive cadence is traced-verbose at ~16 requests per hint. That would be indefensible on the common path, but verbose tracing already costs orders of magnitude more per request than a collection, and its failure mode without this change is the OOM #1301 hit at 5.6 GB.

## Validation

- Six unit tests on the figure computation: envelope length, the stack-trace term, the arena discriminator, the verbose figure, the `Value` floor, and the 4× ratio.

## Follow-up

#1562 off-loads response deallocation to a dedicated thread. Until it lands, reporting external memory makes finalizers run earlier and more often, so *more* arena teardown happens on the JS thread — the ~0.5s #1301 measured — in exchange for lower peak RSS. The common path is unaffected, since untraced responses report ~2 KB. Worth watching on the benchmark rather than blocking on.

### DIFF
--- a/crates/edr_napi/src/provider/response.rs
+++ b/crates/edr_napi/src/provider/response.rs
@@ -11,8 +11,7 @@ use crate::{
     },
 };
 
-/// Bytes reported for a response whose call trace arenas were recorded without
-/// stack snapshots.
+/// Bytes reported for one recorded call trace arena, without stack snapshots.
 ///
 /// A `CallTraceStep` is 192 bytes, so this is a trace of roughly 5,400 steps.
 /// A complex contract call runs many more, which is the intended direction:
@@ -20,8 +19,8 @@ use crate::{
 /// only forgoes some of the benefit.
 const CALL_TRACE_EXTERNAL_MEM_SIZE: i64 = 1024 * 1024;
 
-/// Bytes reported for a response whose call trace steps also carry stack
-/// snapshots, as `verbose_raw_tracing` records them.
+/// Bytes reported for one recorded call trace arena whose steps also carry
+/// stack snapshots, as `verbose_raw_tracing` records them.
 ///
 /// A snapshot is a `Box<[U256]>`, so a step costs roughly 448 bytes rather than
 /// 192. PR #1301 measured 4.9x more retained memory in this configuration, so
@@ -38,10 +37,10 @@ const STACK_TRACE_EXTERNAL_MEM_SIZE: i64 = 4 * 1024;
 /// tree costs several times its serialized length, so this is a floor.
 const VALUE_DATA_EXTERNAL_MEM_SIZE: i64 = 256 * 1024 * 1024;
 
-/// Bytes a response reports for the call trace arenas it carries.
+/// Bytes a response reports for each call trace arena it carries.
 ///
 /// Recording stack snapshots takes a step from roughly 192 bytes to 448, so a
-/// verbosely traced response stands for several times as much.
+/// verbosely recorded arena stands for several times as much.
 pub(crate) const fn call_trace_external_mem_size(verbose_tracing: bool) -> i64 {
     if verbose_tracing {
         VERBOSE_CALL_TRACE_EXTERNAL_MEM_SIZE
@@ -61,9 +60,9 @@ pub struct Response {
 impl Response {
     /// Constructs a new instance.
     ///
-    /// `call_trace_external_mem_size` is what the provider reports for a
-    /// response carrying call trace arenas, which depends on whether it records
-    /// stack snapshots.
+    /// `call_trace_external_mem_size` is what the provider reports for each
+    /// call trace arena the response carries, which depends on whether it
+    /// records stack snapshots.
     pub(crate) fn new(
         inner: edr_napi_core::spec::Response,
         call_trace_external_mem_size: i64,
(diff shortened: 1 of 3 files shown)


Wodann:
/bench hardhat-ref=bench/test-verbosity-levels benchmarks="test vitest*,test mocha*"

crates/edr_napi/src/provider/response.rs
```diff
@@ -3,21 +3,102 @@ use napi::{bindgen_prelude::Either3, Either};
...
+            Either::A(envelope) => i64::try_from(envelope.len()).unwrap_or(i64::MAX),
+            Either::B(_value) => VALUE_DATA_EXTERNAL_MEM_SIZE,
+        };
+
+        if inner.stack_trace_result.is_some() {
+            external_memory += STACK_TRACE_EXTERNAL_MEM_SIZE;
+        }
+
+        if !inner.call_trace_arenas.is_empty() {
+            external_memory += call_trace_external_mem_size;
```
popescuoctavian: Question: should this instead be `call_trace_external_mem_size * call_trace_arenas.len()` for a more accurate representation?

popescuoctavian (approved):
LGTM overall, thanks! Just a small question, up to you if it's worth addressing

crates/edr_napi/src/provider/response.rs
Wodann: I adopted your idea: https://github.com/NomicFoundation/edr/pull/1722/changes/eee884aeb9fead2f0659194476ec7eb0719d71f7. Thanks!

PR #1738, opened 2026-09-07
test: add EIP-8024 new stack instruction tests (Amsterdam)

anaPerezGhiglia:
- added integration test to assert the instructions introduced by EIP-8024 are properly gated by Amsterdam
- Bonus: Introduce new bytecode module to encapsulate hand-built bytecode. I'll refactor existing tests to use this instead in a follow-up PR

### DIFF
--- a/crates/edr_provider/tests/common/bytecode.rs
+++ b/crates/edr_provider/tests/common/bytecode.rs
@@ -53,7 +53,7 @@ impl BytecodeBuilder {
 
     /// Init bytecode wrapping the runtime in the standard constructor that
     /// copies it out as the deployed code.
-    pub fn deployable(&self) -> Bytes {
+    pub fn deployable(self) -> Bytes {
         const CONSTRUCTOR_LEN: u8 = 12;
 
         let runtime_len = u8::try_from(self.runtime.len()).expect("runtime length fits a PUSH1");
--- a/crates/edr_provider/tests/common/bytecode.rs
+++ b/crates/edr_provider/tests/common/bytecode.rs
@@ -1,0 +1,79 @@
+//! Tiny EVM assembler for hand-built test contracts.
+#![cfg(feature = "test-utils")]
+
+use edr_primitives::Bytes;
+
+/// Opcodes used by hand-built test contracts.
+pub mod opcode {
+    pub const CODECOPY: u8 = 0x39;
+    pub const POP: u8 = 0x50;
+    pub const MSTORE: u8 = 0x52;
+    pub const PUSH1: u8 = 0x60;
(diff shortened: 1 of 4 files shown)


crates/edr_provider/tests/common/mod.rs
```diff
@@ -1,6 +1,7 @@
 /// Test utilities for blob transactions.
```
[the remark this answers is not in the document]
anaPerezGhiglia: That's an outer `///` doc on `pub mod blob;` (not a `//!` module doc for common), so it documents only the blob utilities

crates/edr_provider/tests/common/bytecode.rs
```diff
@@ -0,0 +1,79 @@
...
+        self.push1(0)
+            .opcode(opcode::MSTORE)
+            .push1(0x20)
+            .push1(0)
+            .opcode(opcode::RETURN)
+    }
+
+    /// Init bytecode wrapping the runtime in the standard constructor that
+    /// copies it out as the deployed code.
+    pub fn deployable(&self) -> Bytes {
```
Wodann: Question: commonly, the builder pattern consumes `self` when it builds the target output, moving all memory out rather than cloning. In this case you are passing it by reference. Is there a reason you deviated from this pattern?

crates/edr_provider/tests/integration/eip8024.rs
```diff
@@ -0,0 +1,154 @@
...
+    },
+    provider::{new_provider, new_provider_with_config},
+};
+
+const SENDER: Address = address!("0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266");
+
+/// Marker value each contract plants on the stack and must return. The stack
+/// slots around it hold `0` or `7`, so a wrong stack manipulation returns a
+/// different value instead of failing outright.
+const MARKER: u8 = 42;
```
Wodann:
```suggestion
/// Marker value each contract plants on the stack and must return.
const MARKER: u8 = 42;
```

The other text doesn't directly speak about `MARKER` but rather invariants that are expressed elsewhere in the code.

Wodann (commented):
Nice reusable test utilities!

I have one question and one suggestion w.r.t verbose comments. Otherwise looks good to me, after (potentially) addressing those.

Wodann (approved):

crates/edr_provider/tests/common/bytecode.rs
anaPerezGhiglia: no particular reason, I wasn't thinking of this as a builder-culmination step, but I like your perspective.

PR #1739, opened 2026-09-07
refactor(tests): reuse the bytecode builder in pre-existing tests

anaPerezGhiglia:
#1738 introduced a bytecode test module to help build hand-written bytecode, so the intent of a constructed contract is readable from code. This PR refactors the pre-existing hand-assembled contracts to use it

### DIFF
--- a/crates/edr_provider/tests/common/bytecode.rs
+++ b/crates/edr_provider/tests/common/bytecode.rs
@@ -5,9 +5,13 @@ use edr_primitives::Bytes;
 
 /// Opcodes used by hand-built test contracts.
 pub mod opcode {
+    pub const STOP: u8 = 0x00;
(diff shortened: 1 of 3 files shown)


crates/edr_provider/tests/common/bytecode.rs
```diff
@@ -5,9 +5,13 @@ use edr_primitives::Bytes;
 
 /// Opcodes used by hand-built test contracts.
 pub mod opcode {
+    pub const STOP: u8 = 0x00;
```
Wodann: Thought: All of these opcodes are already defined in REVM. Is there a way you can just re-export and use those instead of redefining them all?

Wodann (approved):
One thought on how to reduce code duplication. Otherwise, LGTM! Feel free to merge without asking for re-approval.

PR #1698, opened 2026-08-31
perf: serialize JSON-RPC responses once

Wodann:
Follow-up to #1486, and stacked on it.

`ResponseWithCallTraces.result` was a `serde_json::Value`. Handlers built it with `serde_json::to_value`, then `marshal_response_data` walked the tree again with `serde_json::to_string`. Holding the serialized bytes instead makes the second pass a copy.

The performance gain is not significant but it is a prerequisite for further JSON-RPC round-trip optimisations.

## What changes

- `result` becomes a `Box<serde_json::value::RawValue>`, so each handler serializes once. Readers go through `RawValue::get` or the new `deserialize_result`.
- `marshal_response_data` sizes its buffer from the result's length rather than growing from 128 bytes. That same length also decides the string limit up front, so an oversized response no longer serializes a 250 MB string only to discard it.
- Batch responses are assembled from the already-serialized elements instead of being re-serialized through a `Value::Array`.
- serde_json's `raw_value` feature is enabled workspace-wide, alongside the features already declared there.

## Compatibility

The JSON reaching callers is byte-for-byte unchanged. A `Box<RawValue>` inside the untagged `jsonrpc::ResponseData` serializes exactly as the `Value` did, including through `#[serde(flatten)]`, and arbitrary-precision numbers survive. `marshal_response_data_matches_serde_for_success` and its error counterpart pin both arms against serde's own output.

No `#[napi]` signature changes, so `index.d.ts` is untouched.

## Benchmark Results

[JS scenario runner benchmark](https://github.com/NomicFoundation/edr/actions/runs/33388230122/attempts/2#summary-99621088615) (CI, vs latest main):

| Scenario | main | #1486 | #1698 | #1486 / main | #1698 / main | #1698 / #1486 |
|-|-:|-:|-:|-:|-:|-:|
| **All Scenarios** | **210,629** | **153,020** | **148,833** | **0.73** | **0.71** | **0.97** |
| `synthetix_9a3a109f` | 156,751 | 111,860 | 108,024 | 0.71 | 0.69 | 0.97 |
| `neptune-mutual-blue-protocol_8db6480` | 21,999 | 17,240 | 16,885 | 0.78 | 0.77 | 0.98 |
| `rocketpool_6a9dbfd8` | 12,998 | 8,940 | 8,838 | 0.69 | 0.68 | 0.99 |
| `openzeppelin-contracts_0a5fba7a` | 9,110 | 7,250 | 7,337 | 0.80 | 0.81 | 1.01 |
[...]

PR #1740, opened 2026-09-08
perf(edr_napi): lower the call trace arena external memory report

Wodann:
#1722 made each response report external memory to V8 for the call trace arenas it carries, at a flat 1 MiB per arena. V8 asks for a collection every 64 MiB reported, so that is one full collection per 64 arenas. However, recent optimisations in #1702, #1703, and #1704 reduced the actual memory size.

At `-vvvv` Hardhat requests traces for every call and transaction (`verbosityToIncludeTraces` → `IncludeTraces.All`), and gas estimation contributes one arena per gas probe. A full collection therefore incorrectly landed every few transactions, on a multi-gigabyte heap. The hh3 regression benchmark regressed accordingly, measured on the two push runs either side of the merge (334d190 → 8e4e1f1b):

| benchmark | metric | before | after | Δ |
| --- | --- | --- | --- | --- |
| openzeppelin-contracts / test mocha -vvvv | cpu | 70.1 s | 276.4 s | +294% |
| openzeppelin-contracts / test mocha -vvvv | wall | 64.5 s | 90.4 s | +40% |
| openzeppelin-contracts / test mocha -vvv | cpu | 64.2 s | 89.5 s | +39% |
| lidofinance-core / test mocha -vvvv | cpu | 387.9 s | 512.8 s | +32% |
| ens-contracts / test vitest -vvvv | wall | 4.65 s | 5.65 s | +21% |

The cost tracks verbosity exactly, i.e. the number of arenas per response: `test mocha` without `-v` is unchanged (+0.1%), and all 60 `test solidity` entries are flat within ±2%. The other commit in that measurement window, #1698, runs on every response regardless of verbosity, so it is not implicated.

## Fix

Report 32 KiB per arena instead of 1 MiB, and 128 KiB instead of 4 MiB when stack snapshots are recorded. The 4x ratio between the two is unchanged.

32 KiB is roughly 85 `CallTraceNode`s, and the figure now deliberately excludes the EVM steps an arena also records. Steps are the bulk of an arena, so leaving them out under-reports on purpose: V8 now asks for a collection every ~2048 arenas rather than every 64.

## What this gives up

Peak RSS also improved in those runs — 6% for openzeppelin-contracts, but 49% and 70% for lidofinance-core at `-vvv` and `-vvvv` (4042 MB → 1214 MB). Reporting 32x less memory means the garbage collector will be triggered less often and some of those gains might be reduced. That is the intended trade: runtime over memory.

## Validation
- [ ] A `/bench` run before merging, to confirm the CPU recovery and measure how much RSS comes back.

### DIFF
--- a/crates/edr_napi/src/provider/response.rs
+++ b/crates/edr_napi/src/provider/response.rs
@@ -13,19 +13,20 @@ use crate::{
 
 /// Bytes reported for one recorded call trace arena, without stack snapshots.
 ///
-/// A `CallTraceStep` is 192 bytes, so this is a trace of roughly 5,400 steps.
-/// A complex contract call runs many more, which is the intended direction:
-/// over-reporting costs a collection on every request, while under-reporting
-/// only forgoes some of the benefit.
-const CALL_TRACE_EXTERNAL_MEM_SIZE: i64 = 1024 * 1024;
+/// A `CallTraceNode` is 384 bytes, so this stands for roughly 85 calls. The
+/// EVM steps that an arena also records are deliberately not counted.
+///
+/// The figure under-reports on purpose. V8 asks for a collection every 64 MiB
+/// reported, and collecting a multi-gigabyte heap costs more than the arenas
+/// it releases.
+const CALL_TRACE_EXTERNAL_MEM_SIZE: i64 = 32 * 1024;
 
 /// Bytes reported for one recorded call trace arena whose steps also carry
 /// stack snapshots, as `verbose_raw_tracing` records them.
 ///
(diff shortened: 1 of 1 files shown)


Wodann:
Benchmarks show that the performance regression from https://github.com/NomicFoundation/edr/pull/1722 was reversed by this PR. However, two of the benchmarks show significantly increased peak RSS; which is to be expected when the GC is triggered less often.

This is acceptable based on the trade-off that we prefer runtime performance over peak memory consumption.

popescuoctavian (approved):

PR #1752, opened 2026-09-14
build: upgrade Hardhat v3 to 3.16.0 and trim its patch

nebasuke:
As prep for getting #1574 into shape, I need to bump Hardhat to a version that support the `hardhat-slang-solx` plugin. 

# Claude summary

Moves every Hardhat 3 consumer (`js/helpers`, `js/benchmark`, `js/integration-tests/solidity-tests`, `js/integration-tests/solx-parity-sweep`) from the patched 3.4.5 to 3.16.0. `@nomicfoundation/hardhat-slang-solx` declares `hardhat ^3.15.0` as a peer, so the solx parity sweep could not install it and has been self-skipping since it was written; this unblocks making it a real dependency (#1574).

The patch shrinks from three parts to one. The provider-config rewrite and the hardcoded `disableTransactionGasCap` are both upstream in 3.16.0, leaving only the three internal-module exports our own packages import. `hardhat-utils` / `hardhat-errors` also move off the `3.0.0-next.0` pre-releases to the ranges 3.16.0 uses, and the sweep's `hardhat` range becomes an exact pin like its three siblings.

## Surprises

**The dropped `rpcEndpoints` hunk was load-bearing**, and it was covering a type lie. `createSolidityTestsInput` passed `userConfig.solidityTest` to `solidityTestConfigToSolidityTestRunnerConfigArgs`, whose parameter is `Omit<SolidityTestProfileConfig, "eip712Types">` — the *resolved* profile, where `forking.rpcEndpoints` holds `ResolvedConfigurationVariable`s rather than plain strings. `userConfig` comes from a dynamic `import()` and is `any`, so tsc never saw it and the patch's `try`/`catch` quietly papered over the difference. It now reads `hre.config.test.solidity.profiles.default` off the HRE the function already builds.

**3.16 moved the user-config key from `solidityTest` to `test.solidity`**, so Hardhat had been silently ignoring all five benchmark repo configs — the benchmark only worked because it read the key itself. They now nest correctly. `forge-std` needs `process.env.ALCHEMY_URL ?? ""` as a result: the new key is validated, so an unset secret would fail config resolution outright instead of only the fork suites.

**Three keys in those configs were never real.** `fsPermissions.read` / `readWrite` (solady, uniswap-v4-core) are Foundry's `fs_permissions` access names, never mapped onto Hardhat's `readDirectory` / `dangerouslyReadWriteDirectory`. They are deleted rather than renamed: `createSolidityTestsInput` already grants `DangerouslyReadWriteDirectory` on the repo root, and since `find_permission` lets the longest path win, wiring the names up would *downgrade* `./test/data` and `./out` to read-only. `invariant.seed` (morpho-blue, solady) is not a field of `InvariantConfig` in ours or Foundry's — the invariant runner seeds from `fuzz_config.seed`, which both repos already set to the same value — so it comes out of the `foundry.toml` the forge leg reads too.

**hardhat-utils 4 requires an absolute mutex path.** `edr-helpers`' build mutex threw `InvalidMultiProcessMutexPathError` and cancelled every solidity-tests suite in its `before` hook; it now passes the `os.tmpdir()` location the old name-based constructor derived internally.

## Notes

- Renovate has `hardhat` and `@nomicfoundation/hardhat-*` disabled ("exact-pinned, pnpm-patched, coordinated manually") — this is that coordination.
- All four packages are private; no changeset.


popescuoctavian (approved):
LGTM, thanks!

PR #1753, opened 2026-09-15
perf(cheatcodes): cache private key to wallet derivation for `vm.addr` and `vm.sign`

### LINKED ISSUE #1658: vm.addr() re-derives the public key on every call - cache it
`foundry_cheatcodes::crypto::parse_wallet` performs a full k256 scalar multiplication (`lincomb`/`invert`) on every `vm.addr(privateKey)` call. Fuzz and `setUp` loops call it with a small set of repeated keys.

Measured: **~2% of wall** on aave-v4's Solidity test run; k256 point ops are ~2.5% of test-class CPU weighted.

## Direction

Memoise private-key -> (public key, address) with a small bounded cache. Only the derivation is cacheable — per-message signing is not. Trivial risk; addresses are deterministic, so any behavioural diff means the cache is wrong.

## Implementation plan

````markdown
# Cache `vm.addr()` private-key → address derivation in cheatcodes

## Problem

[Profiling Hardhat 3 e2e scenarios](https://app.notion.com/p/nomicfoundation/Runtime-Profiling-2026-08-05-3b3578cdeaf5808eafdff4e22ba425d0?source=copy_link): aave-v4's `test-solidity` run spends **~1.9%** of total CPU inside `foundry_cheatcodes::crypto::parse_wallet` → k256 EC math (`ProjectivePoint::add/double`, `FieldElement::{invert,square}`, `lincomb`), reached from `Vm::addrCall`. Across all 11 test runs, k256 point arithmetic is ~2.5% weighted. Test suites (especially fuzz campaigns and per-test `setUp`) call `vm.addr(key)` with a small set of repeated keys — each call redoes a full scalar multiplication.

Call-chain evidence: `Vm::addrCall` → `foundry_cheatcodes::crypto::parse_wallet` → `k256::arithmetic::mul::lincomb` / `FieldElement::invert`.

## Task

1. Read `crates/foundry/cheatcodes/src/crypto.rs` (`parse_wallet` and the `addr`/`sign`-family cheatcode implementations) and find every path that derives a public key/address from a private key per call (`vm.addr`, `vm.sign*` variants that re-derive, `createWallet`, …).
2. Add a small bounded memoization for private-key → (public key, address) — e.g. an LRU (or plain HashMap with size cap) keyed by the 32-byte private key. Decide placement: per-`Cheatcodes` instance state is simplest (recreated per test — still wins within fuzz runs and loops); a shared per-suite cache (must be `Send + Sync` — suites run on multiple threads; the `SharedEip712TypeProvider` pattern is prior art) captures cross-test reuse of the standard test keys. Start with the simplest placement that shows the win.
3. Security/behavior notes: keys are test-only material, but don't log them and keep the cache out of any Debug output; signing itself (per-message) must not be cached — only the key→pubkey/address derivation.
4. Verify: `cargo test -p foundry-cheatcodes`, `cargo clippy -p foundry-cheatcodes --all-targets` (the workspace clippy config is strict — watch `indexing_slicing` and friends), plus `cargo check -p edr_napi`.

## Verification (before/after)

Profile before and after with Hardhat's [`pnpm profiler`](https://github.com/NomicFoundation/hardhat/tree/main/scripts/profiler) script (a bare `pnpm profiler` prints its usage):

```bash
cd crates/edr_napi
pnpm build:perf-js
# in the hardhat repo, with the freshly built EDR injected:
NAPI_RS_NATIVE_LIBRARY_PATH=<edr>/crates/edr_napi/edr.linux-x64-gnu.node \
  pnpm profiler --scenario ./end-to-end/aave-v4 \
  --prepare "cold compile" --command "test solidity" --mode system
```

Expect: `parse_wallet`/k256 entries in the aave-v4 symbol profile reduced to noise; all tests passing with identical results (addresses are deterministic — any behavioral diff means the cache is wrong).
````

### PR DESCRIPTION
**Pull request by popescuoctavian**
## Summary

`vm.addr`, `vm.sign` and `vm.signCompact` derived the wallet from the private key on every call. That derivation is a full secp256k1 scalar multiplication and showed up as ~2% of CPU on the aave-v4 Solidity test run, because test suites call these cheatcodes repeatedly with a handful of keys.

This PR memoises the private key -> `PrivateKeySigner` derivation in a bounded, thread-safe `WalletCache` (LRU of 256 entries behind a `parking_lot::Mutex`). Only the derivation is cached; signing is still performed per message.

## Placement

The cache lives on `CheatsConfig`, which is created once per test suite and shared through an `Arc` by every test, fuzz run and stack-trace re-run of that suite. This deviates from upstream Foundry, which keeps an owned `HashMap` on the `Cheatcodes` inspector. That state is cloned per call and discarded after non-committing calls, so it only helps for keys derived in `setUp`; keys used inside a test body or fuzz run are re-derived on every run. The shared per-suite cache covers both.

## Notes

- Invalid keys are rejected before derivation and never cached, so error behaviour is unchanged.
- `WalletCache`'s `Debug` impl omits the entries so private keys never end up in logs or panic messages.
- `lru` was already a dependency of `edr_provider`; it is now a workspace dependency used by both crates.

## Testing on `aave-v4`

On local machine, for aave-v4 (`test solidity`, 1410 tests, fuzz runs 100), `perf` attributes ~1.45% of CPU samples to `vm.addr`/`vm.sign` key derivation before this change and ~0.08% after.

Closes #1658

### CODE REVIEW

File: crates/foundry/cheatcodes/src/config.rs [hunk 1]
```diff
@@ -64,6 +64,8 @@ pub struct CheatsConfig<HardforkT> {
     pub chain_id_to_alias: HashMap<u64, String>,
     /// Mapping of known EIP-712 canonical type definitions by type name.
     pub eip712_types_by_name: HashMap<String, Eip712TypeDef>,
+    /// Memoized private key -> wallet derivations.
+    pub wallet_cache: WalletCache,
```
popescuoctavian, line 68: Note for reviewer:

Storing the cache here is not ideal since it is a runtime state rather than a config. However, it provides the most straightforward solution to implement per-suite caching support (which covers both `setUp` and test derivation).

Other options for storing the cache were explored. The best alternative would be to store it in the `Cheatcodes` [inspector](https://github.com/NomicFoundation/edr/blob/c287dc84c4b2cf6de7bb0ea62cb86f67fae40294/crates/foundry/cheatcodes/src/inspector.rs#L411-L412). However, that does not cover stack trace generation re-runs, which would create a new cache for each run. Covering that case requires also storing the cache in the `InspectorStackBuilder`, which is not ideal.

crates/foundry/cheatcodes/src/crypto.rs
```diff
@@ -236,10 +244,84 @@ fn parse_private_key_p256(private_key: &U256) -> Result<P256SigningKey> {
...
+fn parse_wallet(private_key: &U256) -> Result<PrivateKeySigner> {
     parse_private_key(private_key).map(PrivateKeySigner::from)
 }
 
+/// Bounded, thread-safe memoization of private key -> wallet derivations.
+///
+/// Private keys are test-only material, but they are still never exposed via
+/// [`fmt::Debug`].
+pub struct WalletCache {
+    inner: Mutex<LruCache<U256, PrivateKeySigner>>,
```
Wodann, line 260: Thought: Would a `RwLock` be more efficient?

We can obtain a read-lock first to check presence. On failure, we'd need to get an `upgradable` lock (only one upgradable lock can be owned at time, hence why we first use the read lock), recheck (in case someone else wrote our entry in the meantime), and write to cache.

crates/foundry/cheatcodes/src/crypto.rs
```diff
@@ -236,10 +244,84 @@ fn parse_private_key_p256(private_key: &U256) -> Result<P256SigningKey> {
...
+    }
+}
+
+impl Clone for WalletCache {
+    fn clone(&self) -> Self {
+        Self {
+            inner: Mutex::new(self.inner.lock().clone()),
+        }
+    }
+}
```
Wodann: Question: Is `Clone` ever used?

crates/foundry/cheatcodes/src/crypto.rs
```diff
@@ -284,6 +366,103 @@ mod tests {
...
+        let cache = WalletCache::default();
+
+        // Standard test keys plus a "random-looking" one.
+        let keys: [U256; 4] = [
+            U256::from(1),
+            U256::from(0xBEEF),
+            U256::from(0xA11CE),
+            "0x4c0883a69102937d6231471b5dbb6204fe5129617082792ae468d01a3f362318"
+                .parse()
+                .unwrap(),
```
Wodann: nit: These tests use `unwrap` rather than `expect` with a helpful message.

Wodann (approved):
Just a couple of suggestions for minor improvements. Approved pending the benchmark run: https://github.com/NomicFoundation/edr/actions/runs/34977414771

I just want to validate that we similar gains in `aave-v4` as locally and no regressions in other scenarios.

popescuoctavian:
@Wodann I also ran a Claude-driven derivation-heavy local test to measure improvements:

### Workload

One test contract, `fuzz.runs = 100000`, fixed seed. `setUp` derives four well-known keys with `vm.addr`, then three fuzz tests:

| test | what it exercises |
| --- | --- |
| `testFuzz_AddrRepeatedKeys(uint8)` | `vm.addr` on one of 4 repeated keys, compared to the `setUp` result |
| `testFuzz_SignRepeatedKeys(uint8,bytes32)` | `vm.sign` on a repeated key, then `ecrecover` must match `vm.addr` |
| `testFuzz_AddrDistinctKeys(uint248,bytes32)` | `vm.sign` + `vm.addr` on a fresh key every run: an all-miss control that the cache cannot help |

All three tests pass with identical output in both builds on every run.

### Wall clock

Whole `hardhat test solidity` invocation, no profiler attached, 5 alternating before/after rounds:

| build | rounds (s) | mean |
| --- | --- | --- |
| before | 11.15, 11.25, 10.49, 10.45, 10.41 | **10.75 s** |
| after | 8.07, 7.67, 7.41, 7.27, 7.49 | **7.58 s** |

**−29% wall time.** The three tests run in parallel, so wall time is bounded by the slowest one; the distinct-key control does not speed up, which makes this a lower bound on the per-call improvement.

### CPU attribution (perf, 3 rounds aggregated)

| category (inclusive) | before | after |
| --- | --- | --- |
| total CPU samples | 72,435 | 40,794 (**−44%**) |
| inside `vm.addr` (`addrCall`) | 21,276 (29.4%) | 139 (0.3%) |
| private key → wallet derivation on a miss | 35,569 (49.1%) | 4,693 (11.5%) |
| inside `vm.sign` (`signCall`) | 31,222 (43.1%) | 21,503 (52.7%) |
| `ecrecover` precompile | 12,229 (16.9%) | 12,102 (29.7%) |
| any k256 / secp256k1 arithmetic | 61,791 (85.3%) | 30,818 (75.5%) |

Top exclusive symbols before: `k256::ProjectivePoint::add` (27.3%), `FieldElement::square` (13.5%), `ProjectivePoint::double` (13.5%), `secp256k1_fe_mul_inner` (8.1%) — the frames named in #1658. After, the k256 frames that remain are reached from signing and from the all-miss control test.

### Conclusion

- The derivation cost for repeated keys is eliminated: `vm.addr` falls from ~29% of CPU to noise, and total CPU drops 44%.
- The 11.5% left under the cache lookup is entirely `testFuzz_AddrDistinctKeys`, where every key is new and must be derived; this is the expected floor.
- `vm.sign` and `ecrecover` absolute sample counts are unchanged (their percentages rise only because the total shrank), confirming per-message signing is still performed on every call and only the key → address derivation is cached.
- This micro-benchmark is far more derivation-heavy than a real suite, so the 29% is not what aave-v4 will see. The issue attributed ~2% of aave-v4's `test solidity` CPU to `parse_wallet`; the profile shows that cost is removed in full while everything else is untouched, so that is the expected win there.

crates/foundry/cheatcodes/src/crypto.rs
popescuoctavian: `LruCache::get` needs `&mut self`, so it cannot run under a read lock. In order to use `RwLock::read` we would need to use `LruCache::peek` which does not update recency. In fuzz tests that use 256 distinct keys, the (presumably more frequently used) `setUp` keys would be evicted.

`RwLock::write` with `LruCache::get` would update recency, but is basically the same as a `Mutex` with additional steps.

I ran a quick stress test. Performance-wise, they are comparable for single-thread work. In this case, the purpose should be avoiding additional derivations, which are much more costly (see table below).

A mostly unrelated, but quite interesting discovery is that "`RwLock` read + `peek`" performs way worse on 12 threads. This is because the readers do not queue, they all succeed, and every successful read acquire and release is an atomic read-modify-write on the same word. That word lives on one cache line, and on a multi-core machine an atomic RMW needs the line in exclusive state in the issuing core's cache. This causes the cache line to ping-pong between cores.

| Hit path | 1 thread | 12 threads |
| --- | ---: | ---: |
| `Mutex` + `get` (current, keeps LRU order) | 15.9 ns | 59.6 ns |
| `RwLock` read + `peek` (loses LRU order) | 14.5 ns | 198.2 ns |
| `RwLock` write + `get` (keeps LRU order) | 15.4 ns | 58.3 ns |
| One secp256k1 derivation, for scale | 28,300 ns | |

crates/foundry/cheatcodes/src/crypto.rs
popescuoctavian: It is not, but it was required by `CheatsConfig`. On a quick check, I noticed that it's not needed on `CheatsConfig` either, so I'm dropping it altogether: b0f5faebf22b5419b90a84218117311d09790bd5

crates/foundry/cheatcodes/src/crypto.rs
popescuoctavian: Applied to all other tests as well: 5f88f6839439c58c6237916fc2d19365de2f6073

crates/foundry/cheatcodes/src/crypto.rs
Wodann: Thanks for testing! Let's keep it as is then!

PR #1755, opened 2026-09-16
ci: keep napi's type-definition cache between runs

nebasuke:
# Claude summary

`napi build` regenerates `index.d.ts` by re-running the `#[napi]` proc macros, so it needs a real compile to produce them. It forces one through `NAPI_FORCE_BUILD_<CRATE>`, an env var `napi_build::setup()` declares `cargo::rerun-if-env-changed` on — set to `Date.now()` when `target/napi-rs/<crate>-<hash>/<crate>` is missing, and not set at all once it exists.

`target/napi-rs/` isn't among the paths rust-cache preserves, so every job starts cold and pays that forced rebuild. A job that runs `pnpm build:dev` twice pays a *second* one, because the var goes from set to unset the moment the first build creates the cache — and a set→unset transition is a change like any other.

Measured locally on `edr_napi`, three consecutive `pnpm build:dev`:

| Run | `target/napi-rs/` | `NAPI_FORCE_BUILD_EDR_NAPI` | Result |
| --- | --- | --- | --- |
| A | absent | `1789…` | 361s, rebuild (intended) |
| B | present | *unset* | 95s, rebuild — set→unset |
| C | present | *unset* | **10s, `Finished` in 0.69s** |

Passing `cache-directories: target/napi-rs` through `setup-rust` should collapse A and B. Defaulted in the composite action rather than set per call site: there are eight `setup-rust` invocations across five workflows, and a job that never builds the napi crate simply has no directory to save.

## Why a persisted typedef cache is safe

The cache is only *reused* when cargo reports `Fresh`, which means the sources that produced it are the sources on disk. `output_type_def` appends, but `expand()` truncates the file on the first `#[napi]` expansion per rustc process (`BUILT_FLAG`), so every real compile wipes and rewrites it. `target/napi-rs/` and the cargo artifacts share one cache archive, so they always restore together from the same build.

Verified end-to-end with four `pnpm build:typingFile` runs against a warm cache:

| Step | Source state | Recompiled | probe in `index.d.ts` | `index.d.ts` sha |
| --- | --- | --- | --- | --- |
| baseline | HEAD | yes | absent | = committed |
| add `#[napi]` probe | +1 export | yes | **present** | changed |
| remove probe | HEAD | yes | **absent** | = committed |
| rebuild, no change | HEAD | **no (Fresh)** | absent | = committed |

`git status` clean throughout. The last row is the path this PR newly enables: cargo `Fresh`, typings written from cache, byte-identical to the committed file. The removal step rules out append-mode staleness.

Two further guards. The typedef cache is keyed on profile — `build:dev` (release) and `build:typingFile` (debug) get separate hash folders, so they cannot share a stale cache. And `Check that edr_napi typings file is up to date` uses only `setup-node`, no `setup-rust`: it builds cold every run and then `git diff --exit-code`, so it stays an independent check this PR cannot touch.

## Verification protocol

Run 1 is cold and writes the cache; run 2 (re-run all jobs on the same commit — same `Cargo.lock`, so same rust-cache key) should restore it. In `Run integration tests`, grep for `Compiling edr_napi v`: today that's 1 occurrence on `main` and 2 on #1574. Success is **0**, with the job dropping from ~200s toward ~100s.

On run 1, also check rust-cache's post-job log lists `target/napi-rs`, and grep for `Unexpected input(s)` — `cache-directories` is what `actions-rust-lang/setup-rust-toolchain` documents as its rust-cache pass-through, but I could not confirm it against v1.17.0's `action.yml` offline. A wrong name is a silent no-op, which would look exactly like "the cache didn't help".

## Before merge

Drop `b6ee5b35f` — it widens `cache-save-if` to this branch so run 1 can save at all. Saving from every branch would evict `main`'s caches under the repo's 10GB cap.


nebasuke:
Did not make a material difference, and finnicky.

PR #1756, opened 2026-09-16
feat: add collection of EIP-712 types

Wodann:


crates/edr_solidity_tests/src/inline_config/parse.rs
```diff
@@ -51,44 +42,17 @@ pub struct LocatedContract {
...
+/// Walks the already-built `unit`'s file `file_id` (the id the root was added
+/// under) and returns every contract definition together with its functions
+/// and the offsets required to recover their leading NatSpec. A file id
+/// missing from the unit (e.g. the root file could not be read) yields no
+/// contracts.
+pub(crate) fn locate_contracts_in_unit(
+    unit: &CompilationUnit,
+    file_id: &str,
+) -> Vec<LocatedContract> {
+    let Some(file) = unit.file(file_id) else {
```
Wodann: Thought: I didn't want to modify the behaviour, but is this an error state that we should fail on or should we silently accept it?

crates/edr_solidity_tests/src/multi_runner.rs
```diff
@@ -323,100 +307,55 @@ impl<
...
-        }
+        Ok(contracts
+            .into_iter()
+            .map(|(artifact_id, contract)| {
+                let source = collected_sources
+                    .get(&artifact_id.source)
+                    .map(|collections| {
+                        SuiteSourceData::new(collections, &artifact_id, &contract.abi)
+                    })
+                    .unwrap_or_default();
```
Wodann: thought: Should this fail for any other reason than that `test_source_paths` was empty? If so, should we return an error mode for that?

crates/foundry/cheatcodes/src/fs.rs
```diff
@@ -1595,15 +1589,29 @@ mod tests {
...
-    use crate::CheatsConfig;
+    use crate::{CheatsConfig, CheatsConfigOptions};
 
     fn cheats(
     ) -> Cheatcodes<BlockEnv, TxEnv, (), L1EvmBuilder, HaltReason, SpecId, InvalidTransaction> {
-        let config = CheatsConfig {
-            ffi: true,
-            project_root: PathBuf::from(&env!("CARGO_MANIFEST_DIR")),
-            ..Default::default()
+        let artifact = ArtifactId {
```
Wodann: Note: use a fake `ArtifactId` instead of passing `None`. This won't affect the test itself, just removes the need for `Option`, which only existed to support test cases like this

crates/foundry/cheatcodes/src/inspector.rs
```diff
@@ -557,33 +557,6 @@ pub struct Cheatcodes<
     >,
 }
 
-// This is not derived because calling this in `fn new` with
```
Wodann: Note: No longer necessary.

crates/foundry/evm/evm/src/inspectors/stack.rs
```diff
@@ -95,8 +95,8 @@ impl<HardforkT: HardforkTr, ChainContextT: ChainContextTr>
 
     /// Enable cheatcodes with the given config.
     #[inline]
-    pub fn cheatcodes(mut self, config: Arc<CheatsConfig<HardforkT>>) -> Self {
-        self.cheatcodes = Some(config);
+    pub fn cheatcodes(mut self, config: CheatsConfig<HardforkT>) -> Self {
```
Wodann: Note: the caller doesn't need to know about the internal memory representation, so remove `Arc` from the API.

PR #1759, opened 2026-09-17
fix: repair stale hardhat@3.16.0 patch hash in pnpm-lock.yaml

popescuoctavian:
Fixes CI failures introduced by stale `pnpm-lock.yaml` in #1574.


nebasuke (approved):
Looks like dependency safety check is passing now, LGTM!

PR #1764, opened 2026-09-18
ci(edr): run the integration-test packages sequentially

nebasuke:
Fixes the intermittent `Run integration tests` failure in `pretest` (`TS2304: Cannot find name '…'` from the typings check) seen on `main`, in the merge queue for #1705 and on `feat/eip-8037-state-gas` on 2026-09-17.

Tested locally (and on CI), using Claude.

# Claude summary

**Cause.** Since #1574 both `js/integration-tests/solidity-tests` and `js/integration-tests/solx-parity-sweep` run the root `pnpm build:dev` from `pretest`. `pnpm run --recursive` starts unrelated packages in parallel, so two `napi build`s write `crates/edr_napi/index.d.ts` at the same time and `tsc -p tsconfig.typings-check.json` in `scripts/build_edr_napi.sh` reads a half-written file. Before #1574 the sweep's `pretest` (`scripts/maybe-build.js`) exited early in CI because `@nomicfoundation/hardhat-solx` was not installed, so only one build ran.

**Fix.** `pnpm run --recursive --sequential --filter './js/integration-tests/*' test` in `edr-ci.yml`. Verified on pnpm 11.25.0 that `--sequential` runs the filtered packages one after another (same effect as `--workspace-concurrency=1`, and pnpm switches to plain streamed output instead of prefixed buffering).

**Cost, measured on this PR's run.** The second `pnpm build:dev` is not free: `napi build` recompiles `edr_napi` once more (1m49s) because of napi-rs's `NAPI_FORCE_BUILD_EDR_NAPI` absent→set→unset transition, which only settles on the third consecutive build. The job went from 3m57s (last green `main` before #1574) to 6m15s here; the parity sweep's own tests are 14s of that.

**Not done.** Building once in a dedicated CI step and calling a `testNoBuild` script in both packages (the convention `crates/edr_napi` already uses in `edr-npm-release.yml`) would drop the ~2m second pass and keep local `pnpm test` unchanged. Left out to keep this a one-line revert of the concurrency.


anaPerezGhiglia (approved):
LGTM, thanks for taking care of fixing it! 🙇 

I checked and the execution of a successful job in main after #1574 getting merged was 6m 9s (see [job]( https://github.com/NomicFoundation/edr/actions/runs/35236356667/job/105253592355)), so this doesn't seem to add so much overhead on top of it

PR #1776, opened 2026-09-20
ci(release): install Rust through the repo's setup-rust action

nebasuke:
I've moved over the release workflow to use the same Rust setup as the rest of the CI. Was pushed to do this as the old style of pinning can cause the SHA to change for the #v1 action and this made main `fail`.

# Claude summary

The release workflow was the last place in the repo installing Rust through a third-party action, and the only place building on `stable` rather than the `1.98.1` that `rust-toolchain` pins. It now goes through `./.github/actions/setup-rust` like the other 21 call sites, which needed one new pass-through input on the composite.

`dtolnay/rust-toolchain` is gone from the repo entirely.

## Why it was different

Not a deliberate release policy — an artefact of ordering:

| Date | |
|---|---|
| 2023-10-05 | `edr-npm-release.yml` created (#4415), using `dtolnay/rust-toolchain@stable` |
| 2024-08-21 | `.github/actions/setup-rust` introduced (#609) |
| 2024-09-25 | #678 moves the other workflows over — scoped to replacing `actions-rs`, which this workflow never used |

The release workflow predates the composite by eleven months and fell outside the only sweep that would have migrated it. Everything since has been upkeep of the pin it happened to have (#1178 pinned the SHA, #1341 bumped it, #1678 rewrote the comment to `# v1`), never the `toolchain: stable` underneath.

## Two behaviour changes

**Published binaries move from `stable` to 1.98.1.** Release builds now track `rust-toolchain` — so they move when Renovate bumps it (#1743, #1680, #1557 …) rather than silently following stable. This is the reproducibility win and the reason for the change, but it is a change to shipped artifacts. Passing `toolchain: stable` explicitly reverts just this half while keeping the rest.

**Windows `+crt-static` now comes from the composite's `RUSTFLAGS` augmentation** rather than `.cargo/config.toml`'s `[target.x86_64-pc-windows-msvc]` table. Same flag on the same builds today. The consequence is that env `RUSTFLAGS` replaces that table wholesale — Cargo's "first source wins, no merging" — so a per-target flag added there later would not reach release Windows builds.

`cache: false` keeps the status quo: `dtolnay/rust-toolchain` never cached, and the job's own conditional `actions/cache` step is untouched.

## Checks

Run against upstream's pinned source (`actions-rust-lang/setup-rust-toolchain@166cdcfd`):

- **Empty `target` at the existing 21 call sites is a no-op.** The install is guarded by `if [[ -n $_srt_targets ]]`, and the flag loop `for t in ${_srt_targets//,/ }` iterates zero times when empty.
- **The composite finds `rust-toolchain` from a job with a non-root `working-directory`.** The release `build` job sets `defaults.run.working-directory: ./crates/edr_napi`; had detection missed, it would fall through to `rustup toolchain install stable` and the change would be a silent no-op. `hh3-regression-benchmark.yml:regression-benchmark` is the existing precedent — same composite, no inputs, `working-directory: crates/edr_napi`.

Locally: `actionlint` clean, `zizmor 1.26.1` with `GH_TOKEN` clean, `prettier --check ".github/**/*.yml"` clean.

The release job only runs on release commits, so nothing here exercises it end to end — the two behaviour changes above land first on an actual release.


popescuoctavian (approved):
LGTM, thanks!

